PR #6, opened 2018-07-12
Snapshot apis

xing-yang:
This PR adds snapshot APIs as CRD and adds generated files.

This PR depends on the PR "Add Makefile and Gopkg files"


pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,338 @@
...
+
+// These are valid conditions of a volume snapshot.
+const (
+	// VolumeSnapshotConditionCreating means the snapshot is being created but
+	// it is not cut yet.
+	VolumeSnapshotConditionCreating VolumeSnapshotConditionType = "Creating"
+	// VolumeSnapshotConditionUploading means the snapshot is cut and the application
+	// can resume accessing data if ConditionStatus is True. It corresponds
+	// to "Uploading" in GCE PD or "Pending" in AWS and ConditionStatus is True.
+	// This condition type is not applicable in OpenStack Cinder.
```
jingxu97: maybe make this comment a little more generic. "Uploading" condition is only applicable to the volume plugins which support uploading snapshots to Cloud storage. For example, it corresponds to snapshot "UPLOADING" status in GCE and ...

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,338 @@
...
+	VolumeSnapshotConditionUploading VolumeSnapshotConditionType = "Uploading"
+	// VolumeSnapshotConditionReady is added when the snapshot has been successfully created and is ready to be used.
```
jingxu97: change "is added" to "means"

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,338 @@
...
+	VolumeSnapshotConditionReady VolumeSnapshotConditionType = "Ready"
+	// VolumeSnapshotConditionError means an error occurred during snapshot creation.
+	VolumeSnapshotConditionError VolumeSnapshotConditionType = "Error"
+)
+
+// VolumeSnapshotCondition describes the state of a volume snapshot  at a certain point.
```
jingxu97: remove at a certain point.

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,338 @@
...
+type VolumeSnapshot struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// Spec represents the desired state of the snapshot
+	// +optional
+	Spec VolumeSnapshotSpec `json:"spec" protobuf:"bytes,2,opt,name=spec"`
```
jingxu97: I think Spec should be required

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,338 @@
...
+	metav1.ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	Items []VolumeSnapshot `json:"items" protobuf:"bytes,2,rep,name=items"`
+}
+
+// VolumeSnapshotSpec is the desired state of the volume snapshot
+type VolumeSnapshotSpec struct {
+	// PersistentVolumeClaimName is the name of the PVC being snapshotted
+	// +optional
+	PersistentVolumeClaimName string `json:"persistentVolumeClaimName" protobuf:"bytes,1,opt,name=persistentVolumeClaimName"`
```
jingxu97: this is also a required field

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,338 @@
...
+// VolumeSnapshotData represents the actual "on-disk" snapshot object
+type VolumeSnapshotData struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// Spec represents the desired state of the snapshot
+	// +optional
```
jingxu97: spec should be required?

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,338 @@
...
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// VolumeSnapshotDataList is a list of VolumeSnapshotData objects
+type VolumeSnapshotDataList struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
+	metav1.ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	Items []VolumeSnapshotData `json:"items" protobuf:"bytes,2,rep,name=items"`
```
jingxu97: better to add comments to all fields?

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,338 @@
...
+
+// VolumeSnapshotDataSpec is the spec of the volume snapshot data
+type VolumeSnapshotDataSpec struct {
+	// Source represents the location and type of the volume snapshot
+	VolumeSnapshotSource `json:",inline" protobuf:"bytes,1,opt,name=volumeSnapshotSource"`
+
+	// VolumeSnapshotRef is part of bi-directional binding between VolumeSnapshot
+	// and VolumeSnapshotData
+	// +optional
+	VolumeSnapshotRef *core_v1.ObjectReference `json:"volumeSnapshotRef" protobuf:"bytes,2,opt,name=volumeSnapshotRef"`
```
jingxu97: also required field?

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,338 @@
...
+
+	// PersistentVolumeRef represents the PersistentVolume that the snapshot has been
+	// taken from
+	// +optional
+	PersistentVolumeRef *core_v1.ObjectReference `json:"persistentVolumeRef" protobuf:"bytes,3,opt,name=persistentVolumeRef"`
```
jingxu97: required field?

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,338 @@
...
+	// Driver is the name of the driver to use for this snapshot.
+	// Required.
+	Driver string `json:"driver"`
+
+	// SnapshotHandle is the unique snapshot id returned by the CSI volume
+	// plugin’s CreateSnapshot to refer to the snapshot on all subsequent calls.
+	// Required.
+	SnapshotHandle string `json:"snapshotHandle"`
+
+	// Timestamp when the point-in-time snapshot is taken on the storage
```
jingxu97: it is not very clear that which time is used here.

jingxu97:
cc @liggitt

jingxu97:
cc @saad-ali

pkg/apis/volumesnapshot/v1alpha1/types.go
xing-yang: Done

pkg/apis/volumesnapshot/v1alpha1/types.go
xing-yang: Done.

pkg/apis/volumesnapshot/v1alpha1/types.go
xing-yang: PersistentVolumeClaimSpec is optional in PersistentVolumeClaim API object.  Should we be consistent with that?

pkg/apis/volumesnapshot/v1alpha1/types.go
xing-yang: Added.

pkg/apis/volumesnapshot/v1alpha1/types.go
xing-yang: I see that ClaimRef is an optional field in PersistentVolumeSpec. It becomes non-nil when bound.  I'll add a comment here.

pkg/apis/volumesnapshot/v1alpha1/types.go
xing-yang: Similarly, this becomes non-nil when bound.

pkg/apis/volumesnapshot/v1alpha1/types.go
xing-yang: I see that PersistentVolumeSpec is an optional field in PersistentVolume.

pkg/apis/volumesnapshot/v1alpha1/types.go
xing-yang: I see that PersistentVolumeSpec is an optional field in PersistentVolume.  So we should probably be consistent with that?

pkg/apis/volumesnapshot/v1alpha1/types.go
xing-yang: I think by leaving this one optional, we can support static binding?

jingxu97: oh, I think you are right

pkg/apis/volumesnapshot/v1alpha1/types.go
jingxu97: ok.

pkg/apis/volumesnapshot/v1alpha1/types.go
xing-yang: Ok

pkg/apis/volumesnapshot/v1alpha1/types.go
jingxu97: I didn't see the change?

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,338 @@
...
+	CreationTimestamp metav1.Time `json:"creationTimestamp" protobuf:"bytes,1,opt,name=creationTimestamp"`
+
+	// Represent the latest available observations about the volume snapshot
+	Conditions []VolumeSnapshotCondition `json:"conditions" protobuf:"bytes,2,rep,name=conditions"`
+}
+
+// VolumeSnapshotConditionType is the type of VolumeSnapshot conditions
+type VolumeSnapshotConditionType string
+
+// These are valid conditions of a volume snapshot.
```
liggitt: would recommend reviewing the updated condition guidelines in https://github.com/kubernetes/community/pull/2350/files

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,338 @@
...
+	"encoding/json"
+
+	core_v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const (
+	// VolumeSnapshotDataResourcePlural is "volumesnapshotdatas"
+	VolumeSnapshotDataResourcePlural = "volumesnapshotdatas"
```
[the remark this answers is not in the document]
xing-yang: Do you have a suggestion for this?  "data" is a plural itself.  How do we differentiate between singular and plural if they are the same?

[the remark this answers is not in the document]
xing-yang: Looks like "volumesnapshotdatas" will be generated automatically for clientset when generating code for the CRD. So unless if we rename VolumeSnapshotData to something else, it will be "volumesnapshotdatas".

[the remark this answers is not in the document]
liggitt: > Looks like "volumesnapshotdatas" will be generated automatically for clientset when generating code for the CRD. So unless if we rename VolumeSnapshotData to something else, it will be "volumesnapshotdatas".

That's correct, and is a gap in the client generator. This has also been reported by others (see https://kubernetes.slack.com/archives/C0EG7JC6T/p1528416072000218). However, we shouldn't accept an incorrect API because of that... it would be better to participate in improving the generator to allow specifying the plural.

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,338 @@
...
+	// Items is the list of SnapshotClasses
+	Items []SnapshotClass `json:"items" protobuf:"bytes,2,rep,name=items"`
+}
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// VolumeSnapshotData represents the actual "on-disk" snapshot object
+type VolumeSnapshotData struct {
```
liggitt: is the relationship between VolumeSnapshot and VolumeSnapshotData `1:1` or `1:*`?

pkg/apis/volumesnapshot/v1alpha1/types.go
jingxu97: I read the updated condition guideline again, and have the following thoughts.

Condition should be considered as observations that users cared about instead of state-machine phases. For snapshot, user cares about whether snapshot is created successfully or not first. These can be represented by "ready" and "failed" conditions. The other condition user cares about is when the snapshot is taken (or cut). Because some plugins support uploading which might take a long time. User needs to know when the snapshot is taken (usually in a second) so that application can be resumed. So we could add another condition "snapshotTaken" (or just "Taken") condition. So now we have three conditions "snapshotTaken", "snapshotReady" and "snapshotFailed"

During the snapshot creation, we might see the following conditions at different times.
1. First when user create VolumeSnapshot object, there is no condition yet. (means controller is working on it, but there is no observed state available yet) 
2. The volume plugin starts to create the snapshot. When the snapshot is taken (cut), controller will get a response and add "SnapshotTaken" condition to true. If snapshot is failed, it will be set to false. If there is no uploading phase, Ready condition will also be true right after snapshot is taken.
3. If there is an uploading phase, the controller will add "Ready" condition to true after uploading finishes. If uploading failed, the condition is set to false. 

Please let me know if you have any feedback/comment about it. Thanks!

pkg/apis/volumesnapshot/v1alpha1/types.go
xing-yang: Renamed VolumeSnapshotData to VolumeSnapshotContent.

pkg/apis/volumesnapshot/v1alpha1/types.go
xing-yang: Changed VolumeSnapshotConditionType to "Created", "Available", and "Error" following today's discussion during the snapshot design meeting.

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,334 @@
...
+	// Items is the list of SnapshotClasses
+	Items []SnapshotClass `json:"items" protobuf:"bytes,2,rep,name=items"`
+}
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// VolumeSnapshotContent represents the actual "on-disk" snapshot object
+type VolumeSnapshotContent struct {
```
kangarlou: I know we had some discussion on this in the last call. Wouldn't VolumeSnapshotClaim or VolumeSnapshotRequest as the user-facing object and VolumeSnapshot as the storage backend representation be better names? These names follow the PersistentVolumeClaim-PersistentVolume convention and can be pluralized.

pkg/apis/volumesnapshot/v1alpha1/register.go
```diff
@@ -0,0 +1,58 @@
...
+	// We only register manually written functions here. The registration of the
+	// generated functions takes place in the generated files. The separation
+	// makes the code compile even when the generated files are missing.
+	SchemeBuilder.Register(addKnownTypes)
+}
+
+// addKnownTypes adds the set of types defined in this package to the supplied scheme.
+func addKnownTypes(scheme *runtime.Scheme) error {
+	scheme.AddKnownTypes(SchemeGroupVersion,
+		&SnapshotClass{},
```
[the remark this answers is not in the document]
xing-yang: Ok, will change to VolumeSnapshotClass.

pkg/apis/volumesnapshot/v1alpha1/types.go
jingxu97: Unfortunately, PersistentVolumeClaim does not follow the name pattern as all other user-facing objects. For example, we don't use PodClaim to represent user's pod. We think we should have named PVC and PV as PersistentVolume (user) and PersistentVolumeContent (system admin) at the first place. So instead of following the wrong pattern, we decided to use VolumeSnapshot to represent the user's snapshot.

kangarlou: PVCs don't follow other user-facing objects because other objects don't follow the dual-object model where we have a request and a response or a user-facing representation and an internal representation, but I agree with you that it's better not to follow the wrong pattern and have the more straightforward name (i.e., VolumeSnapshot) exposed to the users.

pkg/apis/volumesnapshot/v1alpha1/register.go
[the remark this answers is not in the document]
xing-yang: Done

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,316 @@
...
+	// Timestamp when the point-in-time snapshot is taken on the storage
+	// system. This timestamp will be generated by the CSI volume driver after
+	// the snapshot is cut. The format of this field should be a Unix nanoseconds
+	// time encoded as an int64. On Unix, the command `date +%s%N` returns
+	// the  current time in nanoseconds since 1970-01-01 00:00:00 UTC.
+	// This field is REQUIRED.
+	CreatedAt int64 `json:"createdAt,omitempty" protobuf:"varint,3,opt,name=createdAt"`
+}
+
+// GetObjectKind is required to satisfy Object interface
```
wackxu: the following is not need, so can remove these lines.

xing-yang: Sure. I'll take a look.  Thanks.

pkg/apis/volumesnapshot/v1alpha1/types.go
garthy: OK I'm late to to conversation so don't really have much right to input. I would prefer sticking to the PVC/PV pattern even if you belive it is wrong as it exists now. I don't belive it is incorrect as @kangarlou says noting else follows the request response pattern. It will as

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -38,18 +38,12 @@ type VolumeSnapshotStatus struct {
...
-	// resumed if it was previously frozen before taking the snapshot. In this case,
-	// it is possible that the snapshot is being uploaded to the cloud. For example,
-	// both GCE and AWS support uploading of the snapshot after it is cut as part of
-	// the Create Snapshot process.
-	// If the timestamp AvailableAt is set, it means the snapshot was available;
-	// Otherwise the snapshot was not available.
-	// +optional
-	AvailableAt *metav1.Time `json:"availableAt" protobuf:"bytes,2,opt,name=availableAt"`
+	// Bound is set to true only if the snapshot is ready to use (e.g., finish uploading if
+	// there is an uplaoding phase) and also VolumeSnapshot andn its VolumeSnapshotContent
```
xing-yang: s/uplaoding/uploading

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -93,15 +88,17 @@ type VolumeSnapshotList struct {
 
 // VolumeSnapshotSpec is the desired state of the volume snapshot
 type VolumeSnapshotSpec struct {
-	// PersistentVolumeClaimName is the name of the PVC being snapshotted
+	// PersistentVolumeClaimName is the name of the PVC being snapshotted.
+	// If not specified, user can create SnapshotContent and bind it with VolumeSnapshot manually.
```
xing-yang: s/SnapshotContent/VolumeSnapshotContent

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -38,18 +38,12 @@ type VolumeSnapshotStatus struct {
...
-	// resumed if it was previously frozen before taking the snapshot. In this case,
-	// it is possible that the snapshot is being uploaded to the cloud. For example,
-	// both GCE and AWS support uploading of the snapshot after it is cut as part of
-	// the Create Snapshot process.
-	// If the timestamp AvailableAt is set, it means the snapshot was available;
-	// Otherwise the snapshot was not available.
-	// +optional
-	AvailableAt *metav1.Time `json:"availableAt" protobuf:"bytes,2,opt,name=availableAt"`
+	// Bound is set to true only if the snapshot is ready to use (e.g., finish uploading if
+	// there is an uploading phase) and also VolumeSnapshot andn its VolumeSnapshotContent
```
xing-yang: s/andn/and

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,224 @@
...
+type VolumeSnapshotSource struct {
+	// CSI (Container Storage Interface) represents storage that handled by an external CSI Volume Driver (Alpha feature).
+	// +optional
+	CSI *CSIVolumeSnapshotSource `json:"csiVolumeSnapshotSource,omitempty"`
+}
+
+// Represents the source from CSI volume snapshot
+type CSIVolumeSnapshotSource struct {
+	// Driver is the name of the driver to use for this snapshot.
+	// Required.
```
saad-ali: Please add:

```
    // This MUST be the same name returned by the CSI GetPluginName() call for
    // that driver.
```

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -0,0 +1,224 @@
...
+// VolumeSnapshotContent represents the actual "on-disk" snapshot object
+type VolumeSnapshotContent struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// Spec represents the desired state of the snapshot data
+	Spec VolumeSnapshotContentSpec `json:"spec" protobuf:"bytes,2,opt,name=spec"`
```
saad-ali: No status in `VolumeSnapshotContent`? No way to know if the object is bound or not?

PR #7, opened 2018-07-12
Add Snapshot Controller

xing-yang:
This PR adds external snapshot controller and main package under cmd.

Which issue(s) this PR fixes:
Fixes #
kubernetes/features#177


pkg/connection/connection.go
```diff
@@ -0,0 +1,337 @@
...
+
+	snapDataName := GetSnapshotDataNameForSnapshot(snapshot)
+
+	persistentVolumeRef, err := ref.GetReference(scheme.Scheme, volume)
+	if err != nil {
+		return nil, nil, fmt.Errorf("unexpected error getting volume reference: %v", err)
+	}
+
+	// Create VolumeSnapshotData in the database
+	snapshotData := &crdv1.VolumeSnapshotData{
```
jingxu97: I feel it is better to create VolumeSnapshotData object in the controller, not in connection code.

cmd/csi-snapshotter/main.go
```diff
@@ -1,7 +1,217 @@
...
+			ResourceVersion:   "514",
+			CreationTimestamp: metav1.Now(),
+		},
+		Spec: crdv1.VolumeSnapshotSpec{
+			PersistentVolumeClaimName: "testpvc",
+			SnapshotClassName:         "testsnapclass",
+		},
+	}
+
+	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(vs)
```
jingxu97: remove this part?

cmd/csi-snapshotter/main.go
```diff
@@ -1,7 +1,217 @@
...
+	// make a new config for our extension's API group, using the first config as a baseline
+	crdClient, _, err := NewClient(config)
+	if err != nil {
+		panic(err)
+	}
+
+	// wait until CRD gets processed
+	err = WaitForSnapshotResource(crdClient)
+	if err != nil {
+		panic(err)
```
jingxu97: why use panic instead of os.Exist like below?

pkg/connection/connection.go
xing-yang: Sure.

pkg/controller/util.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	glog.V(4).Infof("storeObjectUpdate updating %s %q with version %s", className, objName, objAccessor.GetResourceVersion())
+	if err = store.Update(obj); err != nil {
+		return false, fmt.Errorf("error updating %s %q in controller cache: %v", className, objName, err)
+	}
+	return true, nil
+}
+
+func GenSnapshotStatus(snapshotCon *crdv1.VolumeSnapshotCondition) *crdv1.VolumeSnapshotCondition {
```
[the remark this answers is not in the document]
xing-yang: This makes a copy of of the Condition.  Probably not needed.  We can use DeepCopy.

cmd/csi-snapshotter/main.go
xing-yang: Sure.

cmd/csi-snapshotter/main.go
xing-yang: Will change it.

pkg/connection/connection.go
```diff
@@ -0,0 +1,337 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: snapDataName,
+		},
+		Spec: crdv1.VolumeSnapshotDataSpec{
+			VolumeSnapshotRef: &v1.ObjectReference{
```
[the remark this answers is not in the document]
xing-yang: Sure.  I'll check it here.

pkg/connection/connection.go
```diff
@@ -0,0 +1,266 @@
...
+	client := csi.NewControllerClient(c.conn)
+
+	driverName, err := c.GetDriverName(ctx)
+	if err != nil {
+		return "", "", 0, nil, err
+	}
+
+	req := csi.CreateSnapshotRequest{
+		SourceVolumeId:        volume.Spec.CSI.VolumeHandle,
+		Name:                  snapshot.Name,
```
jingxu97: should use snapshot.UID? Different namespaces might have the same snapshot name which will cause problem here.

pkg/connection/connection.go
```diff
@@ -242,14 +242,14 @@ func (c *csiConnection) ListSnapshots(ctx context.Context, snapshotID string) (*
...
+		return nil, 0, err
 	}
 
 	if rsp.Entries == nil || len(rsp.Entries) == 0 {
-		return nil, fmt.Errorf("can not find snapshot for snapshotID %s", snapshotID)
+		return nil, 0, fmt.Errorf("can not find snapshot for snapshotID %s", snapshotID)
 	}
 
-	return rsp.Entries[0].Snapshot.Status, nil
+	return rsp.Entries[0].Snapshot.Status, rsp.Entries[0].Snapshot.CreatedAt, nil
```
jingxu97: I thought this rsp.Entries[0].Snapshot.CreatedAt is when snapshot is cut, not when it is ready or something else.

pkg/connection/connection.go
```diff
@@ -0,0 +1,266 @@
...
+	driverName, err := c.GetDriverName(ctx)
+	if err != nil {
+		return "", "", 0, nil, err
+	}
+
+	req := csi.CreateSnapshotRequest{
+		SourceVolumeId:        volume.Spec.CSI.VolumeHandle,
+		Name:                  snapshotName,
+		Parameters:            parameters,
+		CreateSnapshotSecrets: nil,
```
sngchlko: As you know, some storage providers need secrets to create the snapshot.
Do you have any plan to implement this?

xing-yang: @sngchlko Thanks for the feedback!  Yes, this should be fixed.

cmd/csi-snapshotter/main.go
```diff
@@ -1,7 +1,186 @@
...
+	if err != nil {
+		glog.Error(err.Error())
+		os.Exit(1)
+	}
+	if !supportsCreateSnapshot {
+		glog.Error("CSI driver does not support ControllerCreateSnapshot")
+		os.Exit(1)
+	}
+
+	glog.V(2).Infof("Start NewCSISnapshotController with snapshotter %s", *snapshotter)
```
lpabon: Can be added after, but it would be nice to log the options.

pkg/controller/csi_handler.go
```diff
@@ -0,0 +1,115 @@
...
+		csiConnection:          csiConnection,
+		timeout:                timeout,
+		snapshotNamePrefix:     snapshotNamePrefix,
+		snapshotNameUUIDLength: snapshotNameUUIDLength,
+	}
+}
+
+func (handler *csiHandler) CreateSnapshot(snapshot *crdv1.VolumeSnapshot, volume *v1.PersistentVolume, parameters map[string]string) (string, string, int64, *csi.SnapshotStatus, error) {
+
+	ctx, cancel := context.WithTimeout(context.Background(), handler.timeout)
```
lpabon: nice

pkg/connection/connection.go
xing-yang: @sngchlko It's fixed here: https://github.com/xing-yang/external-snapshotter/pull/8/commits/dd820b5ef0fad3cdf83c533115a8a1269e1ec1f8

xing-yang: Fixed.

cmd/csi-snapshotter/main.go
xing-yang: Sure.

pkg/connection/connection.go
```diff
@@ -0,0 +1,266 @@
...
+	glog.V(5).Infof("CSI CreateSnapshot: %s driver name [%s] snapshot ID [%s] time stamp [%s] status [%s]", snapshot.Name, driverName, rsp.Snapshot.Id, rsp.Snapshot.CreatedAt, *rsp.Snapshot.Status)
+	return driverName, rsp.Snapshot.Id, rsp.Snapshot.CreatedAt, rsp.Snapshot.Status, nil
+}
+
+func (c *csiConnection) DeleteSnapshot(ctx context.Context, snapshotID string) (err error) {
+	client := csi.NewControllerClient(c.conn)
+
+	req := csi.DeleteSnapshotRequest{
+		SnapshotId:            snapshotID,
+		DeleteSnapshotSecrets: nil,
```
sngchlko: Please add secrets as with creating snapshots.

xing-yang: Will do.  Thanks.

cmd/csi-snapshotter/main.go
```diff
@@ -1,7 +1,186 @@
...
+	}
+
+	// Find out if the driver supports create/delete snapshot.
+	supportsCreateSnapshot, err := csiConn.SupportsControllerCreateSnapshot(ctx)
+	if err != nil {
+		glog.Error(err.Error())
+		os.Exit(1)
+	}
+	if !supportsCreateSnapshot {
+		glog.Error("CSI driver does not support ControllerCreateSnapshot")
```
saad-ali: Put driver name info in the error for easier debugging.

pkg/controller/controller.go
```diff
@@ -0,0 +1,439 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/kubernetes/pkg/util/goroutinemap"
+)
+
+type CSISnapshotController struct {
```
saad-ali: This does not need to be a public type (should be private). Callers should not depend directly on a `struct` only on `Interface`.

pkg/controller/csi_snapshot_controller.go
```diff
@@ -0,0 +1,740 @@
+/*
```
saad-ali: Why is this called `csi_snapshot_controller.go`. This looks like general binding logic, not very csi specific?

cmd/csi-snapshotter/main.go
xing-yang: Sure.

pkg/controller/controller.go
xing-yang: Sure.

pkg/controller/csi_snapshot_controller.go
xing-yang: You are right.  Initially we only had controller.go and csi_handler.go and later restructured the logic and modeled controller.go and csi_snapshot_controller.go after pv_controller_base.go and pv_controller.go.  Maybe we can name them snapshot_controller_base.go and snapshot_controller.go.

cmd/csi-snapshotter/main.go
```diff
@@ -1,7 +1,186 @@
...
+	resync                          = flag.Duration("resync", 10*time.Second, "Resync interval of the controller.")
+	connectionTimeout               = flag.Duration("connection-timeout", 1*time.Minute, "Timeout for waiting for CSI driver socket.")
+	csiAddress                      = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
+	createSnapshotContentRetryCount = flag.Int("createSnapshotContentRetryCount", 5, "Number of retries when we create a snapshot data object for a snapshot.")
+	createSnapshotContentInterval   = flag.Duration("createSnapshotContentInterval", 10*time.Second, "Interval between retries when we create a snapshot data object for a snapshot.")
+	resyncPeriod                    = flag.Duration("resyncPeriod", 60*time.Second, "The period that should be used to re-sync the snapshot.")
+	snapshotNamePrefix              = flag.String("snapshot-name-prefix", "snapshot", "Prefix to apply to the name of a created snapshot")
+	snapshotNameUUIDLength          = flag.Int("snapshot-name-uuid-length", -1, "Length in characters for the generated uuid of a created snapshot")
```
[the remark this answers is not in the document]
xing-yang: -1 means Defaults behavior is to NOT truncate.  Will add to the usage string.

cmd/csi-snapshotter/main.go
```diff
@@ -1,7 +1,186 @@
...
+	kubeconfig                      = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Required only when running out of cluster.")
+	resync                          = flag.Duration("resync", 10*time.Second, "Resync interval of the controller.")
+	connectionTimeout               = flag.Duration("connection-timeout", 1*time.Minute, "Timeout for waiting for CSI driver socket.")
+	csiAddress                      = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
+	createSnapshotContentRetryCount = flag.Int("createSnapshotContentRetryCount", 5, "Number of retries when we create a snapshot data object for a snapshot.")
+	createSnapshotContentInterval   = flag.Duration("createSnapshotContentInterval", 10*time.Second, "Interval between retries when we create a snapshot data object for a snapshot.")
+	resyncPeriod                    = flag.Duration("resyncPeriod", 60*time.Second, "The period that should be used to re-sync the snapshot.")
```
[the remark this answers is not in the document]
xing-yang: resync is used in this routine.  resyncPeriod is passed into the controller.

cmd/csi-snapshotter/main.go
```diff
@@ -1,7 +1,186 @@
...
+	// Number of worker threads
+	threads = 10
+
+	// Default timeout of short CSI calls like GetPluginInfo
+	csiTimeout = time.Second
+)
+
+// Command line flags
+var (
+	snapshotter                     = flag.String("snapshotter", "", "Name of the snapshotter. The snapshotter will only create snapshot data for snapshot that request a StorageClass with a snapshotter field set equal to this name.")
```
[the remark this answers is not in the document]
xing-yang: Good idea.  Will do.

cmd/csi-snapshotter/create_crd.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		glog.Fatalf("failed to create VolumeSnapshotResource: %#v, err: %#v",
+			res, err)
+	}
+
+	return nil
+}
+
+// WaitForSnapshotResource waits for the snapshot resource
+func WaitForSnapshotResource(snapshotClient *rest.RESTClient) error {
```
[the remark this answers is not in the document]
xing-yang: Forgot to remove this.  Will remove it.

cmd/csi-snapshotter/main.go
```diff
@@ -1,7 +1,186 @@
...
+	}
+
+	// Connect to CSI.
+	csiConn, err := connection.New(*csiAddress, *connectionTimeout)
+	if err != nil {
+		glog.Error(err.Error())
+		os.Exit(1)
+	}
+
+	// Find driver name.
```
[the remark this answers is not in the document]
xing-yang: Will fix the comments.

cmd/csi-snapshotter/main.go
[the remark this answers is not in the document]
jingxu97: I think we followed the external-provisioner here to set the snapshotter. External-attacher uses driver name directly, I can follow the same here. 

Should we also change external-privisioner too?

pkg/controller/csi_snapshot_controller.go
jingxu97: This csi_snapshot_controller sits between the case controller and csi_handler, so we use the name csi. 
Logically, it follows pv_controller_base and pv_controller design pattern.

xing-yang:
Addressed review comments.

cmd/csi-snapshotter/main.go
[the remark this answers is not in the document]
jsafrane: The name and description is quite confusing then. How can the user tell the difference?

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+			return fmt.Errorf("snapshot %s is bound to a non-existing content %s", uniqueSnapshotName, snapshot.Spec.SnapshotContentName)
+		}
+		content, ok := contentObj.(*crdv1.VolumeSnapshotContent)
+		if !ok {
+			return fmt.Errorf("expected volume snapshot content, got %+v", contentObj)
+		}
+
+		if err := ctrl.bindSnapshotContent(snapshot, content); err != nil {
+			// snapshot is bound but content is not bound to snapshot correctly
+			ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotBoundFailed", fmt.Sprintf("Snapshot failed to bind VolumeSnapshotContent, %v", err))
```
jsafrane: `SnapshotBindFailed`?

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+		}
+
+		// snapshot is already bound correctly, check the status and update if it is ready.
+		glog.V(4).Infof("Check and update snapshot %s status", uniqueSnapshotName)
+		if err = ctrl.checkandUpdateSnapshotStatus(snapshot, content); err != nil {
+			return err
+		}
+		return nil
+	} else { // snapshot.Spec.SnapshotContentName == nil
+		if ContentObj := ctrl.getMatchSnapshotContent(snapshot); ContentObj != nil {
```
jsafrane: lowercase `c`

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+}
+
+// Delete a snapshot
+// 1. Find the SnapshotContent corresponding to Snapshot
+//   1a: Not found => finish (it's been deleted already)
+// 2. Ask the backend to remove the snapshot device
+// 3. Delete the SnapshotContent object
+// 4. Remove the Snapshot from vsStore
+// 5. Finish
+func (ctrl *csiSnapshotController) DeleteSnapshotContentOperation(content *crdv1.VolumeSnapshotContent) error {
```
jsafrane: lowercase `d` in `DeleteSnapshotContentOperation`? This function is not used outside of this pkg and `createSnapshotOperation` is lowercase

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+			snapshotterCredentials, err = GetCredentials(ctrl.client, snapshotterSecretRef)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	err := ctrl.handler.DeleteSnapshot(content, snapshotterCredentials)
+	if err != nil {
+		return fmt.Errorf("failed to delete snapshot %#v, err: %v", content.Name, err)
```
jsafrane: Should we send event to the SnapshotContent so user (admin) can see what's wrong? This applies to all errors in this function, perhaps the error handling could be moved to `deleteSnapshotContent`.

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+		}
+		if status.CreationTime == nil {
+			status.CreationTime = timeAt
+			change = true
+		}
+	case csi.SnapshotStatus_ERROR_UPLOADING:
+		if status.Error == nil {
+			status.Error = &storage.VolumeError{
+				Time:    *timeAt,
+				Message: "Failed to upload the snapshot",
```
jsafrane: I think that this should be propagated also as an event.

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+		return nil, err
+	}
+	snapshotterCredentials, err := GetCredentials(ctrl.client, snapshotterSecretRef)
+	if err != nil {
+		return nil, err
+	}
+
+	driverName, snapshotID, timestamp, csiSnapshotStatus, err := ctrl.handler.CreateSnapshot(snapshot, volume, class.Parameters, snapshotterCredentials)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to take snapshot of the volume, %s: %q", volume.Name, err)
```
jsafrane: errors should start with lowercase

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+	}
+	glog.Infof("Create snapshot driver %s, snapshotId %s, timestamp %d, csiSnapshotStatus %v", driverName, snapshotID, timestamp, csiSnapshotStatus)
+
+	// Update snapshot status with timestamp
+	newSnapshot, err := ctrl.updateSnapshotStatus(snapshot, csiSnapshotStatus, time.Unix(0, timestamp), false)
```
jsafrane: If update fails, you end up with a snapshot in the storage backend and without `VolumeSnapshotContent` for it.

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+			glog.V(4).Infof("getDefaultClass %d defaults found", len(defaultClasses))
+			return nil, fmt.Errorf("%d default StorageClasses were found", len(defaultClasses))
+		}
+		glog.V(5).Infof("getClassFromVolumeSnapshot [%s]: default VolumeSnapshotClassName [%s]", snapshot.Name, defaultClasses[0])
+		return defaultClasses[0], nil
+
+	}
+}
+
+// getClaimFromVolumeSnapshot is a helper function to get PV from VolumeSnapshot.
```
jsafrane: "get PV from VolumeSnapshot" -> "get PVC from VolumeSnapshot"

pkg/controller/snapshot_controller_base.go
```diff
@@ -0,0 +1,439 @@
...
+		_, name, err := cache.SplitMetaNamespaceKey(key)
+		if err != nil {
+			glog.V(4).Infof("error getting name of snapshotData %q to get snapshotData from informer: %v", key, err)
+			return false
+		}
+		content, err := ctrl.contentLister.Get(name)
+		if err == nil {
+			// The volume still exists in informer cache, the event must have
+			// been add/update/sync
+			ctrl.updateContent(content)
```
jsafrane: Should you skip Content that is for another CSI driver to save some memory and CPU?

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+	} else { // snapshot.Spec.SnapshotContentName == nil
+		if ContentObj := ctrl.getMatchSnapshotContent(snapshot); ContentObj != nil {
+			glog.V(4).Infof("Find VolumeSnapshotContent object %s for snapshot %s", ContentObj.Name, uniqueSnapshotName)
+			newSnapshot, err := ctrl.bindandUpdateVolumeSnapshot(ContentObj, snapshot)
+			if err != nil {
+				return err
+			}
+			glog.V(4).Infof("bindandUpdateVolumeSnapshot %v", newSnapshot)
+			return nil
+		} else if snapshot.Status.Error == nil { // Try to create snapshot if no error status is set
```
jsafrane: This looks suspicious. So we never retry creating a snapshot?

And it looks racy, if a create operation is running in parallel, above we check that there is no error. Just after that the creation operation fails and set the error and finishes. This function resumes and enqueues new creation operation.

IMO, `snapshot.Status.Error` should be checked also in the operation. If it is really necessary.

xing-yang:
@jsafrane We are working on unit tests.  Will add soon.

pkg/controller/snapshot_controller.go
jingxu97: The current design is not retry to create a snapshot. Later we will add a retry policy to allow user specify how long to retry. I can add the check error inside the operation too.

pkg/controller/snapshot_controller_base.go
jingxu97: yes, we should add that check.

pkg/controller/snapshot_controller.go
jingxu97: that's a good point. We can add try to update snapshot a few times similar to trying to create VolumeSnapshotContent a few times.

I have been thinking how to handle failures and this (not) retry policy. Maybe it is better to differentiate some failures.
1. If CSI driver fail to create the snapshot and returns the error, the controller will set the error state and not try to issue the creation again.
2. If CSI driver creates the snapshot successfully, but controller fails to update the status or create snapshotContent, the controller can resync the API and try to update the status again. 

I will see how controller can handle these.

pkg/controller/snapshot_controller.go
jingxu97: will fix it.

pkg/controller/snapshot_controller_base.go
```diff
@@ -0,0 +1,439 @@
...
+func (ctrl *csiSnapshotController) deleteContent(content *crdv1.VolumeSnapshotContent) {
+	_ = ctrl.contentStore.Delete(content)
+	glog.V(4).Infof("content %q deleted", content.Name)
+
+	snapshotName := snapshotRefKey(content.Spec.VolumeSnapshotRef)
+	if snapshotName == "" {
+		glog.V(5).Infof("deleteContent[%q]: content not bound", content.Name)
+		return
+	}
+	// sync the vs when its vs is deleted.  Explicitly sync'ing the
```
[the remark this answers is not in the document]
jingxu97: I think we intended to follow pv controller. 
	// sync the claim when its volume is deleted. Explicitly syncing the
	// claim here in response to volume deletion prevents the claim from
	// waiting until the next sync period for its Lost status.

should change from ctrl.contentQueue.Add to ctrl.snapshotQueue.Add.

cmd/csi-snapshotter/main.go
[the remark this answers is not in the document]
xing-yang: Ok, I think we can just use one resyncPeriod parameter.  Fixed.

cmd/csi-snapshotter/main.go
```diff
@@ -1,7 +1,196 @@
...
+)
+
+// Command line flags
+var (
+	snapshotter                     = flag.String("snapshotter", "", "Name of the snapshotter. The snapshotter will only create snapshot content for snapshot that requests a VolumeSnapshotClass with a snapshotter field set equal to this name.")
+	kubeconfig                      = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Required only when running out of cluster.")
+	resync                          = flag.Duration("resync", 10*time.Second, "Resync interval of the controller.")
+	connectionTimeout               = flag.Duration("connection-timeout", 1*time.Minute, "Timeout for waiting for CSI driver socket.")
+	csiAddress                      = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
+	createSnapshotContentRetryCount = flag.Int("createSnapshotContentRetryCount", 5, "Number of retries when we create a snapshot data object for a snapshot.")
```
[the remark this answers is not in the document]
xing-yang: Fixed.

cmd/csi-snapshotter/main.go
```diff
@@ -1,7 +1,196 @@
...
+		glog.Error(err.Error())
+		os.Exit(1)
+	}
+
+	// Pass a context with a timeout
+	ctx, cancel := context.WithTimeout(context.Background(), csiTimeout)
+	defer cancel()
+
+	// Find driver name
+	if snapshotter == nil {
```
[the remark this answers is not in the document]
xing-yang: Fixed.

pkg/connection/connection.go
```diff
@@ -0,0 +1,266 @@
...
+		}
+		if rpc.GetType() == csi.ControllerServiceCapability_RPC_LIST_SNAPSHOTS {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func (c *csiConnection) CreateSnapshot(ctx context.Context, snapshotName string, snapshot *crdv1.VolumeSnapshot, volume *v1.PersistentVolume, parameters map[string]string, snapshotterCredentials map[string]string) (string, string, int64, *csi.SnapshotStatus, error) {
+	glog.V(5).Infof("CSI CreateSnapshot: %s", snapshot.Name)
```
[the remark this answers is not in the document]
xing-yang: Removed "snapshot *crdv1.VolumeSnapshot".

pkg/connection/connection.go
```diff
@@ -0,0 +1,266 @@
...
+	return rsp.Entries[0].Snapshot.Status, rsp.Entries[0].Snapshot.CreatedAt, nil
+}
+
+func (c *csiConnection) Close() error {
+	return c.conn.Close()
+}
+
+func logGRPC(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+	glog.V(5).Infof("GRPC call: %s", method)
+	glog.V(5).Infof("GRPC request: %+v", req)
```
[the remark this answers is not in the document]
xing-yang: Removed the logging for req now.  It is helpful for debugging though.  We can add back later if there is a fix for kubernetes-csi/external-provisioner#82.

pkg/controller/csi_handler.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return nil, 0, fmt.Errorf("failed to list snapshot data %s: %q", content.Name, err)
+	}
+	return csiSnapshotStatus, timestamp, nil
+}
+
+func makeSnapshotName(prefix, snapshotUID string, snapshotNameUUIDLength int) (string, error) {
+	// create persistent name based on a volumeNamePrefix and volumeNameUUIDLength
+	// of PVC's UID
+	if len(prefix) == 0 {
```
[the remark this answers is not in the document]
xing-yang: Fixed.

pkg/controller/util.go
```diff
@@ -0,0 +1,244 @@
...
+	if err = store.Update(obj); err != nil {
+		return false, fmt.Errorf("error updating %s %q in controller cache: %v", className, objName, err)
+	}
+	return true, nil
+}
+
+// GetSnapshotContentNameForSnapshot returns SnapshotData.Name for the create VolumeSnapshotContent.
+// The name must be unique.
+func GetSnapshotContentNameForSnapshot(snapshot *crdv1.VolumeSnapshot) string {
+	return "snapdata-" + string(snapshot.UID)
```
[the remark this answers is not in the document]
xing-yang: Changed to "snapcontent".

pkg/controller/util.go
```diff
@@ -0,0 +1,244 @@
...
+	"strings"
+)
+
+var (
+	keyFunc = cache.DeletionHandlingMetaNamespaceKeyFunc
+)
+
+// GetNameAndNameSpaceFromSnapshotName retrieves the namespace and
+// the short name of a snapshot from its full name
+func GetNameAndNameSpaceFromSnapshotName(name string) (string, string, error) {
```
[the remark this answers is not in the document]
xing-yang: Removed.

pkg/controller/util.go
```diff
@@ -0,0 +1,244 @@
...
+// supported tokens for namespace resolution:
+// - ${volumesnapshotcontent.name}
+// - ${volumesnapshot.namespace}
+//
+// an error is returned in the following situations:
+// - only one of name or namespace is provided
+// - the name or namespace parameter contains a token that cannot be resolved
+// - the resolved name is not a valid secret name
+// - the resolved namespace is not a valid namespace name
+func GetSecretReference(nameKey, namespaceKey string, snapshotClassParams map[string]string, snapContentName string, snapshot *crdv1.VolumeSnapshot) (*v1.SecretReference, error) {
```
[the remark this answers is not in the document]
xing-yang: Good point.  Done.

pkg/controller/snapshot_controller_base.go
```diff
@@ -0,0 +1,439 @@
...
+		if quit {
+			return true
+		}
+		defer ctrl.contentQueue.Done(keyObj)
+		key := keyObj.(string)
+		glog.V(5).Infof("contentWorker[%s]", key)
+
+		_, name, err := cache.SplitMetaNamespaceKey(key)
+		if err != nil {
+			glog.V(4).Infof("error getting name of snapshotData %q to get snapshotData from informer: %v", key, err)
```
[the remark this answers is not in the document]
xing-yang: Done

pkg/controller/util.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	glog.V(4).Infof("storeObjectUpdate updating %s %q with version %s", className, objName, objAccessor.GetResourceVersion())
+	if err = store.Update(obj); err != nil {
+		return false, fmt.Errorf("error updating %s %q in controller cache: %v", className, objName, err)
+	}
+	return true, nil
+}
+
+// GetSnapshotContentNameForSnapshot returns SnapshotData.Name for the create VolumeSnapshotContent.
```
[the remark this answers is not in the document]
xing-yang: Done

pkg/controller/snapshot_controller_base.go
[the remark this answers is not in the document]
xing-yang: Looks like a copy & paste error.  It's fixed now.

pkg/controller/snapshot_controller_base.go
```diff
@@ -0,0 +1,439 @@
...
+	snapshotName := snapshotRefKey(content.Spec.VolumeSnapshotRef)
+	if snapshotName == "" {
+		glog.V(5).Infof("deleteContent[%q]: content not bound", content.Name)
+		return
+	}
+	// sync the vs when its vs is deleted.  Explicitly sync'ing the
+	// vs here in response to content deletion prevents the vs from
+	// waiting until the next sync period for its Release.
+	glog.V(5).Infof("deleteContent[%q]: scheduling sync of vs %s", content.Name, snapshotName)
+	ctrl.contentQueue.Add(snapshotName)
```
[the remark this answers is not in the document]
xing-yang: Fixed.

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+// frequently.
+//
+// This controller supports both dynamic snapshot creation and pre-bound snapshot.
+// In pre-bound mode, objects are created with pre-defined pointers: a VolumeSnapshot
+// points to a specific VolumeSnapshotContent and the VolumeSnapshotContent also
+// points back for this VolumeSnapshot.
+//
+// The dynamic snapshot creation is multi-step process: first controller triggers
+// snapshot creation though csi volume plugin which should return a snapshot after
+// it is created succesfully (however, the snapshot might not be ready to use yet if
```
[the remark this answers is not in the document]
xing-yang: Done

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+	glog.V(4).Infof("synchronizing VolumeSnapshotContent[%s]", content.Name)
+
+	// VolumeSnapshotContent is not bind to any VolumeSnapshot, this case rare and we just return err
+	if content.Spec.VolumeSnapshotRef == nil {
+		// content is not bind
+		glog.V(4).Infof("synchronizing VolumeSnapshotContent[%s]: VolumeSnapshotContent is not bound to any VolumeSnapshot", content.Name)
+		return fmt.Errorf("volumeSnapshotContent %s is not bound to any VolumeSnapshot", content.Name)
+	} else {
+		glog.V(4).Infof("synchronizing VolumeSnapshotContent[%s]: content is bound to snapshot %s", content.Name, snapshotRefKey(content.Spec.VolumeSnapshotRef))
+		// The VolumeSnapshotContent is reserved for a VolumeSNapshot;
```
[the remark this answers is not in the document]
xing-yang: Done

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+		// that VolumeSnapshot has not yet been bound to this VolumeSnapshotCent; the VolumeSnapshot sync will handle it.
```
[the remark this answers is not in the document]
xing-yang: Done

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+
+// syncContent deals with one key off the queue.  It returns false when it's time to quit.
+func (ctrl *csiSnapshotController) syncContent(content *crdv1.VolumeSnapshotContent) error {
+	glog.V(4).Infof("synchronizing VolumeSnapshotContent[%s]", content.Name)
+
+	// VolumeSnapshotContent is not bind to any VolumeSnapshot, this case rare and we just return err
+	if content.Spec.VolumeSnapshotRef == nil {
+		// content is not bind
+		glog.V(4).Infof("synchronizing VolumeSnapshotContent[%s]: VolumeSnapshotContent is not bound to any VolumeSnapshot", content.Name)
+		return fmt.Errorf("volumeSnapshotContent %s is not bound to any VolumeSnapshot", content.Name)
```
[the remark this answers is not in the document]
xing-yang: Will take a look.

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+		}
+	}
+	return nil
+}
+
+// syncSnapshot is the main controller method to decide what to do with a snapshot.
+// It's invoked by appropriate cache.Controller callbacks when a snapshot is
+// created, updated or periodically synced. We do not differentiate between
+// these events.
+// For easier readability, it is split into syncCompleteSnapshot and syncUncompleteSnapshot
```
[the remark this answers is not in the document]
xing-yang: Fixed.

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+			return err
+		}
+		return nil
+	}
+	obj, found, err := ctrl.contentStore.GetByKey(snapshot.Spec.SnapshotContentName)
+	if err != nil {
+		return err
+	}
+	if !found {
+		if _, err = ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotLost", "Bound snapshot has lost reference to VolumeSnapshotContent"); err != nil {
```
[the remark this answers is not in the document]
xing-yang: Fixed

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+	return nil
+}
+
+// updateSnapshotStatusWithEvent saves new snapshot.Status to API server and emits
+// given event on the snapshot. It saves the status and emits the event only when
+// the status has actually changed from the version saved in API server.
+// Parameters:
+//   snapshot - snapshot to update
+//   eventtype, reason, message - event to send, see EventRecorder.Event()
+func (ctrl *csiSnapshotController) updateSnapshotErrorStatusWithEvent(snapshot *crdv1.VolumeSnapshot, eventtype, reason, message string) (*crdv1.VolumeSnapshot, error) {
```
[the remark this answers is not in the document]
xing-yang: You are right. Removed.

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+		glog.V(4).Infof("updating VolumeSnapshot[%s] error status: cannot update internal cache %v", snapshotKey(snapshot), err)
+		return newSnapshot, err
+	}
+	// Emit the event only when the status change happens
+	ctrl.eventRecorder.Event(newSnapshot, eventtype, reason, message)
+
+	return newSnapshot, nil
+}
+
+func (ctrl *csiSnapshotController) updateSnapshotBoundWithEvent(snapshot *crdv1.VolumeSnapshot, eventtype, reason, message string) (*crdv1.VolumeSnapshot, error) {
```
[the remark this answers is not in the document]
xing-yang: Removed

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+// updateSnapshotStatusWithEvent saves new snapshot.Status to API server and emits
+// given event on the snapshot. It saves the status and emits the event only when
+// the status has actually changed from the version saved in API server.
+// Parameters:
+//   snapshot - snapshot to update
+//   eventtype, reason, message - event to send, see EventRecorder.Event()
+func (ctrl *csiSnapshotController) updateSnapshotErrorStatusWithEvent(snapshot *crdv1.VolumeSnapshot, eventtype, reason, message string) (*crdv1.VolumeSnapshot, error) {
+	glog.V(4).Infof("updateSnapshotStatusWithEvent[%s]", snapshotKey(snapshot))
+
+	if snapshot.Status.Error != nil && snapshot.Status.Ready == false {
```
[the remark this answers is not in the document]
xing-yang: Ok.  Removed this check so events will always be sent.

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,773 @@
...
+		return nil
+	}
+}
+
+func (ctrl *csiSnapshotController) syncUnboundSnapshot(snapshot *crdv1.VolumeSnapshot) error {
+	uniqueSnapshotName := snapshotKey(snapshot)
+	glog.V(4).Infof("syncSnapshot %s", uniqueSnapshotName)
+
+	// Snsapshot has errors during its creation. Controller will not try to fix it. Nothing to do.
+	if snapshot.Status.Error != nil {
```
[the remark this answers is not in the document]
xing-yang: Removed

pkg/controller/snapshot_controller.go
xing-yang: Done

pkg/controller/snapshot_controller.go
xing-yang: Done

pkg/controller/snapshot_controller.go
xing-yang: Done

pkg/controller/snapshot_controller.go

pkg/controller/snapshot_controller.go
xing-yang: Done

pkg/controller/snapshot_controller.go
xing-yang: Done

pkg/controller/snapshot_controller_base.go
xing-yang: Done

pkg/controller/snapshot_controller.go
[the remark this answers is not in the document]
xing-yang: Done

pkg/controller/snapshot_controller.go
xing-yang: Added event.

pkg/controller/snapshot_controller.go
xing-yang: Done

jsafrane:
One thing I forgot to mention, some parts use `V(4)` for detailed logging, some `V(5)`. I know it's mixed all over Kubernetes, but it could be better here since this is new code.

xing-yang:
@jsafrane Should we change all the detailed logging level to V(5)?

pkg/connection/connection.go
xing-yang: @sngchlko It's fixed in delete snapshot as well.  Please take a look.

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,766 @@
...
+	}
+
+	// Create VolumeSnapshotContent in the database
+	volumeRef, err := ref.GetReference(scheme.Scheme, volume)
+
+	snapshotContent := &crdv1.VolumeSnapshotContent{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: contentName,
+		},
+		Spec: crdv1.VolumeSnapshotContentSpec{
```
sngchlko: Please add `VolumeSnapshotClassName` in spec.
This will be used for getting secrets in `deleteSnapshotContentOperation`.

xing-yang: I don't see storageClassName gets set in PV Spec in external-provisioner: https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L410.  Do you know where that is set?

sngchlko: I mean `crdv1.VolumeSnapshotContentSpec`

sngchlko: I think you can get `storageClassName` in 'snapshot *crdv1.VolumeSnapshot`.

xing-yang: I know.  My change is modeled after the similar change in external-provisioner.  I didn't find where storageClassName is set in PV spec there.  That's why I didn't set it in Content spec.

When you test delete volume, do you get secrets that you set in StorageClass?

sngchlko: I got an error when I delete the snapshot, not the volume.
And I set `csiSnapshotSecretName` and `csiSnapshotSecretNamespace` in my `VolumeSnapshotClass`.
So I thought snapshot controller get secrets from `VolumeSnapshotClass`  for removing snapshots.
Why does the snapshot controller need storageClassName in PV for removing snapshot?

https://github.com/kubernetes-csi/external-snapshotter/blob/3e12fd6a36f694683ad628265da9b19998fd5878/pkg/controller/snapshot_controller.go#L573

xing-yang: This is how external provisioner gets secrets at delete volume time.  See here: https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L453

However, I don't see storageClassName gets set in PV spec when the volume is created in external provisioner.  So I wonder how secrets gets pass in there.  I'll check if it is set somewhere in the in-tree PV controller.

sngchlko: I just add a line and it works.
```
		Spec: crdv1.VolumeSnapshotContentSpec{
			VolumeSnapshotRef: &v1.ObjectReference{
				Kind:       "VolumeSnapshot",
				Namespace:  snapshot.Namespace,
				Name:       snapshot.Name,
				UID:        snapshot.UID,
				APIVersion: "v1alpha1",
			},
			PersistentVolumeRef: volumeRef,
			VolumeSnapshotSource: crdv1.VolumeSnapshotSource{
				CSI: &crdv1.CSIVolumeSnapshotSource{
					Driver:         driverName,
					SnapshotHandle: snapshotID,
					CreatedAt:      timestamp,
				},
			},
                        // I added
			VolumeSnapshotClassName: snapshot.Spec.VolumeSnapshotClassName,
		},
	}
```
We can get secrets from `VolumeSnapshotClass` as well.
If I wrong, please let me know.

xing-yang: Thanks for validating.  I still need to find out where it is set for PV spec.  Just want to make sure we are handling it consistently.

xing-yang: I found out where this is set for PV spec.  It is set when PV is bound to PVC:
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/persistentvolume/pv_controller.go#L1492

We should do the same thing here.  We can set VolumeSnapshotClass in the content spec when content is bound to snapshot in checkandBindSnapshotContent.  I'll fix that.

xing-yang: @sngchlko This is fixed now.  Can you give a try?  So what you have changed is correct.  We still need that.

sngchlko: Good!! It works. Thanks.

xing-yang: Awesome:).  Thanks!

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,842 @@
...
+	var found bool
+
+	objs := ctrl.contentStore.List()
+	for _, obj := range objs {
+		content := obj.(*crdv1.VolumeSnapshotContent)
+		if content.Spec.VolumeSnapshotRef != nil &&
+			content.Spec.VolumeSnapshotRef.Name == snapshot.Name &&
+			content.Spec.VolumeSnapshotRef.Namespace == snapshot.Namespace &&
+			content.Spec.VolumeSnapshotRef.UID == snapshot.UID {
+			found = true
```
wackxu: do we need check whether the snapshotClass is same?

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,842 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: contentName,
+		},
+		Spec: crdv1.VolumeSnapshotContentSpec{
+			VolumeSnapshotRef: &v1.ObjectReference{
+				Kind:       "VolumeSnapshot",
+				Namespace:  snapshot.Namespace,
+				Name:       snapshot.Name,
+				UID:        snapshot.UID,
+				APIVersion: "v1alpha1",
```
wackxu: it should be `snapshot.storage.k8s.io/v1alpha1`

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,842 @@
...
+// 3. Create the VolumeSnapshotContent object with the snapshot id information.
+// 4. Bind the VolumeSnapshot and VolumeSnapshotContent object
+func (ctrl *csiSnapshotController) createSnapshotOperation(snapshot *crdv1.VolumeSnapshot) (*crdv1.VolumeSnapshot, error) {
+	glog.Infof("createSnapshot: Creating snapshot %s through the plugin ...", snapshotKey(snapshot))
+
+	if snapshot.Status.Error != nil && !isControllerUpdateFailError(snapshot.Status.Error) {
+		glog.V(4).Infof("error is already set in snapshot, do not retry to create: %s", snapshot.Status.Error.Message)
+		return snapshot, nil
+	}
+	class, err := ctrl.GetClassFromVolumeSnapshot(snapshot)
```
wackxu: We may update snapshot when the snapshot  do not set VolumeSnapshotClassName field and use the default  snapshot class , in this case, the follow update will always be version conflict, so we`d better return the snapshot from the function.

pkg/controller/snapshot_controller_base.go
```diff
@@ -0,0 +1,468 @@
...
+		glog.V(5).Infof("snapshotWorker: snapshot namespace [%s] name [%s]", namespace, name)
+		if err != nil {
+			glog.V(4).Infof("error getting namespace & name of snapshot %q to get snapshot from informer: %v", key, err)
+			return false
+		}
+		snapshot, err := ctrl.snapshotLister.VolumeSnapshots(namespace).Get(name)
+		if err == nil {
+			// The volume snapshot still exists in informer cache, the event must have
+			// been add/update/sync
+			if ctrl.shouldProcessSnapshot(snapshot) {
```
jingxu97: here shouldProcessSnapshot might update the snapshot, then this function should return the new version of the snapshot so that the following updateSnapshot can use the new version

xing-yang: Fixed.

xing-yang:
Addressed all review comments.

pkg/controller/snapshot_controller.go
```diff
@@ -0,0 +1,856 @@
...
+//   eventtype, reason, message - event to send, see EventRecorder.Event()
+func (ctrl *csiSnapshotController) updateSnapshotErrorStatusWithEvent(snapshot *crdv1.VolumeSnapshot, eventtype, reason, message string) error {
+	glog.V(5).Infof("updateSnapshotStatusWithEvent[%s]", snapshotKey(snapshot))
+
+	if snapshot.Status.Error != nil && snapshot.Status.Error.Message == message {
+		glog.V(4).Infof("updateSnapshotStatusWithEvent[%s]: the same error %v is already set", snapshot.Name, snapshot.Status.Error)
+		return nil
+	}
+	snapshotClone := snapshot.DeepCopy()
+	if snapshot.Status.Error == nil {
```
jingxu97: remove this line to allow new discovered error to show up.

xing-yang: Done

jsafrane:
lgtm-ish, I went through the code and tests and it looks usable. IMO, it's good for alpha, it needs serious testing.

PR #12, opened 2018-08-22
Add UT for Snapshot controller

wackxu:
/assign @xing-yang @jingxu97 @saad-ali @jsafrane for review

based on https://github.com/kubernetes-csi/external-snapshotter/pull/7


pkg/controller/snapshot_create_test.go
```diff
@@ -0,0 +1,418 @@
...
+						Type:    csi.SnapshotStatus_READY,
+						Details: "success",
+					},
+				},
+			},
+			errors: noerrors,
+			test:   testSyncSnapshot,
+		},
+		//{
+		//	name:                  "6-5 - successful create snapshot with default snapshot class",
```
jsafrane: Why is this commented out? I don't see any other test for default snapshot class.

wackxu: When I add test, I found this test case failed due to this bug https://github.com/kubernetes-csi/external-snapshotter/pull/7#pullrequestreview-148480829,  I can uncomment it since the bug has been fixed.

jsafrane: Ok, then please add a TODO comment to enable the test when the code is fixed.

pkg/controller/framework_test.go
```diff
@@ -0,0 +1,1233 @@
...
+
+	case action.Matches("get", "volumesnapshotcontents"):
+		name := action.(core.GetAction).GetName()
+		content, found := r.contents[name]
+		if found {
+			glog.V(4).Infof("GetVolume: found %s", content.Name)
+			return true, content, nil
+		} else {
+			glog.V(4).Infof("GetVolume: content %s not found", name)
+			return true, nil, fmt.Errorf("Cannot find content %s", name)
```
xing-yang: s/Cannot/cannot

Same comments in other similar places.

pkg/connection/connection_test.go
```diff
@@ -1 +1,772 @@
...
+						Type: &csi.ControllerServiceCapability_Rpc{
+							Rpc: &csi.ControllerServiceCapability_RPC{
+								Type: csi.ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME,
+							},
+						},
+					},
+				},
+			},
+			expectError: false,
+		},
```
xing-yang: Add a test for the case where the plugin supports create_delete_volume and create_delete_snapshot but not list_snapshots

wackxu:
@xing-yang @jingxu97 @saad-ali Updated, PTAL

pkg/controller/snapshot_controller_test.go
```diff
@@ -0,0 +1,91 @@
...
+		t.Errorf("expected content in the cache, got different object instead: %#v", contentObj)
+	}
+
+	if ret {
+		if content.ResourceVersion != version {
+			t.Errorf("expected content with version %s in the cache, got %s instead", version, content.ResourceVersion)
+		}
+	} else {
+		if content.ResourceVersion == version {
+			t.Errorf("expected content with version other than %s in the cache, got %s instead", version, content.ResourceVersion)
```
lpabon: we probably should be using a golang test framework like testify. Writing `t.Errorf(..)` every time is a PAIN! Nice job though.

xing-yang:
Checked with Jing and she is okay with merging this too.

PR #15, opened 2018-08-30
 Switch to use TypedLocalObjectReference in core API

xing-yang:
Now that the DataSource core API change is merged, this PR modifies the snapshot API and controller to use TypedLocalObjectReference from core API.


xing-yang:
Made the changes based on the master branch in https://github.com/kubernetes-csi/external-snapshotter/pull/16.  Closing this PR.

PR #16, opened 2018-08-30
Switch to use TypedLocalObjectReference in core API

xing-yang:
This PR modifies the snapshot API and controller to use TypedLocalObjectReference from core API.


pkg/controller/framework_test.go
```diff
@@ -806,9 +807,10 @@ func newSnapshot(name, className, boundToContent, snapshotUID, claimName string,
 			SelfLink:        "/apis/snapshot.storage.k8s.io/v1alpha1/namespaces/" + testNamespace + "/volumesnapshots/" + name,
 		},
 		Spec: crdv1.VolumeSnapshotSpec{
-			Source: &crdv1.TypedLocalObjectReference{
-				Name: claimName,
-				Kind: "PersistentVolumeClaim",
+			Source: &v1.TypedLocalObjectReference{
+				Name:     claimName,
+				Kind:     "PersistentVolumeClaim",
+				APIGroup: &apiGrp,
```
jingxu97: I think we make APIGroup optional. If it is an empty string, we don't need to specify it here. Could you try and confirm it?

xing-yang: Yes.  We don't need to specify it.  I'll remove this.

xing-yang:
@jingxu97 Addressed your comments.  Please take a look.  Thanks.

pkg/controller/snapshot_controller.go
```diff
@@ -824,13 +825,19 @@ func (ctrl *csiSnapshotController) SetDefaultSnapshotClass(snapshot *crdv1.Volum
...
+	if snapshot.Spec.Source.Kind != pvcKind {
+		return nil, fmt.Errorf("the snapshot source is not the right type. Expected %s, Got %v", pvcKind, snapshot.Spec.Source.Kind)
 	}
 	pvcName := snapshot.Spec.Source.Name
 	if pvcName == "" {
 		return nil, fmt.Errorf("the PVC name is not specified in snapshot %s", snapshotKey(snapshot))
 	}
+	if snapshot.Spec.Source.APIGroup != nil && *(snapshot.Spec.Source.APIGroup) != apiGroup {
+		return nil, fmt.Errorf("the snapshot source does not have the right APIGroup. Expected %s, Got %s", apiGroup, *(snapshot.Spec.Source.APIGroup))
+	}
```
jingxu97: It will print out like "Expected , Got xxx", I wonder it may be better to print out like
Expected "", Got "xxx"? or Expected empty string, Got "xxx"?

PR #17, opened 2018-08-31
Add example yaml files for deployment and testing

xing-yang:
This PR adds example yaml files for deployment and testing of the snapshot feature.

Images will be updated when they are released.

### DIFF
--- a/deploy/kubernetes/setup-csi-snapshotter.yaml
+++ b/deploy/kubernetes/setup-csi-snapshotter.yaml
@@ -1,0 +1,144 @@
+# This YAML file contains all API objects that are necessary to run external
+# CSI snapshotter.
+#
+# In production, this needs to be in separate files, e.g. service account and
+# role and role binding needs to be created once, while stateful set may
+# require some tuning.
+#
+# In addition, hostpath CSI driver is hardcoded as the CSI driver.
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: csi-snapshotter
+ 
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: external-snapshotter-runner
+rules:
+  - apiGroups: [""]
+    resources: ["persistentvolumes"]
+    verbs: ["get", "list", "watch", "create", "delete"]
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["list", "watch", "create", "update", "patch"]
+  - apiGroups: [""]
+    resources: ["endpoints"]
+    verbs: ["list", "watch", "create", "update", "delete", "get"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get", "list"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshotclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshotcontents"]
+    verbs: ["create", "get", "list", "watch", "update", "delete"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshots"]
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
+    verbs: ["create", "list", "watch", "delete"]
+ 
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: csi-snapshotter-role
+subjects:
+  - kind: ServiceAccount
+    name: csi-snapshotter
+    namespace: default
+roleRef:
+  kind: ClusterRole
+  name: external-snapshotter-runner
+  apiGroup: rbac.authorization.k8s.io
+ 
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: csi-snapshotter
+  labels:
+    app: csi-snapshotter
+spec:
+  selector:
+    app: csi-snapshotter
+  ports:
+    - name: dummy
+      port: 12345
+
+---
+kind: StatefulSet
+apiVersion: apps/v1
+metadata:
+  name: csi-snapshotter
+spec:
+  serviceName: "csi-snapshotter"
+  replicas: 1
+  selector:
+    matchLabels:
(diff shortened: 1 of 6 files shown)


examples/pvc.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: hpvc
+spec:
+  storageClassName: csi-hostpath-sc
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+
```
wackxu: remove blank line

examples/restore.yaml
```diff
@@ -0,0 +1,16 @@
...
+  dataSource:
+    name: new-snapshot-demo 
+    kind: VolumeSnapshot
+    apiGroup: snapshot.storage.k8s.io
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+
```
wackxu: ditto

examples/storageclass.yaml
```diff
@@ -0,0 +1,10 @@
+metadata:
```
wackxu: storageclass?

deploy/kubernetes/setup-csi-snapshotter.yaml
```diff
@@ -0,0 +1,146 @@
...
+ 
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: external-snapshotter-runner
+rules:
+  - apiGroups: [""]
+    resources: ["persistentvolumes"]
+    verbs: ["get", "list", "watch", "create", "delete"]
```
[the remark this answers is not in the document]
jingxu97: but the error is for list PVCs, you don't need create/delete ?

deploy/kubernetes/setup-csi-snapshotter.yaml
```diff
@@ -0,0 +1,148 @@
...
+    verbs: ["get", "list"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshotclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshotcontents"]
+    verbs: ["create", "get", "list", "watch", "update", "delete"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshots"]
+    verbs: ["create", "get", "list", "watch", "update", "delete"]
```
jingxu97: no need to have create?

deploy/kubernetes/setup-csi-snapshotter.yaml
```diff
@@ -0,0 +1,148 @@
...
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshots/status"]
```
jingxu97: with the above permission on volumesnapshots, I don't think you need to put /status here?

deploy/kubernetes/setup-csi-snapshotter.yaml
```diff
@@ -0,0 +1,148 @@
...
+            - name: ADDRESS
+              value: /csi/csi.sock
+          imagePullPolicy: "IfNotPresent"
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+        - name: csi-snapshotter
+          image: xyang105/csi:csi-snapshotter
+          args:
+            - "--snapshotter=csi-hostpath"
```
jingxu97: I think we don't need to specify this driver name any more.

deploy/kubernetes/setup-csi-snapshotter.yaml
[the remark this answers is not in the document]
xing-yang: Looks like I didn't read the comments carefully:).  I'll verify.

wackxu:
@xing-yang @jingxu97 I saw there are in-tree clusterRole for external-provisioner and external-attacher, We should also add this for snapshotter, if it is ok, I can add this.

deploy/kubernetes/setup-csi-snapshotter.yaml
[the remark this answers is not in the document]
xing-yang: Verified that I need to keep create/delete too:

I0915 03:26:41.931543       1 controller.go:1113] provision "default/hpvc" class "csi-hostpath-sc": failed to save persistentvolume "pvc-1e1481d2-b897-11e8-944c-000c2967769a": persistentvolumes is forbidden: User "system:serviceaccount:default:csi-snapshotter" cannot create resource "persistentvolumes" in API group "" at the cluster scope

deploy/kubernetes/setup-csi-snapshotter.yaml
xing-yang: We use this in SetDefaultSnapshotClass. Do you mean this is no longer needed if we have admission control?  If that is what you meant, we still need it now until it is set automatically.

To be consistent with the provisioner, I think we should set it too.

deploy/kubernetes/setup-csi-snapshotter.yaml
xing-yang: Done

deploy/kubernetes/setup-csi-snapshotter.yaml
xing-yang: Done

xing-yang:
@wackxu Can you add a link to those here?

wackxu:
@xing-yang When I add in-tree test, I found this, I have add the code and test code.

https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go#L460-L469

https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go#L504-L513

deploy/kubernetes/setup-csi-snapshotter.yaml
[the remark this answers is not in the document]
wackxu: @xing-yang Yes, you are right,  we deploy external-provisioner together with external-snapshotter, so we need this.

xing-yang:
I don't know if they would let us add this for an out-of-tree CRD.  The one they added is CRD used by kubelet so it is different.

Jing, what do you think?

deploy/kubernetes/setup-csi-snapshotter.yaml
jingxu97: https://github.com/kubernetes-csi/external-snapshotter/blob/master/cmd/csi-snapshotter/main.go#L118
snapshotter driver name is returned by this function. User does not need to set it manually.

xing-yang: Sure.  Will remove it.

xing-yang: Fixed

jingxu97:
For ClusterRole question, since snapshots are CRD, currently we don't need to have in-tree clusterRole for them.

PR #19, opened 2018-09-05
Fix updating restore size issue

jingxu97:
This PR fixed the issue of not updating the snapshot restore size after
snapshot is created. Before snapshot is ready, the returned size might
not be accurate. So we need to keep updating the snapshot size during
checking the snapshot status.

### DIFF
--- a/pkg/connection/connection_test.go
+++ b/pkg/connection/connection_test.go
@@ -658,6 +658,7 @@ func TestDeleteSnapshot(t *testing.T) {
 func TestGetSnapshotStatus(t *testing.T) {
 	defaultID := "testid"
 	createdAt := time.Now().UnixNano()
+	size := int64(1000)
 
 	defaultRequest := &csi.ListSnapshotsRequest{
 		SnapshotId: defaultID,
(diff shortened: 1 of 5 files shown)


xing-yang:
Can you please fix the CI errors?

jingxu97:
fixed the errors. PTAL.

pkg/connection/connection_test.go
```diff
@@ -754,6 +757,9 @@ func TestGetSnapshotStatus(t *testing.T) {
 		if test.expectCreateAt != createTime {
 			t.Errorf("test %q: expected createTime: %v, got: %v", test.name, test.expectCreateAt, createTime)
 		}
+		if test.expectSize != size {
+			t.Errorf("test %q: expected size: %v, got: %v", test.name, test.expectSize, createTime)
```
wackxu: createTime --> size

jingxu97:
close this one because https://github.com/kubernetes-csi/external-snapshotter/pull/24 is opened.

pkg/connection/connection_test.go
jingxu97: fixed by PR https://github.com/kubernetes-csi/external-snapshotter/pull/24

PR #24, opened 2018-09-07
Fix updating restore size issue

jingxu97:
This PR fixed the issue of not updating the snapshot restore size after
snapshot is created. Before snapshot is ready, the returned size might
not be accurate. So we need to keep updating the snapshot size during
checking the snapshot status.


wackxu:
LGTM

PR #27, opened 2018-09-17
Fix restore size type in snapshot content

jingxu97:
Change the restore size to int64 for snapshot content. Also add the code
to update the restore size for snapshot content.

### DIFF
--- a/pkg/controller/snapshot_controller.go
+++ b/pkg/controller/snapshot_controller.go
@@ -170,7 +170,7 @@ func (ctrl *csiSnapshotController) syncReadySnapshot(snapshot *crdv1.VolumeSnaps
 			return fmt.Errorf("Cannot convert object from snapshot content store to VolumeSnapshotContent %q!?: %#v", snapshot.Spec.SnapshotContentName, obj)
 		}
 
-		glog.V(5).Infof("syncCompleteSnapshot[%s]: VolumeSnapshotContent %q found", snapshotKey(snapshot), content.Name)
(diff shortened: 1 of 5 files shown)


pkg/controller/snapshot_controller.go
```diff
@@ -637,6 +641,28 @@ func (ctrl *csiSnapshotController) bindandUpdateVolumeSnapshot(snapshotContent *
 	return snapshotCopy, nil
 }
 
+// updateSnapshotContentSize update the restore size for snapshot content
+func (ctrl *csiSnapshotController) updateSnapshotContentSize(content *crdv1.VolumeSnapshotContent, size int64) error {
+	if content.Spec.VolumeSnapshotSource.CSI == nil || size == 0 {
```
xing-yang: size <= 0

jingxu97: done

pkg/controller/snapshot_controller.go
```diff
@@ -637,6 +641,28 @@ func (ctrl *csiSnapshotController) bindandUpdateVolumeSnapshot(snapshotContent *
...
+		return nil
+	}
+	if content.Spec.VolumeSnapshotSource.CSI.RestoreSize != nil && *content.Spec.VolumeSnapshotSource.CSI.RestoreSize > 0 {
```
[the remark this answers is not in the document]
jingxu97: actually no, changed it to check whether the updated size is equal to the old one, 
*content.Spec.VolumeSnapshotSource.CSI.RestoreSize == size

PR #30, opened 2018-09-19
Remove unnecessary verbs for endpoints

xing-yang:
This PR removes unnecessary verbs for "endpoints" resource in the deployment yaml file.


xing-yang:
I got the following error after removing "create" when I was testing on a completely new setup.  So I'm going to keep all the verbs and close this PR now.

E0921 19:31:04.630865       1 leaderelection.go:228] error initially creating leader election record: endpoints is forbidden: User "system:serviceaccount:default:csi-snapshotter" cannot create resource "endpoints" in API group "" in the namespace "default"

PR #36, opened 2018-09-27
Add Admission webhook

jingxu97:
This PR adds the admission webhook for verifying snapshot API objects
and also setting the default snapshot class if it is not specified by the
users

PR #39, opened 2018-10-10
Add Finalizer for VolumeSnapshot and VolumeSnapshotContent

xing-yang:
This PR adds a Finalizer for VolumeSnapshotContent.
If the VolumeSnapshotContent is bound to a VolumeSnapshot,
the VolumeSnapshotContent is being used and cannot be
deleted.
This PR also adds a Finalizer for VolumeSnapshot.
If a volume is being created from the snapshot,
the VolumeSnapshot is being used and cannot be deleted.


xing-yang:
@jingxu97 @wackxu PTAL

pkg/controller/snapshot_controller.go
```diff
@@ -410,6 +454,48 @@ func IsSnapshotBound(snapshot *crdv1.VolumeSnapshot, content *crdv1.VolumeSnapsh
...
+		}
+	}
+
+	glog.V(5).Infof("isSnapshotContentBeingUsed: Snapshot content %s is not being used", content.Name)
+	return false
+}
+
+// isVolumeBeingCreatedFromSnapshot checks if an volume is being created from the snapshot.
+func (ctrl *csiSnapshotController) isVolumeBeingCreatedFromSnapshot(snapshot *crdv1.VolumeSnapshot) bool {
+	pvcList, err := ctrl.client.CoreV1().PersistentVolumeClaims(snapshot.Namespace).List(metav1.ListOptions{})
```
wackxu: we should use `pvcLister` to reduce the  server pressure instead of using client

pkg/controller/snapshot_controller.go
```diff
@@ -85,6 +89,26 @@ const IsDefaultSnapshotClassAnnotation = "snapshot.storage.kubernetes.io/is-defa
...
+		if !isUsed {
+			glog.V(5).Infof("syncContent: Remove Finalizer for VolumeSnapshotContent[%s]", content.Name)
+			return ctrl.removeContentFinalizer(content)
+		}
+	}
+
+	if needToAddContentFinalizer(content) {
+		// Content is not being deleted -> it should have the finalizer. The
+		// finalizer should be added by admission plugin, this is just to add
+		// the finalizer to old volume snapshot contents that were created before
```
wackxu: for now, we do not have admission plugin to add the finalizer for  snapshot content.

pkg/controller/snapshot_controller.go
```diff
@@ -154,6 +178,26 @@ func (ctrl *csiSnapshotController) syncContent(content *crdv1.VolumeSnapshotCont
...
+			glog.V(5).Infof("syncSnapshot: Remove Finalizer for VolumeSnapshot[%s]", snapshot.Name)
+			return ctrl.removeSnapshotFinalizer(snapshot)
+		}
+	}
+
+	if needToAddSnapshotFinalizer(snapshot) {
+		// Snapshot is not being deleted -> it should have the finalizer. The
+		// finalizer should be added by admission plugin, this is just to add
+		// the finalizer to old volume snapshots that were created before
+		// the admission plugin was enabled.
```
wackxu: for now, we do not have admission plugin to add the finalizer for snapshot.

wackxu:
1.Since we are adding admission webhook https://github.com/kubernetes-csi/external-snapshotter/pull/36,  we can also add a `MutatingWebhook` to add finalizer for snapshot and snapshot content when they create.

2.Can we make the protection feature in a separate controller, like  k8s pv_protection_controller, it is convenient for us to test and we should obey single responsibility principle and make the snapshot controller to do the specialized task.

pkg/controller/snapshot_controller.go
xing-yang: Good suggestion.  I think a separate PR should be submitted to add pvcLister to the snapshot controller first, and after that this can be updated to use that.

pkg/controller/snapshot_controller.go
xing-yang: Will remove it for now.

pkg/controller/snapshot_controller.go
xing-yang: Will remove it for now.

xing-yang:
We'll add a webhook to add finalizer for snapshot after #36 is merged.

Jing and I discussed about whether to implement finalizer in a separate controller, similar to pv_protection_controller, but decided to keep it simple and add it to the same controller.  There's a trade-off between separating logic into different controllers and the overhead of maintaining multiple controllers.  PV controller is very big and adding more to it will make it more complicated, while the snapshot controller is much simpler and putting the logic in the same controller makes it easier to maintain.

pkg/controller/snapshot_controller.go
xing-yang: Removed

pkg/controller/snapshot_controller.go
xing-yang: Removed

pkg/controller/snapshot_controller.go
```diff
@@ -154,6 +175,23 @@ func (ctrl *csiSnapshotController) syncContent(content *crdv1.VolumeSnapshotCont
 func (ctrl *csiSnapshotController) syncSnapshot(snapshot *crdv1.VolumeSnapshot) error {
 	glog.V(5).Infof("synchonizing VolumeSnapshot[%s]: %s", snapshotKey(snapshot), getSnapshotStatusForLogging(snapshot))
 
+	if isSnapshotDeletionCandidate(snapshot) {
+		// Volume snapshot should be deleted. Check if it's used
+		// and remove finalizer if it's not.
+		// Check if a volume is being created from snapshot.
+		isUsed := ctrl.isVolumeBeingCreatedFromSnapshot(snapshot)
+		if !isUsed {
+			glog.V(5).Infof("syncSnapshot: Remove Finalizer for VolumeSnapshot[%s]", snapshot.Name)
```
wackxu: `snapshotKey(snapshot)` maybe better for logging

pkg/controller/snapshot_controller.go
```diff
@@ -154,6 +175,23 @@ func (ctrl *csiSnapshotController) syncContent(content *crdv1.VolumeSnapshotCont
...
+			return ctrl.removeSnapshotFinalizer(snapshot)
+		}
+	}
+
+	if needToAddSnapshotFinalizer(snapshot) {
+		// Snapshot is not being deleted -> it should have the finalizer.
+		glog.V(5).Infof("syncSnapshot: Add Finalizer for VolumeSnapshot[%s]", snapshot.Name)
```
wackxu: ditto

pkg/controller/snapshot_controller.go
```diff
@@ -410,6 +448,48 @@ func IsSnapshotBound(snapshot *crdv1.VolumeSnapshot, content *crdv1.VolumeSnapsh
...
+
+	glog.V(5).Infof("isSnapshotContentBeingUsed: Snapshot content %s is not being used", content.Name)
+	return false
+}
+
+// isVolumeBeingCreatedFromSnapshot checks if an volume is being created from the snapshot.
+func (ctrl *csiSnapshotController) isVolumeBeingCreatedFromSnapshot(snapshot *crdv1.VolumeSnapshot) bool {
+	pvcList, err := ctrl.client.CoreV1().PersistentVolumeClaims(snapshot.Namespace).List(metav1.ListOptions{})
+	if err != nil {
+		glog.Errorf("Failed to retrieve PVCs from the API server to check if volume snapshot %s is being used by a volume: %q", snapshot.Name, err)
```
wackxu: ditto

pkg/controller/snapshot_controller.go
```diff
@@ -410,6 +448,48 @@ func IsSnapshotBound(snapshot *crdv1.VolumeSnapshot, content *crdv1.VolumeSnapsh
...
+			if pvc.Spec.DataSource.Kind == snapshotKind && *(pvc.Spec.DataSource.APIGroup) == snapshotAPIGroup {
+				if pvc.Status.Phase == v1.ClaimPending {
+					// A volume is being created from the snapshot
+					glog.Infof("isVolumeBeingCreatedFromSnapshot: volume %s is being created from snapshot %s", pvc.Name, pvc.Spec.DataSource.Name)
+					return true
+				}
+			}
+		}
+	}
+	glog.V(5).Infof("isVolumeBeingCreatedFromSnapshot: no volume is being created from snapshot %s", snapshot.Name)
```
wackxu: ditto

pkg/controller/snapshot_controller.go
jingxu97: could you create an issue for it so that we won't forget?

jingxu97:
lgtm

pkg/controller/snapshot_controller.go
xing-yang: Sure.  Opened issue #70

pkg/controller/snapshot_controller.go
xing-yang: Done

pkg/controller/snapshot_controller.go
xing-yang: Done

pkg/controller/snapshot_controller.go
xing-yang: Done

pkg/controller/snapshot_controller.go
xing-yang: Done

PR #40, opened 2018-10-10
 Add volumesnapshot content deletion policy

jingxu97:
This PR adds DeletionPolicy which describes a policy for end-of-life maintenance of volume snapshot contents.
There are two types of policies, delete and retain. By default, the
policy is delete. In this case, when volume snaphsot is deleted, the
volume snapshot content for it should be deleted, as well as the
physical snapshot. If the policy is set the retain, the volume snapshot
content will remain even if the volume snapshot it binds to is deleted.


pkg/controller/snapshot_controller.go
```diff
@@ -125,9 +125,21 @@ func (ctrl *csiSnapshotController) syncContent(content *crdv1.VolumeSnapshotCont
...
+		if content.Spec.DeletionPolicy != nil {
+			switch *content.Spec.DeletionPolicy {
+			case crdv1.VolumeSnapshotContentRetain:
+				glog.V(4).Infof("VolumeSnapshotContent[%s]: policy is Retain, nothing to do", content.Name)
+		
+			case crdv1.VolumeSnapshotContentDelete:
+				glog.V(4).Infof("VolumeSnapshotContent[%s]: policy is Delete", content.Name)
+				ctrl.deleteSnapshotContent(content)
+			default:
+				// Unknown PersistentVolumeReclaimPolicy
```
xing-yang: s/PersistentVolumeReclaimPolicy/VolumeSnapshotReclaimPolicy

pkg/controller/snapshot_controller.go
```diff
@@ -125,9 +125,21 @@ func (ctrl *csiSnapshotController) syncContent(content *crdv1.VolumeSnapshotCont
...
+				ctrl.eventRecorder.Event(content, v1.EventTypeWarning, "SnapshotUnknownDeletionPolicy", "Volume Snapshot Content has unrecognized deletion policy")
```
xing-yang: So the default policy is actually the same as Retain?

pkg/controller/snapshot_delete_test.go
```diff
@@ -128,10 +129,21 @@ func TestDeleteSync(t *testing.T) {
 			expectedDeleteCalls: []deleteCall{{"sid1-1", nil, nil}},
 			test:                testSyncContent,
 		},
+		{
+			name:                "1-1 - content with retain policy will not be deleted if it is bound to a non-exist snapshot and also has a snapshot uid specified",
```
wackxu: the number of test case is the same as bellow, we can put this and "2-1" case after the "1-11" case and rename it "1-12"

pkg/controller/snapshot_controller.go
jingxu97: fixed

pkg/controller/snapshot_controller.go
jingxu97: no, it should be Delete. I fixed the logic here.

pkg/apis/volumesnapshot/v1alpha1/types.go
```diff
@@ -241,3 +249,15 @@ type CSIVolumeSnapshotSource struct {
...
+
+// DeletionPolicy describes a policy for end-of-life maintenance of volume snapshot contents
+type DeletionPolicy string
+
+const (
+	// VolumeSnapshotContentDelete means the snapshot content will be deleted from Kubernetes on release from its volume snapshot.
+	// The volume plugin must support Deletion.
+	VolumeSnapshotContentDelete DeletionPolicy = "Delete"
+	// VolumeSnapshotContentRetain means the volume will be left in its current phase (Released) for manual reclamation by the administrator.
+	// The default policy is Retain.
```
wackxu: update the comment here since the default policy `Delete`

jingxu97: Sorry actually here we should set default policy for VolumeSnapshotContent is Retain.
Similar to Persistent Volumes, if VolumeSnapshotContent is created manually, the default set is Retain. If VolumeSnapshotContent is dynamically created through VolumeSnapshot, the default policy is Delete which is set through VolumeSnaphshotClass.

wackxu:
The CI is failed, could you update failed UT?

pkg/controller/snapshot_delete_test.go
jingxu97: fixed

xing-yang:
@jingxu97 There's CI failure.  Can you check?

pkg/controller/snapshot_controller.go
xing-yang: For default, it won't call deleteSnapshotContent.  So default is still equal to Retain?  You said it should be the same as Delete earlier.

jingxu97: there are two cases. 
For dynamic creation, we add deletePolicy when creating snapshotContent. Check "createSnapshotOperation" function in controller

For static binding case, if no deletePolicy is set, by default it is retain.

wackxu:
LGTM

xing-yang:
Looks good now.  Will merge this as soon as RC3 is cut.

PR #46, opened 2018-10-19
Fix issue in snapshot worker

jingxu97:
During processing snapshot, if snapshot should be deleted or snapshot
status is already ready, there is no need to check and update
storageclass.

### DIFF
--- a/pkg/controller/snapshot_controller_base.go
+++ b/pkg/controller/snapshot_controller_base.go
@@ -238,10 +238,7 @@ func (ctrl *csiSnapshotController) snapshotWorker() {
 			glog.Errorf("expected vs, got %+v", vsObj)
 			return false
 		}
-		newSnapshot, err := ctrl.checkAndUpdateSnapshotClass(snapshot)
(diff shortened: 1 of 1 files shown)


pkg/controller/snapshot_controller_base.go
```diff
@@ -238,10 +238,7 @@ func (ctrl *csiSnapshotController) snapshotWorker() {
 			glog.Errorf("expected vs, got %+v", vsObj)
 			return false
 		}
-		newSnapshot, err := ctrl.checkAndUpdateSnapshotClass(snapshot)
```
xing-yang: This check is still needed because some CSI plugins need to use secrets stored in SnapshotClass to delete the snapshot.

jingxu97: Then deletion will fail with some error msg, right?
The reason I don't want to put the check here is that if SnapshotClass is deleted, the snapshot status become not ready and user cannot use the snapshot to create volume.

PR #47, opened 2018-10-28
Add Delete Volume Finalizer

xing-yang:
This PR adds Finalizer on the snapshot source PVC to prevent it
from being deleted when a snapshot is being created from it.

### DIFF
--- a/pkg/controller/snapshot_controller.go
+++ b/pkg/controller/snapshot_controller.go
@@ -192,11 +192,19 @@ func (ctrl *csiSnapshotController) syncSnapshot(snapshot *crdv1.VolumeSnapshot) 
 		return ctrl.addSnapshotFinalizer(snapshot)
 	}
 
+	klog.V(5).Infof("syncSnapshot[%s]: check if we should remove finalizer on snapshot source and remove it if we can", snapshotKey(snapshot))
+	// Check if we should remove finalizer on snapshot source and remove it if we can.
+	errFinalizer := ctrl.checkandRemoveSnapshotSourceFinalizer(snapshot)
+	if errFinalizer != nil {
+		klog.Errorf("error check and remove snapshot source finalizer for snapshot [%s]: %v", snapshot.Name, errFinalizer)
+		// Log an event and keep the original error from syncUnready/ReadySnapshot
+		ctrl.eventRecorder.Event(snapshot, v1.EventTypeWarning, "ErrorSnapshotSourceFinalizer", "Error check and remove PVC Finalizer for VolumeSnapshot")
+	}
+
 	if !snapshot.Status.ReadyToUse {
 		return ctrl.syncUnreadySnapshot(snapshot)
 	}
 	return ctrl.syncReadySnapshot(snapshot)
-
 }
 
 // syncReadySnapshot checks the snapshot which has been bound to snapshot content successfully before.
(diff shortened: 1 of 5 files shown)


xing-yang:
@wackxu PTAL.  Thanks.

wackxu:
I think we should not  depend on annotation `SnapshotsInCreationCountAnnotation` on pvc to judge  the pvc is being in used by snapshot.
1、we should also consider snapshot that user created with non-empty `SnapshotContentName` .
2、when a snapshot is deleted, we should also decrease the `SnapshotsInCreationCountAnnotation`
3、if `SnapshotsInCreationCountAnnotation` is increased when create snapshot but then the controller is  failed and another controller get the lead election, then the  `SnapshotsInCreationCountAnnotation`  will increased again.

wackxu:
We can traverse the snapshot in the cache to determine if pvc is being used. like the pvc_protection_controller in the k8s.
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/pvcprotection/pvc_protection_controller.go#L210-L237

xing-yang:
> We can traverse the snapshot in the cache to determine if pvc is being used. like the pvc_protection_controller in the k8s.
> https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/pvcprotection/pvc_protection_controller.go#L210-L237

This seems to be a good idea.  I'll explore it.  Thanks.

xing-yang:
Rebased.

xing-yang:
@wackxu Addressed your comments.  Use snapshotLister to check if a snapshot is being created from a PVC.

xing-yang:
@jsafrane PTAL.  Thanks.

pkg/controller/snapshot_controller.go
```diff
@@ -1068,3 +1089,117 @@ func (ctrl *csiSnapshotController) removeSnapshotFinalizer(snapshot *crdv1.Volum
 	klog.V(5).Infof("Removed protection finalizer from volume snapshot %s", snapshotKey(snapshot))
 	return nil
 }
+
+// addSnapshotSourceFinalizer adds a Finalizer for VolumeSnapshot Source PVC.
+func (ctrl *csiSnapshotController) addSnapshotSourceFinalizer(snapshot *crdv1.VolumeSnapshot) error {
```
jsafrane: I think you can merge `addSnapshotSourceFinalizer` and `needToAddSnapshotSourceFinalizer` together into e.g. `ensureSnapshotSourceFinalizer` and save one `getClaimFromVolumeSnapshot` and couple of common lines of code.

xing-yang: Sure.  Merged the two functions into one.

xing-yang:
Addressed one comment.  Will add tests soon.

xing-yang:
Hi @jsafrane, I added tests. PTAL.

xing-yang:
e2e test failed because external-snapshotter does not have "update" rbac rule for updating PVC object to add a Finalizer:

https://github.com/kubernetes/kubernetes/blob/master/test/e2e/testing-manifests/storage-csi/external-snapshotter/rbac.yaml

Error message in e2e tests:
```
Apr 25 02:24:54.933: INFO: At 2019-04-25 02:14:52 +0000 UTC - event for snapshot-xxdgm: {csi-snapshotter csi-hostpath } SnapshotCreationFailed: Failed to create snapshot: snapshot controller failed to update pvc-flxrc on API server: persistentvolumeclaims "pvc-flxrc" is forbidden: User "system:serviceaccount:default:csi-snapshotter" cannot update resource "persistentvolumeclaims" in API group "" in the namespace "provisioning-7646"
```

PR #48, opened 2018-11-01
add snapshotter args for deploy csi-snapshotter

wackxu:
@xing-yang @jingxu97


deploy/kubernetes/setup-csi-snapshotter.yaml
```diff
@@ -110,6 +110,7 @@ spec:
         - name: csi-snapshotter
           image: quay.io/k8scsi/csi-snapshotter:v0.4.0
           args:
+            - "--snapshotter=csi-hostpath"
```
xing-yang: This line is not required so we removed it.  It will be detected automatically.  Have you run into any problem without this line?

wackxu: No, just want to keep with provisioner and make the args explicitly

PR #53, opened 2018-11-05
deploy: split out RBAC definitions

pohly:
Splitting out the RBAC definitions into a separate file has the
advantage that it can be used as-is without editing in other
deployments. For example, the kubernetes-csi/docs example can
use this rbac.yaml file instead of a local copy.

### DIFF
--- a/deploy/kubernetes/rbac.yaml
+++ b/deploy/kubernetes/rbac.yaml
@@ -1,0 +1,65 @@
+# Together with the RBAC file for external-provisioner, this YAML file
+# contains all RBAC objects that are necessary to run external CSI
+# snapshotter.
+#
+# In production, each CSI driver deployment has to be customized:
+# - to avoid conflicts, use non-default namespace and different names
+#   for non-namespaced entities like the ClusterRole
+# - optionally rename the non-namespaced ClusterRole if there
+#   are conflicts with other deployments
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: csi-snapshotter
+
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  # rename if there are conflicts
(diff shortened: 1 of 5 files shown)


deploy/kubernetes/rbac.yaml
```diff
@@ -0,0 +1,67 @@
...
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["list", "watch", "create", "update", "patch"]
+  - apiGroups: [""]
+    resources: ["endpoints"]
+    verbs: ["list", "watch", "create", "update", "delete", "get"]
```
xing-yang: Can you update this to only include create, update, get?

pohly: Sure, if that's enough. I just copied the existing rules.

Do I need to test or are you 100% sure that it'll work?

pohly: Actually, now that we discuss the content: is this permission for endpoints needed in global scope or is it enough in the app's namespace (i.e. in a Role)?

For leadership election in external-attacher and external-provisioner we originally had too broad permissions, using a Role + RoleBinding was enough there.

xing-yang: Yes, I tested it.  I meant to submit a PR to update that.  Since you are changing this, it's good to update it.

xing-yang: You are right, this is broader than what is needed.  I don't want to be dependent on rbac.yaml in external-provisioner repo either.  I'd like example yaml files in this repo to work together.  If you can separate the rbac yaml into two, one for external-provisioner and one for external-snapshotter, that will be great.

PR #58, opened 2018-11-14
Breaking change: Update snapshotter to use CSI spec 1.0

msau42:
Snapshot controller handling of uploading and error states are disabled by this change.

We need this so we can update external-provisioner which depends on this repo.


msau42:
/hold

Need to update csi-test when it's cut

msau42:
Ok I think I managed to not make it completely breaking.  I updated the controller logic to be able to handle the ready=true case.  But handling uploading and error states are disabled by this change.

PR #61, opened 2018-11-17
Call CreateSnapshot to check if snapshot is processed

xing-yang:
This PR calls CreateSnapshot to check if the snapshot post-cut
processing (uploading) is complete and update status accordingly.

Note: external-snapshotter is no longer broken after this fix.


xing-yang:
@jingxu97 @msau42 Please take a look.

pkg/controller/snapshot_controller.go
```diff
@@ -421,13 +421,65 @@ func (ctrl *csiSnapshotController) checkandBindSnapshotContent(snapshot *crdv1.V
 	return nil
 }
 
+func (ctrl *csiSnapshotController) prepareCreateSnapshotInput(snapshot *crdv1.VolumeSnapshot) (*crdv1.VolumeSnapshotClass, *v1.PersistentVolume, string, map[string]string, error) {
```
wackxu: Since this function is extract from the `createSnapshotOperation`  function, so we can call this function in the `createSnapshotOperation` to reduce the repeat code

pkg/controller/snapshot_controller.go
```diff
@@ -421,13 +421,65 @@ func (ctrl *csiSnapshotController) checkandBindSnapshotContent(snapshot *crdv1.V
...
 	if err != nil {
-		return nil, fmt.Errorf("failed to check snapshot status %s with error %v", snapshot.Name, err)
+		return nil, fmt.Errorf("failed to prepare input parameters to call create snapshot of the volume, %s: %q", volume.Name, err)
 	}
-	timestamp := time.Now().UnixNano()
-	newSnapshot, err := ctrl.updateSnapshotStatus(snapshot, status, timestamp, size, IsSnapshotBound(snapshot, content))
+	driverName, snapshotID, timestamp, size, readyToUse, err := ctrl.handler.CreateSnapshot(snapshot, volume, class.Parameters, snapshotterCredentials)
+	if err != nil {
+		// Don't return here. Update status by calling updateSnapshotStatus below
+		glog.Errorf("checkandUpdateBoundSnapshotStatusOperation: failed to create snapshot %q", err)
```
wackxu: here we call `CreateSnapshot` is to query whether the snapshot is ready to use, can we update the message here to `
checkandUpdateBoundSnapshotStatusOperation: failed to call create snapshot for check whether the snapshot is ready to use %q`

xing-yang:
@wackxu Addressed your comments.  Please take a look.  Thanks.

pkg/controller/snapshot_controller.go
```diff
@@ -421,13 +421,65 @@ func (ctrl *csiSnapshotController) checkandBindSnapshotContent(snapshot *crdv1.V
...
+	}
+
+	volume, err := ctrl.getVolumeFromVolumeSnapshot(snapshot)
+	if err != nil {
+		glog.Errorf("prepareCreateSnapshotInput failed to get PersistentVolume object [%s]: Error: [%#v]", snapshot.Name, err)
+		return nil, nil, "", nil, err
+	}
+
+	// Create VolumeSnapshotContent name
+	contentName := GetSnapshotContentNameForSnapshot(snapshot)
```
jingxu97: maybe put this function outside of this function since it is not used when called by checkandUpdateBoundSnapshotStatusOperation?

maybe call prepareCreateSnapshotInput to getCreateSnapshotInfo?

pkg/controller/snapshot_controller.go
```diff
@@ -682,42 +705,22 @@ func (ctrl *csiSnapshotController) updateSnapshotStatus(snapshot *crdv1.VolumeSn
...
-
-	/* TODO FIXME
-	switch csistatus.Type {
-	case csi.SnapshotStatus_READY:
-		if bound {
-			status.Ready = true
-			// Remove the error if checking snapshot is already bound and ready
-			status.Error = nil
-			change = true
+	if inErr != nil {
```
jingxu97: I don't think we need to pass inErr into the function. There are two places this function updateSnapshotStatus is called. If there is error during createSnapshotCall, we have a separate logic to update error status.

pkg/controller/snapshot_controller.go
```diff
@@ -421,13 +421,65 @@ func (ctrl *csiSnapshotController) checkandBindSnapshotContent(snapshot *crdv1.V
...
+	var readyToUse bool = false
+	class, volume, _, snapshotterCredentials, err := ctrl.prepareCreateSnapshotInput(snapshot)
+	if err != nil {
+		return nil, fmt.Errorf("failed to prepare input parameters to create snapshot %s: %q", snapshot.Name, err)
+	}
+	driverName, snapshotID, timestamp, size, readyToUse, err := ctrl.handler.CreateSnapshot(snapshot, volume, class.Parameters, snapshotterCredentials)
 	if err != nil {
-		return nil, fmt.Errorf("failed to check snapshot status %s with error %v", snapshot.Name, err)
+		// Don't return here. Update status by calling updateSnapshotStatus below
+		glog.Errorf("checkandUpdateBoundSnapshotStatusOperation: failed to call create snapshot to check whether the snapshot is ready to use %q", err)
```
jingxu97: I still think we should return error after calling CreateSnapshot, then the called of this function checkandUpdateBoundSnapshotStatus will update the error status with event.

pkg/controller/snapshot_controller.go
xing-yang: How about getCreateSnapshotInput?  "Info" would be confusing here because "Info" includes other information belonging to a snapshot.

xing-yang:
@jingxu97 Addressed your comments.  Please take a look.  Thanks.

pkg/controller/snapshot_controller.go
jingxu97: GetSnapshotContentNameForSnapshot is actually construct content name if using dynamic provisioning. Here if we use this function, statically binding case won't work? If snapshot is already bound, we can get reference name directly from snapshot object, right?

I am not familiar with the secrete ref function.  Why it needs contentName?

xing-yang: contentName is used to construct secret namespace parameters. The secret is also needed at delete snapshot time when the snapshot object may not exist any more so we can't rely on the snapshot object.  The implementation to support secrets here is consistent with how it is done for persistent volumes.

I don't know if static binding can support secrets because snapshot class (and storage class) is not required for static binding so there may not be a way to pass in the secrets.  I'll need to do more investigation on that.

jingxu97: How about this, change the logic to
if snapshot object already has a snapshotContentRef, use it, otherwise, construct the content name. Could you change the name from GetSnapshotContentNameForSnapshot to makeSnapshotContentName....?

PR #67, opened 2018-11-20
fix fail UT

wackxu:
@xing-yang @jingxu97


xing-yang:
/lgtm
/approve
Thanks for fixing this!

PR #68, opened 2018-11-22
fix TODO use time.Time for createSnapshot

wackxu:
fix TODO: clean this up, we probably don't need this translation layer and can just use time.Time


pkg/controller/snapshot_create_test.go
```diff
@@ -27,10 +27,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var timeNow = time.Now().UnixNano()
+var timeNow = time.Now()
+var timeNowStamp = timeNow.UnixNano()
```
xing-yang: Is it necessary to convert this to UnixNano?  Can we use timeNow directly in the tests?

wackxu: No, in volumesnapshot, the type of  `CreationTime` in VolumeSnapshotStatus is `*metav1.Time`, while in `CSIVolumeSnapshotSource` , the type for`CreationTime` is `*int64`, so keep two types of time is convenient for us to construct `snapshot` and `content`.

xing-yang:
This is picked up by https://github.com/kubernetes-csi/external-snapshotter/pull/119.  Will close this one.  Thanks @wackxu.

PR #71, opened 2018-11-27
Rename Ready to ReadyToUse

xing-yang:
This PR renames Ready to ReadyToUse in VolumeSnapshot Status
to be consistent with the name in CSI spec.

This should be added to the release note.


jingxu97:
/lgtm
Where we should put the release note?

Issue #70, opened 2018-11-27
Add pvcLister to snapshot controller


xing-yang:
We should add pvcLister to the snapshot controller, and after that changes should be made to snapshot controller to use pvcLister instead of using the client.  This is more efficient.

This was brought up during code reviews of https://github.com/kubernetes-csi/external-snapshotter/pull/39

PR #72, opened 2018-11-27
add pvcLister to snapshot controller

wackxu:
fix https://github.com/kubernetes-csi/external-snapshotter/issues/70
@xing-yang @jingxu97


xing-yang:
/lgtm

Thanks for the fix!

xing-yang:
@wackxu This PR was merged earlier but caused a hang when creating a snapshot.  So I had to revert it.  Please fix the problem and re-submit.  Thanks.

xing-yang:
@wackxu When fixing this issue, please also add a unit test to catch this hang.  Thanks.

wackxu:
@xing-yang Sorry for that, will take a look.

PR #73, opened 2018-11-28
Revert "add pvcLister to snapshot controller"

xing-yang:
Reverts kubernetes-csi/external-snapshotter#72

PR #72 caused a hang when creating a snapshot.  As a result, the snapshot was never created.  Need to revert it.


xing-yang:
PR #72 caused a hang in snapshot creation.  Need to revert it.

PR #75, opened 2018-11-30
add pvcLister to snapshot controller

### LINKED ISSUE #70: Add pvcLister to snapshot controller

wackxu:
Fixes #70


xing-yang:
Tested this PR and it works now.

PR #76, opened 2018-11-30
Use logging util

### LINKED ISSUE #62: Take advantage of `csi_secret` in CSI 1.0
CSI 1.0 decorates senstive fields with `csi_secret`. Let's take advantage of this feature to programmatically ensure no sensitive fields are ever logged by this side car container.

msau42:
Fixes #62


pkg/connection/connection.go
```diff
@@ -267,8 +268,9 @@ func (c *csiConnection) Close() error {
 
 func logGRPC(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 	glog.V(5).Infof("GRPC call: %s", method)
+	glog.V(5).Infof("GRPC request: %s", protosanitizer.StripSecrets(req))
```
msau42: @xing-yang I noticed snapshotter wasn't logging the request, so I added it back. I also quickly scanned through rest of snapshotter code to see if it was logging requests elsewhere and didn't find anything.  Could you double check?

xing-yang: Yes, we removed the logging of the request when Jan reviewed it due to the secrets issue.  This is the only spot.

msau42:
xref: https://github.com/kubernetes-csi/external-provisioner/issues/82

PR #78, opened 2018-12-01
Add new reserved prefixed parameter keys

xing-yang:
This PR adds new reserved prefixed parameter keys which are stripped
out of parameter list, and adds deprecation notice for old keys and
keep their behavior the same.

```
csiParameterPrefix = "csi.storage.k8s.io/"
prefixedSnapshotterSecretNameKey      = csiParameterPrefix + "snapshotter-secret-name"
prefixedSnapshotterSecretNamespaceKey = csiParameterPrefix + "snapshotter-secret-namespace"
```

This fix is modeled after a similar fix in external-provisioner: https://github.com/kubernetes-csi/external-provisioner/pull/178 

/assign @davidz627 @jingxu97 @saad-ali @msau42


jingxu97:
/lgtm
Thanks!

PR #82, opened 2019-01-06
Update exported function details with source code comments.

humblec:

### DIFF
--- a/pkg/connection/connection.go
+++ b/pkg/connection/connection.go
@@ -72,6 +72,7 @@ var (
 	_ CSIConnection = &csiConnection{}
 )
 
+// New returns a CSI connection object.
 func New(address string, timeout time.Duration) (CSIConnection, error) {
 	conn, err := connect(address, timeout)
 	if err != nil {
--- a/pkg/controller/csi_handler.go
+++ b/pkg/controller/csi_handler.go
@@ -42,6 +42,7 @@ type csiHandler struct {
 	snapshotNameUUIDLength int
 }
 
+// NewCSIHandler returns a handler which includes the csi connection and Snapshot name details
 func NewCSIHandler(
 	csiConnection connection.CSIConnection,
 	timeout time.Duration,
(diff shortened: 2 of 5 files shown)


humblec:
@xing-yang any idea why the test fails ? somehow travis log is not verbose and locally running the failed line gives me below.
 
```
[root@localhost external-snapshotter]# go fmt $(go list ./... | grep -v vendor) | wc -l 
0
[root@localhost external-snapshotter]# 
```

pkg/connection/connection.go
```diff
@@ -72,6 +72,7 @@ var (
 	_ CSIConnection = &csiConnection{}
 )
 
+//New returns a CSI connection object.
```
xing-yang: Add a space after "//"

pkg/controller/csi_handler.go
```diff
@@ -42,6 +42,7 @@ type csiHandler struct {
 	snapshotNameUUIDLength int
 }
 
+// NewCSIHandler returns a handler which include the csi connection and Snapshot name details
```
xing-yang: s/include/includes

pkg/controller/util.go
```diff
@@ -68,7 +68,7 @@ const (
 )
 
 var snapshotterSecretParams = deprecatedSecretParamsMap{
-	name: "Snapshotter",
+	name:                         "Snapshotter",
```
xing-yang: I think this change will cause "go fmt" to fail in 1.10.3.

pkg/connection/connection.go
humblec: Done.

pkg/controller/csi_handler.go
humblec: Done.

pkg/controller/util.go
humblec: ah.. It looks like thats causing it. I have removed this change for now.

PR #90, opened 2019-01-28
Fix #89: Increase probe timeout

### LINKED ISSUE #89: Probe timeout is too low
Analogous to the [external-attacher issue](https://github.com/kubernetes-csi/external-attacher/issues/67), we have a timeout for the probe request that is set to 1 second, which is too low for many plugins that want to do a thorough health check.

We should increase this timeout to 1 minute like we did on the external-attacher.

Akrog:
The Probe timeout is set to 1 second, which is insufficient for most
plugins that want to do a thorough health check.

Change the timeout to the default of 1 minute, like we did in the
external-attacher.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -754,7 +754,6 @@
     "github.com/kubernetes-csi/csi-test/driver",
     "google.golang.org/grpc",
     "google.golang.org/grpc/codes",
-    "google.golang.org/grpc/connectivity",
     "google.golang.org/grpc/status",
     "k8s.io/api/core/v1",
     "k8s.io/api/storage/v1",
--- a/cmd/csi-snapshotter/main.go
+++ b/cmd/csi-snapshotter/main.go
@@ -126,6 +126,12 @@ func main() {
 		os.Exit(1)
 	}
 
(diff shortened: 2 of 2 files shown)


Akrog:
PR #92 messed up the timeout configuration and broke this PR.
Will wait until we merge PR #93, or an equivalent PR, to update this PR to work again.

xing-yang:
Hi @Akrog, https://github.com/kubernetes-csi/external-snapshotter/pull/93 is merged. Do you still need this PR?  Thanks.

Akrog:
@xing-yang I have checked the code and the new sidecar and everything is fine now.  Although I noticed that the [second commit](https://github.com/kubernetes-csi/external-snapshotter/pull/93/commits/35d8184c7b2af154300c24678fd9edd9c559ce94#diff-13b763d9a35ba25a826b73c5bad03da7R69) did not update the help of the command line flag and it still says 10 seconds.

xing-yang:
@Akrog Thanks for verifying the fix!  I'm fixing the help msg here: https://github.com/kubernetes-csi/external-snapshotter/pull/118

PR #92, opened 2019-02-14
external-snapshotter should use CSI connection lib

andrewsykim:
/assign @msau42 @jsafrane @vladimirvivien 

Updating CSI sidecars to use the common connection library in csi-lib-utils.


msau42:
We may need a followup PR similar to https://github.com/kubernetes-csi/external-provisioner/pull/230 in order to use a different connection timeout for the operations

andrewsykim:
Noted, I also forgot to remove usage of connection timeout in a few places, I'll follow up :)

PR #93, opened 2019-02-14
add csi timeout flag

andrewsykim:
Following up on https://github.com/kubernetes-csi/external-snapshotter/pull/92#issuecomment-463705342 to add operation timeout flag. 

/assign @msau42 @xing-yang

### DIFF
--- a/cmd/csi-snapshotter/main.go
+++ b/cmd/csi-snapshotter/main.go
@@ -51,7 +51,7 @@ const (
 	threads = 10
 
 	// Default timeout of short CSI calls like GetPluginInfo
-	csiTimeout = time.Second
+	defaultCSITimeout = 10 * time.Second
 )
 
 // Command line flags
@@ -66,6 +66,7 @@ var (
 	snapshotNamePrefix              = flag.String("snapshot-name-prefix", "snapshot", "Prefix to apply to the name of a created snapshot")
 	snapshotNameUUIDLength          = flag.Int("snapshot-name-uuid-length", -1, "Length in characters for the generated uuid of a created snapshot. Defaults behavior is to NOT truncate.")
 	showVersion                     = flag.Bool("version", false, "Show version.")
+	csiTimeout                      = flag.Duration("timeout", defaultCSITimeout, "The timeout for any RPCs to the CSI driver. Default is 10s.")
 
 	leaderElection          = flag.Bool("leader-election", false, "Enables leader election.")
 	leaderElectionNamespace = flag.String("leader-election-namespace", "", "The namespace where the leader election resource exists. Defaults to the pod namespace if not set.")
@@ -142,7 +143,7 @@ func main() {
 	}
 
 	// Pass a context with a timeout
-	ctx, cancel := context.WithTimeout(context.Background(), csiTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), *csiTimeout)
 	defer cancel()
 
 	// Find driver name
@@ -155,10 +156,9 @@ func main() {
 	klog.V(2).Infof("CSI driver name: %q", *snapshotterName)
 
 	// Check it's ready
-	if err = csirpc.ProbeForever(csiConn, csiTimeout); err != nil {
+	if err = csirpc.ProbeForever(csiConn, *csiTimeout); err != nil {
 		klog.Errorf("error waiting for CSI driver to be ready: %v", err)
 		os.Exit(1)
-
 	}
 
 	// Find out if the driver supports create/delete snapshot.
@@ -191,7 +191,7 @@ func main() {
 		*createSnapshotContentRetryCount,
 		*createSnapshotContentInterval,
 		snapShotter,
-		*connectionTimeout,
+		*csiTimeout,
 		*resyncPeriod,
 		*snapshotNamePrefix,
 		*snapshotNameUUIDLength,
--- a/cmd/csi-snapshotter/main.go
+++ b/cmd/csi-snapshotter/main.go
@@ -51,7 +51,7 @@ const (
 	threads = 10
 
 	// Default timeout of short CSI calls like GetPluginInfo
-	defaultCSITimeout = 10 * time.Second
+	defaultCSITimeout = time.Minute
 )
 
 // Command line flags


cmd/csi-snapshotter/main.go
```diff
@@ -141,7 +142,7 @@ func main() {
 	glog.V(2).Infof("CSI driver name: %q", *snapshotter)
 
 	// Check it's ready
-	if err = waitForDriverReady(csiConn, *connectionTimeout); err != nil {
+	if err = waitForDriverReady(csiConn, *operationTimeout); err != nil {
```
andrewsykim: Passing in operation timeout here seems awkward. Should waitForDriverReady just poll forever instead?

msau42: This is related: https://github.com/kubernetes-csi/csi-lib-utils/pull/17
cc @jsafrane

hakanmemisoglu:
@andrewsykim Any plans for this PR. It looks like (at least for my case) the snapshotter is unusable without this PR.

andrewsykim:
@hakanmemisoglu can you expand on what you need here? external-snapshotter on master has already been updated to use `ProbeForever` (see [here](https://github.com/kubernetes-csi/external-snapshotter/blob/master/cmd/csi-snapshotter/main.go#L157-L161)) so this PR may not be needed anymore. Did you want external-snapshotter to exit after some timeout?

hakanmemisoglu:
@andrewsykim  
I see the snapshotter timeouts whenever it makes a CSI call (CreateSnapshot for example).
I will dig it more but my first guess is `connectionTimeout` argument is eventually passing to the `handler` in controller, and context cancels in short time. I think that the default timeout for CSI calls needs to be longer. 
Another issue is that `1s` `csiTimeout` might not be ideal for certain environments.

msau42:
I think the request here is to have a different timeout value for CSI CreateSnapshot calls than what's used for the initial connection to the CSI driver.

andrewsykim:
PR updated

cmd/csi-snapshotter/main.go
```diff
@@ -51,7 +51,7 @@ const (
 	threads = 10
 
 	// Default timeout of short CSI calls like GetPluginInfo
-	csiTimeout = time.Second
+	defaultCSITimeout = time.Second
```
msau42: We should probably optimize the default for the primary operations like createSnapshot instead of the metadata operations.

andrewsykim: Makes sense, any suggestions for what this should be? 5s?

msau42: @xing-yang do you have any ideas on what would be a good number?

external-provisioner uses 10 seconds, external-attacher uses 15 seconds.

xing-yang: This depends on the storage backend.  We can set the default to 10 seconds, same as external-provisioner.

andrewsykim: set to 10s

cmd/csi-snapshotter/main.go
```diff
@@ -155,10 +156,9 @@ func main() {
 	klog.V(2).Infof("CSI driver name: %q", *snapshotterName)
 
 	// Check it's ready
-	if err = csirpc.ProbeForever(csiConn, csiTimeout); err != nil {
+	if err = csirpc.ProbeForever(csiConn, *csiTimeout); err != nil {
```
xing-yang: There is still just one timeout in the main. 

If you take a look of https://github.com/kubernetes-csi/external-attacher/blob/master/cmd/csi-attacher/main.go#L124 and https://github.com/kubernetes-csi/external-attacher/blob/master/cmd/csi-attacher/main.go#L131, there are two different timeouts.  In ProbeForever, it uses "timeout".  In context.WithTimeout, it uses csiTimeout.

Also the sequence is different.  In attacher, ProbeForever is called before context.WithTimeout and GetDriverName.  Should we do the same here?

andrewsykim: Maybe I'm missing something here. I think that in the other sidecars, those two values should be the same as well. For external-attacher, it seems like `csiTimeout` is only used for `GetDriverName` which seems odd to me. I can update the other sidecars to always use `--timeout` for ProbeForever &  GetDriverName. I think I also made the code harder to read because I assigned `csiTimeout` as the flag var where as in other places it's just assigned to `timeout`.

msau42: Having one timeout for all operations seems reasonable to me

andrewsykim: sgtm! This PR as-is reflects that. I will follow-up in the other sidecars to do the same

cmd/csi-snapshotter/main.go
hakanmemisoglu: @andrewsykim 
I saw couple of cases (GKE regional, Netapp etc) took more than 10s for cutting snapshot (between 30-45 s). 
If we set up default timeout to a higher value, a default experience for users will be better since they have lower chance to see false errors. Of course, it will have some drawbacks such as waiting unnecessarily for the timeouts. However, I still think that positive parts are more important than the drawbacks I mentioned.

xing-yang: I do think 30s is more realistic.  I suggested 10s because the default for external-provisioner is 10s, however, creating a volume usually takes much longer time than cutting a snapshot.

@hakanmemisoglu Have you tested this fix?  Does it address your problem now?

hakanmemisoglu: I had a basic fix where I was setting it to 1min (which is excess). I will update the comment when I try this one.

xing-yang: The downside of waiting for a snapshot to be cut for unnecessarily long time is that it may not be desirable for an application if it is frozen before the snapshotting.

msau42: How do the snapshot timeouts work? Do the users specify their desired timeout value?  Snapshots timeout handling is trickier than the other sidecars because the other sidecars will retry forever, whereas snapshots are only tried once.

hakanmemisoglu: @xing-yang @andrewsykim 
It works perfectly.

Tested with:
- Helm repo: stable/mysql
- PVC size: 8GiB (with almost no data)
- GKE 1.13.4-gke.10 (Public preview)
- Non-regional snapshot

xing-yang: Thanks @hakanmemisoglu !

andrewsykim: So what's the consensus for default time out? 30s?

xing-yang: @msau42 User will be specifying the desired timeout value using this timeout parameter.  In the ExecutionHook we are still designing, there will be other timeout value to control how long we should wait between the start of a freeze and the start of a thaw.

You made a good point that other sidecars will retry while snapshot does not.  Given that, it is better to set default to a longer time.  I looked at the old code, we set this to 1 minute before the recent change.  I’d prefer setting default to the original value of 1 minute.  Otherwise driver users will be surprised when snapshotting starts to fail with a much shorter default timeout value.
https://github.com/kubernetes-csi/external-snapshotter/blob/release-1.0/cmd/csi-snapshotter/main.go#L54

andrewsykim: sgtm, updating to 1m

xing-yang:
/lgtm
@msau42 do you have other comments?

hakanmemisoglu:
@andrewsykim @xing-yang @msau42  Thanks for the PR!

andrewsykim:
@hakanmemisoglu my pleasure! thanks for raising your concerns here!

PR #96, opened 2019-02-18
Fix #95: Call GetSnapshotStatus to check snapshot status

### LINKED ISSUE #95: GetSnapshotStatus is never called
In snapshot_controller.go, the function"CreateSnapshot" is called over and over again during  my development and test`.  Thus, a lot of snapshots were created in my test environment.

Finally, I found the function "GetSnapshotStatus" is never called is external-snapshotter. So "ListSnapshots"  function is never called.

I wonder if this is a bug?  If not , I do not think it is a good idea to call "CreateSnapshot"  in the function "checkandUpdateBoundSnapshotStatusOperation".

cwdsuzhou:
Fix #95.
In  csi-snapshotter, function "GetSnapshotStatus" is never called and "CreateSnapshot" is called in function "GetSnapshotStatus".

In this PR, I have fixed #95 and changed the some unit tests.

### DIFF
--- a/pkg/controller/snapshot_controller.go
+++ b/pkg/controller/snapshot_controller.go
@@ -72,7 +72,7 @@ import (
 // this snapshot. After that, the controller will keep checking the snapshot status
 // though csi snapshot calls. When the snapshot is ready to use, the controller set
 // the status "Bound" to true to indicate the snapshot is bound and ready to use.
-// If the createtion failed for any reason, the Error status is set accordingly.
+// If the creation failed for any reason, the Error status is set accordingly.
 // In alpha version, the controller not retry to create the snapshot after it failed.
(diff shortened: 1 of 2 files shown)


xing-yang:
We replaced the call to ListSnapshots with CreateSnapshot to check snapshot status when adding support to CSI spec 1.0:
https://github.com/kubernetes-csi/external-snapshotter/pull/61

As stated in the CSI spec, ListSnapshots is optional.  CreateSnapshot is the preferred way to check snapshot status.
https://github.com/container-storage-interface/spec/blob/master/spec.md#createsnapshot-deletesnapshot-listsnapshots

The CreateSnapshot call in a CSI driver should be idempotent.  So repeated call to CreateSnapshot with same parameters should return the same result and just create one snapshot, not multiples.

Issue #60, opened 2018-11-17
Break out snapshot CRD libraries into separate repo


msau42:
This makes it easier to update external-provisioner dependencies

xing-yang:
This is a good idea.  I'll work on this after the v1.13 release.

msau42:
cc @nikhita is this something that the k8s publishing bot can help with?

nikhita:
> cc @nikhita is this something that the k8s publishing bot can help with?

Yes :)

From our conversation on Slack: iiuc this repo is supposed to be canonical source of truth and the  `client` (and maybe the `api`) directory needs to be published/synced to another repo.

For this, we'll just need a set of rules like https://github.com/kubernetes/kubernetes/blob/master/staging/publishing/rules.yaml. Very roughly, it'll look something like:

```
rules:
- destination: published-client-repo-name
  library: true
  branches:
  - source:
      branch: master
      dir: pkg/client
    name: master
    dependencies:
    - repository: api
      branch: master
    - repository: apimachinery
      branch: master
    - repository: client-go
      branch: master
```

We'll also need a cluster to run this bot. Maybe wg-k8s-infra can help with this? cc @dims 

/cc @sttts

nikhita:
To describe more concrete next steps: If we can get approval from sig-architecture (i'm guessing this falls under [code-organization](https://github.com/kubernetes/community/tree/master/sig-architecture#subprojects)) and sig-storage, I can help move forward with this.

msau42:
We discussed this in the sig-arch meeting today.  The suggestion is to first look into switching over to go modules.  If that doesn't work, then we can come back and discuss the staging/publishing bot option.

PR #100, opened 2019-03-09
Remove constraints and update all vendor pkgs

msau42:
I am removing k8s constraints so that it's hopefully easier to update external-provisioner k8s dependencies.  I think this is safe to do now since the APIs that snapshotter needs are in official releases. If there is a need in the future to set a constraint on an specific version, then we can add it at that point.  I also want to make more progress on #60 so that the external-provisioner doesn't have to import external-snapshotter, only the apis.

Edit: since snapshotter still depends on alpha APIs, it's easier to handle breaking alpha changes if all the k8s packages are on the same release.  So I am fixing all the k8s dependencies to release-1.14 so we can unblock external-provisioner


msau42:
/assign @xing-yang @jingxu97 
can one of you help test this?

xing-yang:
I don't think it is right to get rid of K8S dependencies in this project.  It does depend on in-tree APIs.

msau42:
K8s APIs should be stable.  The exception is the alpha PVC data source field.  However, I think if that API does ever break, we will know about it because compilation will fail, and at that time then we can add the constraint if we don't want to fix it.

msau42:
Also it should be noted that by not specifying any constraints, dep will try to find the latest tagged version to use.  If there are no tags that follow semantic versioning, only then will it use master head.

msau42:
I brought up an alpha cluster and ran the Snapshot e2es with my private image and they passed:
```
Ran 2 of 3260 Specs in 130.989 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 3258 Skipped PASS

Ginkgo ran 1 suite in 2m13.104037756s
Test Suite Passed
2019/03/10 13:44:08 process.go:155: Step './hack/ginkgo-e2e.sh --ginkgo.focus=VolumeSnapshotDataSource' finished in 2m14.479080652s
```

xing-yang:
I still don't think we should remove k8s constraints from this repo.  If all repos under kubernetes-csi have k8s constraints removed, I would consider that is the common practice for kubernetes-csi repos (even though it is still unusual in general).  I don't think external-snapshotter repo should be handling k8s constraints so differently.

msau42:
I intend to go clean up all the sidecars, but am starting with snapshotter first because it's blocking external provisioner updates.

It took me a while to understand the basic concepts of dep as it different from other dependency management paradigms.  One major concept is that the updates to existing packages are controlled: the lock file fixes the version, until you explicitly run "dep ensure -update". That way you can thoroughly test out any vendor update before committing them, and also on your own timeline.

As I understand it, there are two main reasons to have a constraint:

* You need to pull in a build that has not been officially tagged yet
* You are dependent on some API or behavior in some older release and you cannot update your application to use the latest

As I understand it, the constraints on k8s were added originally because we needed to pull in k8s API changes that were not officially released yet. But now that they are, I think it is fine to remove the constraint and pull in the latest official releases. K8s apis have a well defined deprecation policy, and the apis we depend on are also ones that we maintain so we can coordinate any potential future changes. I don't think the current status of snapshotter needs any of the two reasons which is why I think it is safe to remove the constraints now.  We always have the option to add it back in the future if we need to.

Gopkg.toml
```diff
@@ -11,51 +11,7 @@ required = [
...
-  name = "github.com/kubernetes-csi/csi-test"
-  version = "1.0.0-rc2"
-
-[[constraint]]
-  name = "google.golang.org/grpc"
-  version = "1.9.2"
-
-[[constraint]]
-  name = "k8s.io/api"
-  version = "kubernetes-1.13.0-beta.1"
```
msau42: change branch = release-1.14

msau42:
Set all the constraints to release-1.14.  Tested private image with e2es and they passed.

Gopkg.toml
```diff
@@ -11,51 +11,34 @@ required = [
...
-[[constraint]]
-  name = "google.golang.org/grpc"
-  version = "1.9.2"
 
+# Branches are used for k8s.io constraints because at the
+# moment we need latest 1.14 APIs for external-provisioner.
+# This can be set to an official 1.14 tag once 1.14 is released.
+# The dependency on external-provisioner should be removed with #60.
+#
+# Once snapshotter depends on only stable K8s APIs, then all constraints should be removable.
```
xing-yang: I still don't think constraints should be removed.  Can you remove this line please?  Also we are still working on designing new features such as ExecutionHook and PVC taints that could impact in-tree APIs, so that could affect the dependencies on in-tree APIs too.

Gopkg.toml
```diff
@@ -11,51 +11,34 @@ required = [
...
+  branch = "release-1.14"
 
 [[constraint]]
   name = "k8s.io/client-go"
-  version = "kubernetes-1.13.0-beta.1"
-
-[[override]]
-  name = "github.com/json-iterator/go"
-  version = "1.1.4"
+  branch = "release-11.0"
```
xing-yang: Is this the latest stable branch that supports k8s 1.14?

Gopkg.toml
msau42: They impact the dependencies right now because it's an alpha feature and k8s doesn't offer any compatibility guarantees for alpha features. But once an API is GA, then it's very hard to change it.  I agree that while snapshotter is using alpha APIs, then we may have compatibility issues.

Gopkg.toml
msau42: yes, the 10.0 release supports k8s 1.13, so the 11.0 is for 1.14.

Gopkg.toml
msau42: Since snapshotter is still in alpha and is undergoing significant changes, I will remove this line.  But once the API and feature is stable, I would strongly encourage reconsidering this

PR #107, opened 2019-03-29
Add Leader Election Support 

andrewsykim:
Adds lease based leader election support for external snapshotter. Requires users to set flags `--leader-election` and `--leader-election-namespace`.

/assign @msau42 @xing-yang @jsafrane


Gopkg.toml
```diff
@@ -9,7 +9,7 @@ required = [
 ]
 
 
-[[constraint]]
+[[override]]
```
andrewsykim: This is to keep the CSI spec at v1.0.0 since we have v1.1.0 at csi-lib-utils

cmd/csi-snapshotter/main.go
```diff
@@ -60,10 +61,14 @@ var (
 	snapshotNamePrefix              = flag.String("snapshot-name-prefix", "snapshot", "Prefix to apply to the name of a created snapshot")
 	snapshotNameUUIDLength          = flag.Int("snapshot-name-uuid-length", -1, "Length in characters for the generated uuid of a created snapshot. Defaults behavior is to NOT truncate.")
 	showVersion                     = flag.Bool("version", false, "Show version.")
+
+	leaderElection          = flag.Bool("leader-election", false, "Enables leader election.")
+	leaderElectionNamespace = flag.String("leader-election-namespace", "", "The namespace where the leader election resource exists. It should be the same namespace where the attacher runs.")
```
andrewsykim: I would much prefer it if we tried to discover the namespace instead of having the user set it. Left this here to be consistent with external attacher

msau42: Can namespace be passed in as an ENV variable? then you can use downward API to get the namespace.  I guess you can also set the argument via downward API already.  I'm unsure if we want a mix of ENV and args.

msau42:
Also update the README, examples and rbac

cmd/csi-snapshotter/main.go
andrewsykim: That's fair. I'll leave this for now.

deploy/kubernetes/rbac-external-provisioner.yaml
```diff
@@ -59,7 +59,7 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
 
 ---
-# Provisioner must be able to work with endpoints in current namespace
+# Provisioner must be able to work with endpoints and leases in current namespace
```
msau42: @xing-yang why do we need external-provisioner examples here?  shouldn't we be referring to external-provisioner repo?

cmd/csi-snapshotter/main.go
jsafrane: external-provisioner discovers the namespace via serviceaccount token:

https://github.com/kubernetes-csi/external-provisioner/blob/fd73d0d7a28b894aacecf0c8749db53c9cd27456/vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/controller/controller.go#L1388-L1401

IMO, it's not so bad to follow. Adding it to csi-lib-util could be useful.

deploy/kubernetes/rbac-external-provisioner.yaml
xing-yang: This is just easier for folks who want to know how to get started with snapshot. They look at this one repo and can find an example of how to test snapshot, without going through multiple repos.

msau42: The difficulty is keeping it up to date as external-provisioner gets updated.

andrewsykim:
/hold 

Waiting for https://github.com/kubernetes-csi/csi-lib-utils/pull/24

deploy/kubernetes/rbac-external-provisioner.yaml
xing-yang: If we don't update the provisioner image, these rbac files will stay the same.  If we need to update the image, we'll need to update the rbac files. We have a note in README to remind folks to refresh the rbac files when the provisioner image is updated: https://github.com/kubernetes-csi/external-snapshotter/tree/master/deploy/kubernetes

deploy/kubernetes/setup-csi-snapshotter.yaml
```diff
@@ -89,9 +89,15 @@ spec:
           args:
             - "--csi-address=$(ADDRESS)"
             - "--connection-timeout=15s"
+            - "--leader-election=true"
+            - "--leader-election-namespace=$(POD_NAMESPACE)"
```
msau42: Actually, do we need to set the namespace?

andrewsykim: nope, updating

deploy/kubernetes/rbac.yaml
```diff
@@ -47,6 +47,10 @@ rules:
   - apiGroups: ["apiextensions.k8s.io"]
     resources: ["customresourcedefinitions"]
     verbs: ["create", "list", "watch", "delete"]
+    # access to leases is required for leader election
```
msau42: Oh actually, we only need access to the Lease object in the namespace of the sidecar right? not all namespaces in the cluster?

andrewsykim: yes, just just the current namespace. I was lazy and didn't want to create a dedicated role just for this but let me do that now

Gopkg.toml
xing-yang: Do you still need this override now that version is 1.1.0?

cmd/csi-snapshotter/main.go
xing-yang: Why does the namespace for snapshotter have to be in the same namespace as the attacher? Snapshotter could be running without an attacher.

msau42: Do we actually need the namespace field for snapshotter?

Only provisioner and attacher have legacy leader election that we may need backwards compat for.

xing-yang: Looking at this merged PR for resizer, it says namespace "Defaults to the pod namespace if not set"?  I am confused about the namespace.
https://github.com/kubernetes-csi/external-resizer/pull/31/files#diff-6bd6f3383abbcca8c0512632cea597e5R47

deploy/kubernetes/rbac.yaml
andrewsykim: done

Gopkg.toml
xing-yang: This should be a constraint, not override because it is in sync with csi-lib-utils.

andrewsykim: fixed

cmd/csi-snapshotter/main.go
xing-yang: Probably "Defaults to the pod namespace if not set" is still the right comment here too.  The pod is where the leader election resource runs.

andrewsykim: fixed

deploy/kubernetes/setup-csi-snapshotter.yaml
```diff
@@ -89,6 +89,7 @@ spec:
           args:
             - "--csi-address=$(ADDRESS)"
             - "--connection-timeout=15s"
+            - "--leader-election=true"
```
xing-yang: By default leader-election should be false in the deployment yaml file.  User can change to true when needed.

andrewsykim: ah okay, I was trying to be consistent with examples in other sidecars, they have leader election enabled in all the examples

PR #111, opened 2019-04-08
Prow testing

pohly:
This imports csi-release-tools (unchanged copy) and adds a .prow.sh file in 944be88.


pohly:
/hold

https://github.com/kubernetes/test-infra/pull/12088 has to be merged first

release-tools/util.sh
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env bash
+
+# Copyright 2014 The Kubernetes Authors.
```
xing-yang: nit: should be 2019

release-tools/verify-shellcheck.sh
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Kubernetes Authors.
```
xing-yang: 2019

release-tools/util.sh
pohly: The original code actually is from 2014. I just copied it.

release-tools/verify-shellcheck.sh
pohly: Same here.

pohly:
@xing-yang I can't change these commits here. They are verbatim copies from https://github.com/kubernetes-csi/csi-release-tools/commits/master

If you want to review such changes before they are merged, then I can add you as reviewer of future changes in that repo.

If you see something that absolutely has to be changed now, then we'll have to treat that as bugs against csi-release-tools. But these nits are probably something that we can leave unchanged. Agreed?

xing-yang:
@pohly If these are copies from https://github.com/kubernetes-csi/csi-release-tools/commits/master and cannot be changed in this repo, can you add some description at the top of this PR to clarify that?

pohly:
> If these are copies from https://github.com/kubernetes-csi/csi-release-tools/commits/master and cannot be changed in this repo, can you add some description at the top of this PR to clarify that?

I was assuming that component maintainers are familiar with how `csi-release-tools` (= `release-tools` in each component) is maintained. But sure, I can add something.

PR #122, opened 2019-05-08
Update deployment images to v1.1.0

xing-yang:
**What type of PR is this?**
/kind bug

This PR updates deployment images to v1.1.0.

Fixes #

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/setup-csi-snapshotter.yaml
+++ b/deploy/kubernetes/setup-csi-snapshotter.yaml
@@ -72,7 +72,7 @@ spec:
       serviceAccount: csi-snapshotter
       containers:
         - name: csi-provisioner
-          image: quay.io/k8scsi/csi-provisioner:v1.0.1
+          image: quay.io/k8scsi/csi-provisioner:v1.1.0
           args:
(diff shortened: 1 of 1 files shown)


xing-yang:
@msau42 @j-griffith PTAL.

deploy/kubernetes/setup-csi-snapshotter.yaml
```diff
@@ -72,7 +72,7 @@ spec:
       serviceAccount: csi-snapshotter
       containers:
         - name: csi-provisioner
-          image: quay.io/k8scsi/csi-provisioner:v1.0.1
+          image: quay.io/k8scsi/csi-provisioner:v1.1.0
```
msau42: should the examples here just use canary so we don't have to remember to keep updating it every release?

xing-yang: Canary images may not work with each other before they are released. For example, I may have modified external-snapshotter APIs in the master but have not got time to update external-provisioner code yet, so the canary images won't work with each other.

PR #123, opened 2019-05-09
Add groupName in doc.go when generating files

### LINKED ISSUE #112: Missing 'groupName' comment in doc.go - wrong group for fake objects
The group name of the APIs is "snapshot.storage.k8s.io" as specified here:

https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/apis/volumesnapshot/v1alpha1/register.go#L23

However, "fake" objects are generated with "volumesnapshot"

https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/client/clientset/versioned/typed/volumesnapshot/v1alpha1/fake/fake_volumesnapshot.go#L37

It seems that doc.go is missing the comment 
`
// +groupName=snapshot.storage.k8s.io
`

Due to this issue, we cannot use fake entities to do unit tests.

xing-yang:
**What type of PR is this?**
/kind bug

Adds groupName in doc.go when generating files so that the fake volumesnapshot has the correct groupName.

Fixes #112

```release-note
Changes the API group name for the fake VolumeSnapshot object to "snapshot.storage.k8s.io" to be in-sync with the group name of the real VolumeSnapshot object. As a result, the generated interfaces for clientset and informers of VolumeSnapshot are also changed from "VolumeSnapshot" to "Snapshot".
```


xing-yang:
@msau42 @j-griffith PTAL.  Thanks.

msau42:
Is this a breaking change? Do we need a release note?

xing-yang:
There is no API change here, so I initially thought release note is not needed. However, if someone uses informers and clientset of volumesnapshot directly, they will have to change their code.  So let me add a release note for that.

msau42:
Also the group name is changing right?

xing-yang:
The group name is only changed for the fake object which can be used for testing (to make sure they are in-sync with the real group name).  The group name for the real VolumeSnapshot object remains the same.  To my surprise, this also changes "VolumeSnapshot" to "Snapshot" when referencing the objects in the clientset and informers.

PR #129, opened 2019-06-19
Leader election lock name should include driver name

### LINKED ISSUE #128: Leader election lock name should include driver name
To be consistent with the other sidecars

zhucan:
**What type of PR is this?**
/kind bug

CSI has many controllers with leader election functionality. It's best to disambiguate between all the different locks used for LE

Fixes #128 

```release-note
Add a prefix "external-snapshotter-leader" and the driver name to the snapshotter leader election lock name.
```

### DIFF
--- a/cmd/csi-snapshotter/main.go
+++ b/cmd/csi-snapshotter/main.go
@@ -73,8 +73,8 @@ var (
 )
 
 var (
-	version                = "unknown"
-	leaderElectionLockName = "external-snapshotter-leader-election"
+	version = "unknown"
+	prefix  = "external-snapshotter-leader"
 )
 
 func main() {
@@ -214,7 +214,8 @@ func main() {
 	if !*leaderElection {
 		run(context.TODO())
 	} else {
-		le := leaderelection.NewLeaderElection(kubeClient, leaderElectionLockName, run)
+		lockName := fmt.Sprintf("%s-%s", prefix, strings.Replace(*snapshotterName, "/", "-", -1))
+		le := leaderelection.NewLeaderElection(kubeClient, lockName, run)
 		if *leaderElectionNamespace != "" {
 			le.WithNamespace(*leaderElectionNamespace)
 		}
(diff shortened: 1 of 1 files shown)


xing-yang:
Please remove > before /kind bug

zhucan:
> Please remove > before /kind bug

copy that

cmd/csi-snapshotter/main.go
```diff
@@ -214,7 +215,8 @@ func main() {
 	if !*leaderElection {
 		run(context.TODO())
 	} else {
-		le := leaderelection.NewLeaderElection(kubeClient, leaderElectionLockName, run)
+		lockName := fmt.Sprintf("%s-%s", prefix, strings.Replace(*snapshotterName, "/", "-", -1))
```
xing-yang: I think we don't need this prefix at all.  The lockName used in the provisioner does not have a prefix:
https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L202

msau42: We do need the prefix otherwise the name will conflict with other sidecar's locks.  provisioner unfortunately has more backwards compatibility concerns so it's not so easy to fix: https://github.com/kubernetes-csi/external-provisioner/issues/295

xing-yang: Sound good.

cmd/csi-snapshotter/main.go
```diff
@@ -73,8 +74,8 @@ var (
 )
 
 var (
-	version                = "unknown"
-	leaderElectionLockName = "external-snapshotter-leader-election"
+	version = "unknown"
+	prefix  = "external-snapshotter-leader"
```
xing-yang: @msau42 Is there any limitation on the length of the lockName?  This prefix is a little longer.  Other than that, I'm fine with it.

msau42: The attacher and resizer all use "external-" prefix. Probably we should just be the same to be consistent.

In terms of length, the name is used as the ObjectMeta.Name, which is 253 characters

xing-yang: Sure.

PR #130, opened 2019-06-19
minimize rbac permission by setting secrets as optional

hoyho:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

 /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Set secret rbac as optional.
Some drivers may require a secret while some don't.
By default, we should give less permission as possible because `ClusterRole` is quite a broad.
Other sidecars like external-provisioner and external-attacher have comment out that. It should be ok to do it in external-snapshotter.
ref https://kubernetes-csi.github.io/docs/secrets-and-credentials.html#secret-rbac-rules
related issue https://github.com/kubernetes-csi/external-provisioner/issues/164

Fixes #

```release-note
None
```

### DIFF
--- a/deploy/kubernetes/rbac.yaml
+++ b/deploy/kubernetes/rbac.yaml
@@ -32,9 +32,13 @@ rules:
   - apiGroups: [""]
     resources: ["events"]
     verbs: ["list", "watch", "create", "update", "patch"]
-  - apiGroups: [""]
-    resources: ["secrets"]
-    verbs: ["get", "list"]
+  # Secret permission is optional.
+  # Enable it if your driver needs secret.
+  # For example, `csi.storage.k8s.io/snapshotter-secret-name` is set in VolumeSnapshotClass.
+  # See https://kubernetes-csi.github.io/docs/secrets-and-credentials.html for more details.
+  #  - apiGroups: [""]
+  #    resources: ["secrets"]
+  #    verbs: ["get", "list"]
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshotclasses"]
     verbs: ["get", "list", "watch"]


deploy/kubernetes/rbac.yaml
```diff
@@ -51,6 +51,14 @@ rules:
     resources: ["customresourcedefinitions"]
     verbs: ["create", "list", "watch", "delete", "get", "update"]
 
+#Secret permission is optional.
```
xing-yang: Add a space after #

deploy/kubernetes/rbac.yaml
```diff
@@ -51,6 +51,14 @@ rules:
...
+#Enable it if you need value from secret.
```
xing-yang: s/if you need value from secret/if your driver needs secret

deploy/kubernetes/rbac.yaml
```diff
@@ -51,6 +51,14 @@ rules:
...
+#For example, you have key `csi.storage.k8s.io/controller-publish-secret-name` in StorageClass.parameters.
```
xing-yang: For example, `csi.storage.k8s.io/snapshotter-secret-name` is set in VolumeSnapshotClass.

deploy/kubernetes/rbac.yaml
hoyho: thanks for the advice.

deploy/kubernetes/rbac.yaml
hoyho: fixed

deploy/kubernetes/rbac.yaml
hoyho: fixed

deploy/kubernetes/rbac.yaml
```diff
@@ -51,6 +51,14 @@ rules:
     resources: ["customresourcedefinitions"]
     verbs: ["create", "list", "watch", "delete", "get", "update"]
 
+# Secret permission is optional.
+# Enable it if your driver needs secret.
+# For example, `csi.storage.k8s.io/snapshotter-secret-name` is set in VolumeSnapshotClass.
+# See https://kubernetes-csi.github.io/docs/secrets-and-credentials.html for more details.
```
xing-yang: Comments should be added above lines #35-37 and comment out lines #35-37.

hoyho: updated

deploy/kubernetes/rbac.yaml
```diff
@@ -32,9 +32,13 @@ rules:
   - apiGroups: [""]
     resources: ["events"]
     verbs: ["list", "watch", "create", "update", "patch"]
-  - apiGroups: [""]
-    resources: ["secrets"]
-    verbs: ["get", "list"]
+# Secret permission is optional.
```
xing-yang: I think you need to add space and make sure this aligns with - apiGroups: [""] on line 32.  yaml file is space sensitive.

hoyho: Actually, no need to worry about that. Comment lines in yaml file will be ignored. I have test it and it can apply/create. Maybe some CO have different behavior, so I also update the yaml file as you recommend

PR #138, opened 2019-07-10
Check if ListSnapshot is supported during GetSnapshotStatus

### LINKED ISSUE #132: Snapshotter uses ListSnapshots on drivers which do not advertise it.
Snaphotter uses a function called GetSnapshotStatus() to do the binding from a VolumeSnapshot to a VolumeSnapshotContent. GetSnapshotStatus() then uses the csi.ListSnapshot() call even when the driver does not support it. Therefore this causes an error and the binding does not occur.

ggriffiths:
**What type of PR is this?**
/kind bug

This PR introduces a check to see if ListSnapshot is supported before calling it GetSnapshotStatus.

If ListSnapshot is not supported, we just return that the snapshot is ReadyToUse and no creation time/restore time.

Fixes #132

n/a

```release-note
Snapshotter will no longer call ListSnapshots if the CSI Driver does not support this operation. 
```

### DIFF
--- a/pkg/snapshotter/snapshotter.go
+++ b/pkg/snapshotter/snapshotter.go
@@ -101,13 +101,36 @@ func (s *snapshot) DeleteSnapshot(ctx context.Context, snapshotID string, snapsh
 	return nil
 }
 
+func (s *snapshot) isListSnapshotsSupported(ctx context.Context) (bool, error) {
+	client := csi.NewControllerClient(s.conn)
+	capRsp, err := client.ControllerGetCapabilities(ctx, &csi.ControllerGetCapabilitiesRequest{})
+	if err != nil {
+		return false, err
+	}
+
+	for _, cap := range capRsp.Capabilities {
+		if cap.GetRpc().GetType() == csi.ControllerServiceCapability_RPC_LIST_SNAPSHOTS {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
 func (s *snapshot) GetSnapshotStatus(ctx context.Context, snapshotID string) (bool, time.Time, int64, error) {
 	client := csi.NewControllerClient(s.conn)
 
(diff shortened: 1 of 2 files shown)


pkg/snapshotter/snapshotter.go
```diff
@@ -101,13 +101,37 @@ func (s *snapshot) DeleteSnapshot(ctx context.Context, snapshotID string, snapsh
...
+
+	return false, nil
+}
+
 func (s *snapshot) GetSnapshotStatus(ctx context.Context, snapshotID string) (bool, time.Time, int64, error) {
 	client := csi.NewControllerClient(s.conn)
 
+	// If the driver does not support ListSnapshots, assume the snapshot ID is valid.
+	listSnapshotsSupported, err := s.isListSnapshotsSupported(ctx)
+	if err != nil {
```
ggriffiths: We could also just return `return true, time.Time{}, 0, nil` here, but that seems like it could cause other problems.

Also, if the GetCapabilities call fails, I figured the caller should know.

xing-yang:
Please check test failures.

xing-yang:
I think we need a release note for this change.

ggriffiths:
> Please check test failures.

Fixed test failures and added a new test for when ListSnapshots is not supported.

pkg/snapshotter/snapshotter.go
xing-yang: I'm fine with this.

xing-yang:
@hakanmemisoglu Can you help test this fix?  Thanks.

hakanmemisoglu:
Tested with Ceph (RBD) CSI driver (v1.0.0). 
It managed to bind previously created VolumeSnapshot and VolumeSnapshotContent.

xing-yang:
Thanks @hakanmemisoglu!

PR #143, opened 2019-07-12
Prow update

pohly:
**What type of PR is this?**
/kind failing-test

We want to support testing under Kubernetes 1.15 and need updated rules for that. This PR also fixes the problem that the current rules reference a 1.14 deployment of "canary" images, which became problematic when making incompatible changes in the hostpath driver (driver name change).

See https://github.com/kubernetes-csi/external-provisioner/pull/313

```release-note
NONE
```

### DIFF
--- a/release-tools/build.make
+++ b/release-tools/build.make
@@ -118,14 +118,39 @@ test-fmt:
 	fi
 
 # This test only runs when dep >= 0.5 is installed, which is the case for the CI setup.
+# When using 'go mod', we allow the test to be skipped in the Prow CI under some special
+# circumstances, because it depends on accessing all remote repos and thus
+# running it all the time would defeat the purpose of vendoring:
+# - not handling a PR or
+# - the fabricated merge commit leaves go.mod, go.sum and vendor dir unchanged
+# - release-tools also didn't change (changing rules or Go version might lead to
+#   a different result and thus must be tested)
 .PHONY: test-vendor
 test: test-vendor
 test-vendor:
 	@ echo; echo "### $@:"
-	@ case "$$(dep version 2>/dev/null | grep 'version *:')" in \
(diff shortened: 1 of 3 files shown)


msau42:
Trying to fix alpha testing with https://github.com/kubernetes/test-infra/pull/13434

pohly:
> Trying to fix alpha testing with kubernetes/test-infra#13434

So the problem is that the current csi-release-tools rules only support alpha clusters on Kubernetes >= 1.15?

We probably don't loose much test coverage by simply avoid that problem and limiting alpha testing to recent Kubernetes. However, a better solution would be to make the alpha cluster creation depend on the Kubernetes version. If you want, then I can look into that.

msau42:
Feel free to look into extending the prow job to handle different versions, but I think it will be painful to maintain if the configs are going to be different and incompatible across releases. Maybe can we store a template per version as files in the repo and then patch it in script?  I think maintaining all the different configs in script is going to be very difficult to read.

pohly:
Michelle Au <notifications@github.com> writes:
> Feel free to look into extending the prow job to handle different
> versions, but I think it will be painful to maintain if the configs
> are going to be different and incompatible across releases.

I briefly looked at it and I agree, this is more work than is justified
right now. Let's go with your approach of limiting alpha testing to 1.15.

pohly:
> Trying to fix alpha testing with kubernetes/test-infra#13434

So the problem is that the current csi-release-tools rules only support alpha clusters on Kubernetes >= 1.15?

We probably don't loose much test coverage by simply avoid that problem and limiting alpha testing to recent Kubernetes. However, a better solution would be to make the alpha cluster creation depend on the Kubernetes version. If you want, then I can look into that.
t

PR #153, opened 2019-08-03
Add exponential backoff retries to CreateSnapshot

### LINKED ISSUE #134: Problems dealing with snapshot create requests timing out
When a CSI plugin is passed a CreateSnapshot request and the caller (snapshotter sidecar) times out, the snapshotter sidecar marks this as an error and does not retry the snapshot. Further, as the call only timed out and did not fail, the storage provider may have actually created the said snapshot (although delayed).

When such an snapshot is deleted, there are no requests to the CSI plugin to delete the same, which cannot be issued by the sidecar as it does not have the `SnapID`.

The end result of this is that the snapshot is leaked on the storage provider.

The question/issue hence is as follows,

Should the snapshot be retried on timeouts from the CreateSnapshot call?

Based on the ready_to_use parameter in the CSI spec [1] and possibilities of application freeze as the snapshot is taken, I would assume this operation cannot be done indefinitely. But, also as per the spec timeout errors, the behavior should be a retry, as implemented for volume create and delete operations in the provisioner sidecar [2].

So to fix the potential snapshot leak by the storage provider, should the snapshotter sidecar retry till it gets an error from the plugin or a success with a `SnapID`, but mark the snapshot as bad/unusable as it was not completed in time (to honor the application freeze times and such)?

[1] CSI spec `ready_to_use` section: https://github.com/container-storage-interface/spec/blob/master/spec.md#the-ready_to_use-parameter

[2] timeout handling in provisioner sidecar: https://github.com/kubernetes-csi/external-provisioner#csi-error-and-timeout-handling

ggriffiths:

**What type of PR is this?**
/kind bug

* Adds exponential retries to the external snapshotter

Fixes #134 

This PR is modeled after @jsafrane's retry work on the external-provisioner:
- Changes to `pkg/snapshotter/` are similar to https://github.com/kubernetes-csi/external-provisioner/commit/8203a03c47ce2b86a2a2c2421d74345b76183b14
- Changes to `pkg/controller` are based on:
  - https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/commit/c19e26be3fd97c9f9d97a47d5198fdfa14a38aa9 
  - https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/commit/540eb48f9f04d2ebaa249ae79ad1b1307ab0e50d

Ready for review now. 

```release-note
external-snapshotter will now do exponential backoff retries after a failure instead of timing out.
```

### DIFF
--- a/cmd/csi-snapshotter/main.go
+++ b/cmd/csi-snapshotter/main.go
@@ -62,6 +62,8 @@ var (
 	csiAddress                      = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
 	createSnapshotContentRetryCount = flag.Int("create-snapshotcontent-retrycount", 5, "Number of retries when we create a snapshot content object for a snapshot.")
 	createSnapshotContentInterval   = flag.Duration("create-snapshotcontent-interval", 10*time.Second, "Interval between retries when we create a snapshot content object for a snapshot.")
+	retryIntervalStart              = flag.Duration("retry-interval-start", time.Second, "Initial retry interval of failed snapshot creation. It doubles with each failure, up to retry-interval-max.")
+	retryIntervalMax                = flag.Duration("retry-interval-max", 5*time.Minute, "Maximum retry interval of failed snapshot creation.")
 	resyncPeriod                    = flag.Duration("resync-period", 60*time.Second, "Resync interval of the controller.")
 	snapshotNamePrefix              = flag.String("snapshot-name-prefix", "snapshot", "Prefix to apply to the name of a created snapshot")
 	snapshotNameUUIDLength          = flag.Int("snapshot-name-uuid-length", -1, "Length in characters for the generated uuid of a created snapshot. Defaults behavior is to NOT truncate.")
@@ -176,6 +178,8 @@ func main() {
 		coreFactory.Core().V1().PersistentVolumeClaims(),
 		*createSnapshotContentRetryCount,
 		*createSnapshotContentInterval,
+		*retryIntervalStart,
+		*retryIntervalMax,
 		snapShotter,
 		*csiTimeout,
 		*resyncPeriod,
--- a/pkg/controller/csi_handler.go
+++ b/pkg/controller/csi_handler.go
@@ -30,7 +30,7 @@ import (
 
 // Handler is responsible for handling VolumeSnapshot events from informer.
 type Handler interface {
-	CreateSnapshot(snapshot *crdv1.VolumeSnapshot, volume *v1.PersistentVolume, parameters map[string]string, snapshotterCredentials map[string]string) (string, string, time.Time, int64, bool, error)
+	CreateSnapshot(snapshot *crdv1.VolumeSnapshot, volume *v1.PersistentVolume, parameters map[string]string, snapshotterCredentials map[string]string) (string, string, time.Time, int64, bool, snapshotter.SnapshottingState, error)
 	DeleteSnapshot(content *crdv1.VolumeSnapshotContent, snapshotterCredentials map[string]string) error
 	GetSnapshotStatus(content *crdv1.VolumeSnapshotContent) (bool, time.Time, int64, error)
 }
@@ -58,19 +58,20 @@ func NewCSIHandler(
 	}
 }
 
-func (handler *csiHandler) CreateSnapshot(snapshot *crdv1.VolumeSnapshot, volume *v1.PersistentVolume, parameters map[string]string, snapshotterCredentials map[string]string) (string, string, time.Time, int64, bool, error) {
+func (handler *csiHandler) CreateSnapshot(snapshot *crdv1.VolumeSnapshot, volume *v1.PersistentVolume, parameters map[string]string, snapshotterCredentials map[string]string) (string, string, time.Time, int64, bool, snapshotter.SnapshottingState, error) {
 
 	ctx, cancel := context.WithTimeout(context.Background(), handler.timeout)
 	defer cancel()
 
 	snapshotName, err := makeSnapshotName(handler.snapshotNamePrefix, string(snapshot.UID), handler.snapshotNameUUIDLength)
 	if err != nil {
-		return "", "", time.Time{}, 0, false, err
+		return "", "", time.Time{}, 0, false, snapshotter.SnapshottingFinished, err
 	}
 	newParameters, err := removePrefixedParameters(parameters)
 	if err != nil {
-		return "", "", time.Time{}, 0, false, fmt.Errorf("failed to remove CSI Parameters of prefixed keys: %v", err)
+		return "", "", time.Time{}, 0, false, snapshotter.SnapshottingFinished, fmt.Errorf("failed to remove CSI Parameters of prefixed keys: %v", err)
 	}
+
 	return handler.snapshotter.CreateSnapshot(ctx, snapshotName, volume, newParameters, snapshotterCredentials)
 }
 
--- a/pkg/controller/snapshot_controller.go
+++ b/pkg/controller/snapshot_controller.go
@@ -18,10 +18,8 @@ package controller
 
 import (
 	"fmt"
-	"strings"
-	"time"
-
 	crdv1 "github.com/kubernetes-csi/external-snapshotter/pkg/apis/volumesnapshot/v1beta1"
+	"github.com/kubernetes-csi/external-snapshotter/pkg/snapshotter"
 	v1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
@@ -34,6 +32,8 @@ import (
 	"k8s.io/kubernetes/pkg/util/goroutinemap"
 	"k8s.io/kubernetes/pkg/util/goroutinemap/exponentialbackoff"
 	"k8s.io/kubernetes/pkg/util/slice"
+	"strings"
+	"time"
 )
 
 // ==================================================================
@@ -363,20 +363,37 @@ func (ctrl *csiSnapshotController) storeContentUpdate(content interface{}) (bool
 
 // createSnapshot starts new asynchronous operation to create snapshot
 func (ctrl *csiSnapshotController) createSnapshot(snapshot *crdv1.VolumeSnapshot) error {
-	klog.V(5).Infof("createSnapshot[%s]: started", snapshotKey(snapshot))
-	opName := fmt.Sprintf("create-%s[%s]", snapshotKey(snapshot), string(snapshot.UID))
+	key := snapshotKey(snapshot)
+	klog.V(5).Infof("createSnapshot[%s]: started", key)
+	opName := fmt.Sprintf("create-%s[%s]", key, string(snapshot.UID))
 	ctrl.scheduleOperation(opName, func() error {
-		snapshotObj, err := ctrl.createSnapshotOperation(snapshot)
+		snapshotObj, state, err := ctrl.createSnapshotOperation(snapshot)
 		if err != nil {
 			ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotCreationFailed", fmt.Sprintf("Failed to create snapshot: %v", err))
 			klog.Errorf("createSnapshot [%s]: error occurred in createSnapshotOperation: %v", opName, err)
+
+			// Handle state:
+			if state == snapshotter.SnapshottingFinished {
+				// Snapshotting finished, remove obj from snapshotsInProgress.
+				ctrl.snapshotsInProgress.Delete(key)
+			} else if state == snapshotter.SnapshottingInBackground {
+				// Snapshotting still in progress.
+				klog.V(4).Infof("createSnapshot [%s]: Temporary error received, adding Snapshot back in queue: %v", key, err)
+				ctrl.snapshotsInProgress.Store(key, snapshotObj)
+			} else {
+				// State is SnapshottingNoChange. Don't change snapshotsInProgress.
+			}
+
 			return err
 		}
+
+		// If no errors, update the snapshot.
 		_, updateErr := ctrl.storeSnapshotUpdate(snapshotObj)
 		if updateErr != nil {
 			// We will get an "snapshot update" event soon, this is not a big error
-			klog.V(4).Infof("createSnapshot [%s]: cannot update internal cache: %v", snapshotKey(snapshotObj), updateErr)
+			klog.V(4).Infof("createSnapshot [%s]: cannot update internal cache: %v", key, updateErr)
 		}
+
 		return nil
 	})
 	return nil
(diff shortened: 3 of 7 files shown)


ggriffiths:
@jsafrane Can you take a look to see if I'm on the right track here? I added a few notes in the PR description for tasks I still need to do.

pkg/snapshotter/snapshotter.go
```diff
@@ -44,26 +49,61 @@ type Snapshotter interface {
...
+// 100% finished (with or without success)
+type SnapshottingState string
+
+const (
+	// SnapshottingInBackground tells the controller that snapshotting may be in progress
+	// after CreateSnapshotExt exits.
+	SnapshottingInBackground SnapshottingState = "Background"
+	// SnapshottingFinished tells the controller that snapshotting is not running in the background
+	// and has exited successfully or with errors.
+	SnapshottingFinished SnapshottingState = "Finishing"
```
jsafrane: Finishing -> Finished

pkg/snapshotter/snapshotter.go
```diff
@@ -44,26 +49,61 @@ type Snapshotter interface {
 }
 
 type snapshot struct {
-	conn *grpc.ClientConn
+	conn    *grpc.ClientConn
+	backoff *wait.Backoff
 }
 
+// SnapshottingState is the state of volume snapshotting. It tells the controller if snapshotting could
+// be in progress in the background after a CreateSnapshotExt() call returns or the snapshotting is
```
jsafrane: CreateSnapshotExt -> CreateSnapshot

pkg/snapshotter/snapshotter.go
```diff
@@ -44,26 +49,61 @@ type Snapshotter interface {
...
+func (s *snapshot) CreateSnapshot(ctx context.Context, snapshotName string, volume *v1.PersistentVolume, parameters map[string]string, snapshotterCredentials map[string]string) (string, string, time.Time, int64, bool, SnapshottingState, error) {
 	klog.V(5).Infof("CSI CreateSnapshot: %s", snapshotName)
 	if volume.Spec.CSI == nil {
-		return "", "", time.Time{}, 0, false, fmt.Errorf("CSIPersistentVolumeSource not defined in spec")
+		return "", "", time.Time{}, 0, false, SnapshottingFinished, fmt.Errorf("CSIPersistentVolumeSource not defined in spec")
 	}
 
 	client := csi.NewControllerClient(s.conn)
 
 	driverName, err := csirpc.GetDriverName(ctx, s.conn)
```
jsafrane: Consider caching the driver name (in a separate PR), it should not change during snapshotter lifetime.

pkg/controller/csi_handler.go
```diff
@@ -71,7 +71,13 @@ func (handler *csiHandler) CreateSnapshot(snapshot *crdv1.VolumeSnapshot, volume
 	if err != nil {
 		return "", "", time.Time{}, 0, false, fmt.Errorf("failed to remove CSI Parameters of prefixed keys: %v", err)
 	}
-	return handler.snapshotter.CreateSnapshot(ctx, snapshotName, volume, newParameters, snapshotterCredentials)
+
+	driverName, snapId, timestamp, size, readyToUse, snapshotterState, err := handler.snapshotter.CreateSnapshot(ctx, snapshotName, volume, newParameters, snapshotterCredentials)
+	if err != nil {
+		return
+	}
+	return driverName, snapId, timestamp, size, readyToUse, err
```
jsafrane: this does not look correct - what is SnapshotterState return value here?

pkg/controller/csi_handler.go
```diff
@@ -58,7 +58,7 @@ func NewCSIHandler(
 	}
 }
 
-func (handler *csiHandler) CreateSnapshot(snapshot *crdv1.VolumeSnapshot, volume *v1.PersistentVolume, parameters map[string]string, snapshotterCredentials map[string]string) (string, string, time.Time, int64, bool, error) {
+func (handler *csiHandler) CreateSnapshot(snapshot *crdv1.VolumeSnapshot, volume *v1.PersistentVolume, parameters map[string]string, snapshotterCredentials map[string]string) (string, string, time.Time, int64, bool, snapshotter.SnapshotterState, error) {
```
jsafrane: Where is this function called? How is the new return value processed? That should be the core of this PR :-)

pkg/snapshotter/snapshotter.go
```diff
@@ -73,17 +113,65 @@ func (s *snapshot) CreateSnapshot(ctx context.Context, snapshotName string, volu
 		Secrets:        snapshotterCredentials,
 	}
 
-	rsp, err := client.CreateSnapshot(ctx, &req)
-	if err != nil {
-		return "", "", time.Time{}, 0, false, err
+	var rsp *csi.CreateSnapshotResponse
+	var lastCreateError error
+	err = wait.ExponentialBackoff(*s.backoff, func() (bool, error) {
```
jsafrane: This will block the whole goroutine for a long time. We did the same approach in  the CSI external provisioner and it did not work well. For example:

* When user deletes Snapshot object, the controller most probably does not need to continue in this loop.
* When the last (fifth) `CreateSnapshot` returns that the snapshot is still in progress, it won't wait until the snapshot is done (or at least I cannot check what happens, processing of `csiHandler.CreateSnapshot` return values is not included in this PR.)

Instead, we have exponential backoff in the [main controller loop itself](https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/blob/a82cca075cb14e43242bfa0326db049deab949dd/controller/controller.go#L888) and it keeps "claims in progress" (i.e. those that were not confirmed to be finished by the driver) [separately](https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/blob/master/controller/controller.go#L970), continuing to call their `Provision()` until the driver confirms it's done.

jsafrane:
Is it ready for review? I see failing tests. And please don't add `cmd/csi-snapshotter/csi-snapshotter` binary to git :-).

ggriffiths:
> Is it ready for review? I see failing tests. And please don't add `cmd/csi-snapshotter/csi-snapshotter` binary to git :-).

Yeah it's not quite ready yet. Still need to fix tests. Will mark it ready for review when it is :-) 

And thanks I'll remove it - added it accidentally

pkg/snapshotter/snapshotter.go
ggriffiths: Added an issue to fix this in a separate PR.

ggriffiths:
@msau42 @xing-yang tests are passing now and this PR is ready for review. I did not add Exponential retries to the `contentWorker/syncContent` flow, but I could add it in this PR or a future one if needed.

xing-yang:
Hi @ggriffiths, in the "Special notes for your reviewer" section of this PR description, can you please add links to the PRs from external-provisioner that you modeled after?  I assume they are these 2 commits in the following?  Thanks!

https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/commit/c19e26be3fd97c9f9d97a47d5198fdfa14a38aa9
https://github.com/kubernetes-csi/external-provisioner/commit/8203a03c47ce2b86a2a2c2421d74345b76183b14

pkg/snapshotter/snapshotter.go
```diff
@@ -75,15 +95,43 @@ func (s *snapshot) CreateSnapshot(ctx context.Context, snapshotName string, volu
...
+	if !ok {
+		// This is not gRPC error. The operation must have failed before gRPC
+		// method was called, otherwise we would get gRPC error.
+		// We don't know if any previous CreateVolume is in progress, be on the safe side.
+		return false
+	}
+	switch st.Code() {
+	case codes.Canceled, // gRPC: Client Application cancelled the request
+		codes.DeadlineExceeded,  // gRPC: Timeout
+		codes.Unavailable,       // gRPC: Server shutting down, TCP connection broken - previous CreateVolume() may be still in progress.
```
xing-yang: Change "Volume" to "Snapshot"?

ggriffiths:
Just updated the PR description with the 3 commits this is mostly modeled on @xing-yang.

cmd/csi-snapshotter/main.go
```diff
@@ -63,6 +63,9 @@ var (
 	csiAddress                      = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
 	createSnapshotContentRetryCount = flag.Int("create-snapshotcontent-retrycount", 5, "Number of retries when we create a snapshot content object for a snapshot.")
 	createSnapshotContentInterval   = flag.Duration("create-snapshotcontent-interval", 10*time.Second, "Interval between retries when we create a snapshot content object for a snapshot.")
+	retryIntervalStart              = flag.Duration("retry-interval-start", time.Second, "Initial retry interval of failed snapshot creation. It doubles with each failure, up to retry-interval-max.")
+	retryIntervalMax                = flag.Duration("retry-interval-max", 5*time.Minute, "Maximum retry interval of failed snapshot creation.")
+	failedSnapshotThreshold         = flag.Int("failed-snapshot-threshold", 15, "The maximum number of retries on snapshot failures. Set 0 to retry indefinitely. Default is 15.")
```
msau42: I don't think we should have a failed snapshot threshold. We should retry forever until user cancels the operation by deleting the VolumeSnapshot object.

pkg/controller/snapshot_controller.go
```diff
@@ -651,16 +668,16 @@ func (ctrl *csiSnapshotController) createSnapshotOperation(snapshot *crdv1.Volum
 	}
 
 	if err != nil {
-		return nil, err
+		return nil, snapshotter.SnapshottingFinished, err
```
msau42: should all of these be Background, since we now have a provider snapshot object, but haven't created a K8s object for it yet?

cmd/csi-snapshotter/main.go
ggriffiths: Removed `failedSnapshotThreshold` and made it retry forever until the snapshot object cannot be found. I add this check in  `updateSnapshot`.

pkg/controller/snapshot_controller.go
ggriffiths: Good point, updated all of these until SnapshottingInBackground

pkg/controller/snapshot_controller.go
```diff
@@ -364,17 +365,33 @@ func (ctrl *csiSnapshotController) createSnapshot(snapshot *crdv1.VolumeSnapshot
...
+			}
+
 			return err
 		}
+
+		// If no errors, update the snapshot.
 		_, updateErr := ctrl.storeSnapshotUpdate(snapshotObj)
 		if updateErr != nil {
 			// We will get an "snapshot update" event soon, this is not a big error
 			klog.V(4).Infof("createSnapshot [%s]: cannot update internal cache: %v", snapshotKey(snapshotObj), updateErr)
```
msau42: do we continue to retry updating in this case? @xing-yang

pkg/controller/snapshot_controller.go
```diff
@@ -611,30 +628,30 @@ func (ctrl *csiSnapshotController) checkandUpdateBoundSnapshotStatusOperation(sn
...
 // 3. Create the VolumeSnapshotContent object with the snapshot id information.
 // 4. Bind the VolumeSnapshot and VolumeSnapshotContent object
-func (ctrl *csiSnapshotController) createSnapshotOperation(snapshot *crdv1.VolumeSnapshot) (*crdv1.VolumeSnapshot, error) {
+func (ctrl *csiSnapshotController) createSnapshotOperation(snapshot *crdv1.VolumeSnapshot) (*crdv1.VolumeSnapshot, snapshotter.SnapshottingState, error) {
 	klog.Infof("createSnapshot: Creating snapshot %s through the plugin ...", snapshotKey(snapshot))
 
 	if snapshot.Status.Error != nil && !isControllerUpdateFailError(snapshot.Status.Error) {
 		klog.V(4).Infof("error is already set in snapshot, do not retry to create: %s", snapshot.Status.Error.Message)
-		return snapshot, nil
+		return snapshot, snapshotter.SnapshottingInBackground, nil
```
msau42: I think the errors before `CreateSnapshot` should be NoChange.  The reasoning behind this is: we haven't issued a new `CreateSnapshot` yet, however this could be our subsequent retry, so the state should remain the same as before.

ggriffiths: Changed these.

ggriffiths:
One thing to note is that I haven't added a `snapshotInformer` similar to the provisioner's `claimInformer`:
* https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/blob/master/controller/controller.go#L798-L807
* https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/blob/master/controller/controller.go#L963

I didn't think I would need to add this, but now I realize I'm not really accessing `snapshotsInProgress` as a result (see https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/blob/master/controller/controller.go#L963 for how this is used in the provisioner. Currently I'm only adding/removing it based on the SnapshottingState. @msau42

msau42:
I don't think you need to follow external-provisioner exactly.  I think the main thing that's missing is in `snapshotWorker()`: if you can't find the object from the informer (because it's been deleted), then try finding it from `snapshotsInProgress`.

Also cc @wongma7 since he is familiar with the provisioner design.

ggriffiths:
Sounds good, I've updated the PR to check snapshotsInProgress if we don't find one from the informer.

pkg/controller/snapshot_controller_base.go
```diff
@@ -230,6 +240,28 @@ func (ctrl *csiSnapshotController) snapshotWorker() {
...
+				return false
+			}
+
+			newSnapshot, err := ctrl.checkAndUpdateSnapshotClass(snapshot)
+			if err == nil {
+				klog.V(5).Infof("passed checkAndUpdateSnapshotClass for snapshot %q", key)
+				ctrl.updateSnapshot(newSnapshot)
+			}
+
+			ctrl.snapshotQueue.Forget(keyObj)
```
msau42: I don't think we want to forget and delete in this case.  We should probably follow the whole L230-235 section.  Can we inject the snapshotsInProgressCheck before L230 so they both go through the same codepath?

ggriffiths: Makes sense, updated to remove the forget/delete and reworked the informer/snapshotsInProgress checks.

ggriffiths:
@xing-yang should I redo this PR for a future snapshotter release? Now that the split controller work is done.

xing-yang:
> should I re-do this PR for a future snapshotter release? Now that the split controller work is done.

Yes, please do this after the 1.17 release.  I'm actually still working on improving the controller logic.

ggriffiths:
> > should I re-do this PR for a future snapshotter release? Now that the split controller work is done.
> 
> Yes, please do this after the 1.17 release. I'm actually still working on improving the controller logic.

Sounds good, will do :+1:
/assign ggriffiths

humblec:
@ggriffiths are you working on this rebase ? would like to see this  with `1.18` :)

xing-yang:
Hi @humblec, I think this PR is not needed any more.  In the Beta version of external-snapshotter, we made some changes.  I tested myself and don't see this issue any more.  That's why I asked the person who opened the issue to re-test.

humblec:
> Hi @humblec, I think this PR is not needed any more. In the Beta version of external-snapshotter, we made some changes. I tested myself and don't see this issue any more. That's why I asked the person who opened the issue to re-test.

Thanks @xing-yang for clarifying it.. Then lets double confirm and close this . :+1:

ggriffiths:
Closing this it was put on hold for the beta API changes last release. If this is still needed after testing #134 with the beta API, I'll submit a separate PR.

PR #156, opened 2019-08-09
Cherry-pick: Check if ListSnapshots is supported during GetSnapshotStatus

hakanmemisoglu:
**What type of PR is this?**

/kind bug

Cherry-picks a fix where external snaphotter fails in prebound snapshot case with a driver
that does not support `ListSnapshots` request.
Release 1.2 doesn't have the fix in the master. 

This is a cherry-pick from master.

```release-note
Cherry picks PR #138: Prebound snapshots will work correctly with CSI drivers that does not support ListSnasphots.
```

### DIFF
--- a/pkg/snapshotter/snapshotter.go
+++ b/pkg/snapshotter/snapshotter.go
@@ -101,13 +101,36 @@ func (s *snapshot) DeleteSnapshot(ctx context.Context, snapshotID string, snapsh
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


hakanmemisoglu:
Does this repo have something similar: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-release/cherry-picks.md

xing-yang:
No, we don't have a cherry-pick script.

PR #160, opened 2019-08-12
Update k8s.io dependencies to master to get selflink fix in client-go

msau42:
**What type of PR is this?**
/kind bug

SelfLink is being deprecated in https://github.com/kubernetes/kubernetes/pull/80978 and we need to update the client-go dependency to pick up a fix in `GetReference`.

Fixes #

```release-note
Fixes issue when SelfLink removal is turned on in Kubernetes.
```


msau42:
The tests passed, but are stuck in cluster teardown.  Seems to be a test infra issue

/test pull-kubernetes-csi-external-snapshotter 
/test pull-kubernetes-csi-external-snapshotter-alpha

Gopkg.toml
```diff
@@ -13,19 +13,29 @@ required = [
...
+# TODO: all the k8s.io dependencies are overrides due to
+# csi-lib-utils constraints. Once 1.16.0 is released, update
+# all of these versions back to a constraint.
+
 # The dependency on external-provisioner should be removed with #60.
-[[constraint]]
+[[override]]
   name = "k8s.io/api"
-  version = "kubernetes-1.14.0"
+  branch = "master"
```
xing-yang: Do we have to use master now?  Is there a specific version that we can use, i.e., v1.16.0-alpha.3?

msau42: No, the commit we need is after v1.16.0-alpha.3

xing-yang: Ok.  We should update this to a specific version when it becomes available later.

PR #166, opened 2019-08-24
WIP: Split the controller into common and sidecar controllers

xing-yang:
**What type of PR is this?**
/kind feature

This PR splits the snapshot controller into two, a common controller to be deployed on the controller node, and a sidecar controller to be deployed together with the CSI driver.

Fixes #

```release-note
There are now a common snapshot controller and a sidecar snapshot controller. Only the sidecar controller should be deployed with the CSI driver.
```


xing-yang:
Opened a new PR based on the new snapshot beta APIs.  Closing this one.

PR #175, opened 2019-10-09
Cherry-pick to 1.0: Verify PV/PVC binding and driver

xing-yang:
Backport PV/PVC binding change (#172) to release-1.0 branch.

Note: Resolved merge conflicts in the following files:
       pkg/controller/snapshot_controller.go
       pkg/controller/snapshot_create_test.go

**What type of PR is this?**
/kind bug

Fixed checking of binding of PVC when creating snapshots.

Fixes #

```release-note
Cherry pick PR #172: Added extra verification of source PersistentVolumeClaim before creating snapshot.
```

### DIFF
--- a/pkg/controller/snapshot_controller.go
+++ b/pkg/controller/snapshot_controller.go
@@ -844,9 +844,38 @@ func (ctrl *csiSnapshotController) getVolumeFromVolumeSnapshot(snapshot *crdv1.V
 		return nil, fmt.Errorf("failed to retrieve PV %s from the API server: %q", pvName, err)
 	}
 
+	// Verify binding between PV/PVC is still valid
+	bound := ctrl.IsVolumeBoundToClaim(pv, pvc)
+	if bound == false {
+		glog.Warningf("binding between PV %s and PVC %s is broken", pvName, pvc.Name)
+		return nil, fmt.Errorf("claim in dataSource not bound or invalid")
+	}
+
+	// Verify driver for PVC is the same as driver for VolumeSnapshot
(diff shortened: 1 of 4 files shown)


pkg/controller/snapshot_controller.go
```diff
@@ -844,11 +844,40 @@ func (ctrl *csiSnapshotController) getVolumeFromVolumeSnapshot(snapshot *crdv1.V
...
 
+	// Verify binding between PV/PVC is still valid
+	bound := ctrl.IsVolumeBoundToClaim(pv, pvc)
+	if bound == false {
+		glog.Warningf("binding between PV %s and PVC %s is broken", pvName, pvc.Name)
+		return nil, fmt.Errorf("claim in dataSource not bound or invalid")
+	}
+
+	// Verify driver for PVC is the same as driver for VolumeSnapshot
+	if pv.Spec.PersistentVolumeSource.CSI == nil || pv.Spec.PersistentVolumeSource.CSI.Driver != ctrl.snapshotterName {
```
msau42: Did we allow snapshotter name to be different from driver name in 1.0.x? If so, then we probably can't check this

xing-yang: "snapshotter" is an input option, however, we do check in the controller to make sure it is the same as the driver name:

https://github.com/kubernetes-csi/external-snapshotter/blob/release-1.0/pkg/controller/snapshot_controller_base.go#L318

msau42:
/hold
let me double check makefile first

msau42:
/hold cancel
makefile looks fine, when this merges, it will tag v1.0-canary

PR #183, opened 2019-10-21
Run only unit tests for now

xing-yang:
We have chicken and egg problems between external-snapshotter and hostpath csi driver as we're moving snapshots from alpha to beta and making changes in the way snapshots is deployed. So this PR temporarily disables other tests until the snapshot beta API PR is merged.

**What type of PR is this?**
/kind failing-test

This PR disable all other tests except the unit tests until
the snapshot beta API PR is merged.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.prow.sh
+++ b/.prow.sh
@@ -1,5 +1,7 @@
 #! /bin/bash
 
+CSI_PROW_TESTS="unit"
+
 . release-tools/prow.sh
 
 main


release-tools/prow.sh
```diff
@@ -242,7 +242,9 @@ configvar CSI_PROW_DEP_VERSION v0.5.1 "golang dep version to be used for vendor
 # thus only makes sense in repos which provide their own CSI
 # driver. Repos can enable sanity testing by setting
 # CSI_PROW_TESTS_SANITY=sanity.
-configvar CSI_PROW_TESTS "unit parallel serial parallel-alpha serial-alpha sanity" "tests to run"
+# NOTE(xyang): Run only unit tests until snapshot beta API PR is merged.
+configvar CSI_PROW_TESTS "unit" "tests to run"
```
msau42: instead of modifying this, I would modify https://github.com/kubernetes-csi/external-snapshotter/blob/master/.prow.sh instead to override this variable

Issue #186, opened 2019-10-24
Move APIs and generated clients outside pkg


msau42:
It could also be exported as a go module.

yuxiangqian:
this would need another to resolve

xing-yang:
I think the snapshot-controller should stay with the APIs and generated clients in the same repo.  Otherwise there's no way to validate the APIs and we'll have more problems trying to sync up 3 repos.

msau42:
Yes I think same repo but separate go modules would be great

msau42:
Here's an example of csi-proxy creating a separate go module for their client library: https://github.com/kubernetes-csi/csi-proxy/tree/master/client

xing-yang:
Thanks!  Will take a look.

Kartik494:
> Yes I think same repo but separate go modules would be great

Hi @msau42 According to this we don't have to move apis and generated clients outside pkg?
Rather within pkg we have to create go module for apis and generated client like this [client library](https://github.com/kubernetes-csi/csi-proxy/tree/master/client) Please let me know regarding this 
Thanks

msau42:
While technically you don't need to move it out of pkg, I think it makes more sense from a code organization perspective to move it out to a top level client

Kartik494:
Thanks,  i will follow up on this.

xing-yang:
This is done.

PR #202, opened 2019-12-02
Update client-go

bertinatto:
**What type of PR is this?**

/kind feature

This PR updates `k8s.io/client-go` to the intermediate `kubernetes-1.17.0-rc.1`
version. This surfaced another problem, though: the auto-generated code
had to be updated as well. To achieve that, I imported
`k8s.io/code-generator` first and then executed `hack/update-generated-code.sh`.

I needed this change in order to use the new `v1.CSINode` in the
external-provisioner, which depends on the snapshotter.

```release-note
NONE
```


msau42:
One idea I had in https://github.com/kubernetes-csi/external-snapshotter/issues/186 was to separate the API + generated clients into a separate go module. I think that will help a lot with fixing dependency issues on the snapshotter

xing-yang:
Code is generated differently now.  @yuxiangqian PTAL.
/hold

pkg/apis/volumesnapshot/v1beta1/zz_generated.deepcopy.go
```diff
@@ -2,23 +2,26 @@
...
+
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Code generated by controller-gen. DO NOT EDIT.
+// Code generated by deepcopy-gen. DO NOT EDIT.
```
xing-yang: We need to use controller-gen, not deepcopy-gen.

bertinatto:
As we discussed in today's meeting, closing this in favour of https://github.com/kubernetes-csi/external-snapshotter/pull/202#issuecomment-560465403.

/close

PR #203, opened 2019-12-03
update to client-go

### LINKED ISSUE #200: Update client-go to kubernetes-1.16.3
We use VolumeSnapshot CRD api and clients in our [Stash](https://github.com/stashed) project. We are trying to use the 1.16.3 client libraries. But we are blocked on this project, since it uses 1.14 client-go. Can you please update the client-go to 1.16.3 ?

Thanks.

yuxiangqian:
1. update clientset, deepcopy using code-generator
2. add a dummy file tools.go to force "go mod vendor" to see
code-generator as dependencies
3. add a script to update CRD
4. add a README to document CRD updating steps

**What type of PR is this?**
/kind feature

https://github.com/kubernetes-csi/external-snapshotter/pull/202
https://github.com/kubernetes-csi/external-snapshotter/issues/200

Fixes #200 

Separating api/clientset into different modules will be in a different PR

NONE


yuxiangqian:
cc @msau42

yuxiangqian:
cc @bertinatto

hack/README
```diff
@@ -0,0 +1,13 @@
+This README documents how to generate CRD yaml files using ./hack/update-crd.sh
+after updating /pkg/apis/volumesnapshot/v1beta1/types.go
+
+step 1: run ./hack/update-crd.sh from top directory, new yaml files should have
+        been created under ./config/crd
+step 2: add the github PR annotation to any yaml files updated into metadata
+        section of the generated type, i.e.,
+				api-approved.kubernetes.io: "https://github.com/kubernetes-csi/external-snapshotter/pull/139"
+step 3: remove any metadata sections from the yaml file which does not belong
```
msau42: example?

hack/update-crd.sh
```diff
@@ -0,0 +1,45 @@
...
+set -o pipefail
+
+## find or download controller-gen
+CONTROLLER_GEN=$(which controller-gen)
+
+if [ "$CONTROLLER_GEN" = "" ]
+then
+  TMP_DIR=$(mktemp -d);
+  cd $TMP_DIR;
+  go mod init tmp;
```
msau42: what is this step supposed to do?

hack/update-crd.sh
```diff
@@ -0,0 +1,45 @@
...
+  go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.4;
```
msau42: does this version have to match with k8s.io/code-generator?

hack/README
```diff
@@ -0,0 +1,13 @@
+This README documents how to generate CRD yaml files using ./hack/update-crd.sh
```
msau42: when do you run "update-generated-code.sh"?

Can we potentially add a make rule that does both?

hack/update-crd.sh
yuxiangqian: this step allows "go get" to work under the temp module folder.

hack/update-crd.sh
yuxiangqian: controller-gen is a different tool to code-generator, and they theoretically should work independently. both will operate on types.go.

hack/README
yuxiangqian: absolutely, I will actually include a make rule in next PR to separate modules.
update-generated-code.sh should follow the similar pattern, it needs to run after any changes to types.go.

msau42: Since both scripts are under the same folder, maybe put in instructions for both?  Since it's very likely that if you change types.go, you'll want to regenerate both

yuxiangqian: SG. I will add it.
BTW, is it a standard to have a .md file for README? or its fine to leave it plain text.

hack/README
yuxiangqian: will add.

hack/README
msau42: I think md would be useful

hack/README
```diff
@@ -0,0 +1,13 @@
...
+after updating /pkg/apis/volumesnapshot/v1beta1/types.go
+
+step 1: run ./hack/update-crd.sh from top directory, new yaml files should have
+        been created under ./config/crd
+step 2: add the github PR annotation to any yaml files updated into metadata
+        section of the generated type, i.e.,
+				api-approved.kubernetes.io: "https://github.com/kubernetes-csi/external-snapshotter/pull/139"
```
xing-yang: Can we have multiple PR annotations like this?  For example, when we update it next time, should we still keep this PR 139 and add a new PR annotation on top of this?

hack/README
```diff
@@ -0,0 +1,13 @@
...
+step 3: remove any metadata sections from the yaml file which does not belong
+        to the generated type. TODO(xiangqian): this is to make sure the
+				generated CRD is compatibile with apiextensions.k8s.io/v1. Once
+				controller-gen supports generating CRD with apiextensions.k8s.io/v1,
+				switch to use the correct version of controller-gen and remove step3.
```
xing-yang: s/step3/step  3

hack/update-crd.sh
```diff
@@ -0,0 +1,45 @@
...
+
+if [ "$CONTROLLER_GEN" = "" ]
+then
+  echo "ERROR: failed to get controller-gen";
+  exit 1;
+fi
+
+SCRIPT_ROOT=$(unset CDPATH && cd $(dirname "${BASH_SOURCE[0]}")/.. && pwd)
+
+$CONTROLLER_GEN crd:trivialVersions=true,preserveUnknownFields=false paths=${SCRIPT_ROOT}/pkg/apis/volumesnapshot/v1beta1
```
xing-yang: What does trivialVersions mean?

yuxiangqian: trivialVersions tells controller-gen not to generate multi-version CRD, in our case, we do not need it.

hack/README
yuxiangqian: no we should not, I have updated the . README to indicate that

hack/README
yuxiangqian: updated

hack/README.md
```diff
@@ -0,0 +1,34 @@
+# Scripts User Guild
```
msau42: Guide

msau42:
/release-note-none
/approve
will let @xing-yang lgtm

One nit: can you separate the vendor update into a 2nd commit?

PR #206, opened 2019-12-09
Update README

### LINKED ISSUE #154: Update README to indicate controller architecture changes
1. Common controller will watch VolumeSnapshotContent and bind VolumeSnapshotContent with VolumeSnapshot.
2. Snapshotter will watch VolumeSnapshot to invoke corresponding gRPC call, create VolumeSnapshotContent.
3. Deletion of a snapshot should follow similar pattern

xing-yang:
**What type of PR is this?**
/kind documentation

Fixes #154

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,71 +1,149 @@
+[![Build Status](https://travis-ci.org/kubernetes-csi/external-snapshotter.svg?branch=master)](https://travis-ci.org/kubernetes-csi/external-snapshotter)
+
 # CSI Snapshotter
 
-The CSI external-snapshotter is part of Kubernetes implementation of [Container Storage Interface (CSI)](https://github.com/container-storage-interface/spec).
+The CSI snapshotter is part of Kubernetes implementation of [Container Storage Interface (CSI)](https://github.com/container-storage-interface/spec).
 
-The volume snapshot feature supports CSI v1.0 and it has been an Alpha feature in Kubernetes since v1.12.
+The volume snapshot feature supports CSI v1.0 and higher. It was introduced as an Alpha feature in Kubernetes v1.12 and has been promoted to an Beta feature in Kubernetes 1.17.
+
 
 ## Overview
 
-CSI Snapshotter is an external controller that watches Kubernetes Snapshot CRD objects and triggers CreateSnapshot/DeleteSnapshot against a CSI endpoint. Full design can be found at Kubernetes proposal at [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/csi-snapshot.md)
+With the promotion of Volume Snapshot to beta, the feature is now enabled by default on standard Kubernetes deployments instead of being opt-in.
+
+The move of the Kubernetes Volume Snapshot feature to beta also means:
+* A revamp of volume snapshot APIs.
+* The CSI external-snapshotter sidecar is split into two controllers, a snapshot controller and a CSI external-snapshotter sidecar.
+
+The snapshot controller is deployed by the Kubernetes distributions and is responsible for watching the VolumeSnapshot CRD objects and manges the creation and deletion lifecycle of snapshots.
+
+The CSI external-snapshotter sidecar watches Kubernetes VolumeSnapshotContent CRD objects and triggers CreateSnapshot/DeleteSnapshot against a CSI endpoint.
+
+Blog post for the beta feature can be found [here](https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-feature-cis-volume-snapshot-beta)
+
+
+## Compatibility
+
+This information reflects the head of this branch.
+
+| Compatible with CSI Version                                                                | Container Image             | Min K8s Version | Snapshot CRD version |
+| ------------------------------------------------------------------------------------------ | ----------------------------| --------------- | -------------------- |
+| [CSI Spec v1.0.0](https://github.com/container-storage-interface/spec/releases/tag/v1.0.0) | quay.io/k8scsi/csi-snapshotter | 1.17         | v1beta1              |
+| [CSI Spec v1.0.0](https://github.com/container-storage-interface/spec/releases/tag/v1.0.0) | quay.io/k8scsi/snapshot-controller | 1.17     | v1beta1              |
+
+
+## Feature Status
+
+The `VolumeSnapshotDataSource` feature gate was introduced in Kubernetes 1.12 and it is enabled by default in Kubernetes 1.17 when the volume snapshot feature is promoted to beta.
+
 
 ## Design
 
-External snapshotter follows [controller](https://github.com/kubernetes/community/blob/master/contributors/devel/controllers.md) pattern and uses informers to watch for `VolumeSnapshot` and `VolumeSnapshotContent` create/update/delete events. It filters out these objects with `Snapshotter==<CSI driver name>` specified in the associated VolumeSnapshotClass object and then processes these events in workqueues with exponential backoff.
+Both the snapshot controller and CSI external-snapshotter sidecar follow [controller](https://github.com/kubernetes/community/blob/master/contributors/devel/controllers.md) pattern and uses informers to watch for events. The snapshot controller watches for `VolumeSnapshot` and `VolumeSnapshotContent` create/update/delete events.
 
-### Snapshotter
+The CSI external-snapshotter sidecar only watches for `VolumeSnapshotContent` create/update/delete events. It filters out these objects with `Driver==<CSI driver name>` specified in the associated VolumeSnapshotClass object and then processes these events in workqueues with exponential backoff.
 
-Snapshotter talks to CSI over socket (/run/csi/socket by default, configurable by -csi-address). The snapshotter then:
+The CSI external-snapshotter sidecar talks to CSI over socket (/run/csi/socket by default, configurable by -csi-address).
 
-* Discovers the supported snapshotter name by `GetDriverName` call.
+### Hightlights in the snapshot v1beta1 APIs
 
-* Uses ControllerGetCapabilities for find out if CSI driver supports `ControllerServiceCapability_RPC_CREATE_DELETE_SNAPSHOT` and `ControllerServiceCapability_RPC_LIST_SNAPSHOTS` calls. Otherwise, the controller will not start.
+* DeletionPolicy is a required field in both VolumeSnapshotClass and VolumeSnapshotContent. This way the user has to explicitly specify it, leaving no room for confusion.
+* VolumeSnapshotSpec has a required Source field. Source may be either a PersistentVolumeClaimName (if dynamically provisioning a snapshot) or VolumeSnapshotContentName (if pre-provisioning a snapshot).
+* VolumeSnapshotContentSpec has a required Source field. This Source may be either a VolumeHandle (if dynamically provisioning a snapshot) or a SnapshotHandle (if pre-provisioning volume snapshots).
+* VolumeSnapshot contains a Status to indicate the current state of the volume snapshot. It has a field BoundVolumeSnapshotContentName to indicate the VolumeSnapshot object is bound to a VolumeSnapshotContent.
+* VolumeSnapshotContent contains a Status to indicate the current state of the volume snapshot content. It has a field SnapshotHandle to indicate that the VolumeSnapshotContent represents a snapshot on the storage system.
 
-* Processes new/updated/deleted `VolumeSnapshots`: The snapshotter only processes `VolumeSnapshot` that has `snapshotter` specified in its `VolumeSnapshotClass` matches its driver name. The process workflow is as follows
-  * If the snapshot status is `Ready`, the controller checks whether the snapshot and its content still binds correctly. If there is any problem with the binding (e.g., snapshot points to a non-exist snapshot content), update the snapshot status and emit event.
-  * If the snapshot status is not ready, there are two cases.
-    * `SnapshotContentName` is not empty: the controller verifies whether the snapshot content exists and also binds to the snapshot. If verification passes, the controller binds the snapshot and its content objects and marks it is ready. Otherwise, it updates the error status of the snapshot.
-    * `SnapshotContentName` is set empty: the controller will first check whether there is already a content object which binds the snapshot correctly with snapshot uid (`VolumeSnapshotRef.UID`) specified. If so, the controller binds these two objects. Otherwise, the controller issues a create snapshot operation. Please note that if the error status shows that snapshot creation already failed before, it will not try to create snapshot again.
-
-* Processes new/updated/deleted `VolumeSnapshotContents`: The snapshotter only processes `VolumeSnapshotContent` in which the CSI driver specified in the spec matches the controller's driver name.
-  * If the `VolumeSnapshotRef` is set to nil, skip this content since it is not bound to any snapshot object.
-  * Otherwise, the controller verifies whether the content object is correctly bound to a snapshot object. In case the `VolumeSnapshotRef.UID` is set but it does not match its snapshot object or snapshot no long exists, the content object and its associated snapshot will be deleted.
 
 ## Usage
 
-### Running on command line
+The Volume Snapshot feature now depends on a new, volume snapshot controller in addition to the volume snapshot CRDs. Both the volume snapshot controller and the CRDs are independent of any CSI driver. Regardless of the number CSI drivers deployed on the cluster, there must be only one instance of the volume snapshot controller running and one set of volume snapshot CRDs installed per cluster.
 
-For debugging, it is possible to run snapshotter on command line. For example,
+Therefore, it is strongly recommended that Kubernetes distributors bundle and deploy the controller and CRDs as part of their Kubernetes cluster management process (independent of any CSI Driver).
 
-```bash
-csi-snapshotter -kubeconfig ~/.kube/config -v 5 -csi-address /run/csi/socket
-```
+If your cluster does not come pre-installed with the correct components, you may manually install these components by executing the following steps.
 
-### Running in a statefulset
+Install Snapshot Beta CRDs:
+* kubectl create -f config/crd
+* https://github.com/kubernetes-csi/external-snapshotter/tree/master/config/crd
+* Do this once per cluster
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,12 +2,19 @@
...
 
 ## Overview
 
 CSI Snapshotter is an external controller that watches Kubernetes Snapshot CRD objects and triggers CreateSnapshot/DeleteSnapshot against a CSI endpoint. Full design can be found at Kubernetes proposal at [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/csi-snapshot.md)
 
+With the promotion of Volume Snapshot to beta, the feature is now enabled by default on standard Kubernetes deployments instead of being opt-in.
+
+The move of the Kubernetes Volume Snapshot feature to beta also means:
+* A revamp of volume snapshot APIs.
+* The CSI external-snapshotter sidecar is split into two controllers, a common snapshot controller and a CSI external-snapshotter sidecar.
```
msau42: I would remove the word "common" and just call it snapshot controller

README.md
```diff
@@ -30,23 +37,30 @@ Snapshotter talks to CSI over socket (/run/csi/socket by default, configurable b
   * If the `VolumeSnapshotRef` is set to nil, skip this content since it is not bound to any snapshot object.
   * Otherwise, the controller verifies whether the content object is correctly bound to a snapshot object. In case the `VolumeSnapshotRef.UID` is set but it does not match its snapshot object or snapshot no long exists, the content object and its associated snapshot will be deleted.
 
-## Usage
+## Deployment
```
msau42: Do we also want to provide some recommended RBACs for users (not sidecar, not controller) to operate on the CRs?

README.md
xing-yang: Removed

README.md
xing-yang: Does that include admins?  Users can only see VolumeSnapshot, Admins can see both.

msau42: Correct, I mean provide different recommended rbacs for different personas

README.md
```diff
@@ -1,52 +1,75 @@
...
+The snapshot controller is deployed by the Kubernetes distributions and is responsible for watching the VolumeSnapshot CRD objects.
 
-### Snapshotter
+The CSI external-snapshotter sidecar watches Kubernetes VolumeSnapshotContent CRD objects and triggers CreateSnapshot/DeleteSnapshot against a CSI endpoint.
 
-Snapshotter talks to CSI over socket (/run/csi/socket by default, configurable by -csi-address). The snapshotter then:
+Blog post for the beta feature can be found [here](https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-feature-cis-volume-snapshot-beta)
 
-* Discovers the supported snapshotter name by `GetDriverName` call.
+## Controller Workflow
```
msau42: How much of this is relevant to a vendor looking to integrate these components, and how much can go into a "design.md" (or even KEP) file?

README.md
```diff
@@ -55,12 +78,24 @@ If you want to run external-snapshotter with higher availability, you can enable
 --leader-election=true
 ```
 
+### Upgrade from Alpha to Beta
+
+The change from Alpha to Beta snapshot APIs is not backward compatible.
+
+If you have already deployed Alpha snapshot APIs and controller and want to upgrade to Beta, you need to do the following:
```
msau42: You can also set the deletepolicy to retain, delete the alpha objects, and then import them as pre-existing snapshots using the beta APIs

README.md
```diff
@@ -1,52 +1,75 @@
-# CSI Snapshotter
+# CSI Snapshot Feature
```
msau42: Can you also mirror the README format that the other sidecars have?

README.md
```diff
@@ -1,52 +1,75 @@
...
 
-The CSI external-snapshotter is part of Kubernetes implementation of [Container Storage Interface (CSI)](https://github.com/container-storage-interface/spec).
+The CSI snapshot feature is part of Kubernetes implementation of [Container Storage Interface (CSI)](https://github.com/container-storage-interface/spec).
```
msau42: I see a few main audiences for this readme:

1. Users wanting to use the snapshot feature.
2. CSI driver authors wanting to integrate snapshot support for their driver
3. Kubernetes distributions/deployers that want to support snapshots.
4. Developers making changes to the components.

For 1), I would point them to the K8s docs.

And then clearly state the audience for the rest of this README, which I think should be 2) and 3).

For 4), we should have another doc like "DEVELOPING.md"

README.md
xing-yang: Most of this section is actually already in the blog post.  I added it here because in the old README, there's some design overview on how the controller is implemented.  I can just remove this and refer to the blog which is already linked.  I can also submit a KEP later if that's preferred.

README.md
xing-yang: Sure.  I'll take a look of README of other sidecars.  I think the old README mirrored the one in external-attacher.

README.md
msau42: I think we should capture detailed design and workflows somewhere other than a blog post. I think KEP is a natural place for this

README.md
xing-yang: With Beta, I'll have to talk about the snapshot controller and the CSI snapshotter.  So it won't be exactly the same as other sidecars.

README.md
```diff
@@ -1,66 +1,145 @@
...
-External snapshotter follows [controller](https://github.com/kubernetes/community/blob/master/contributors/devel/controllers.md) pattern and uses informers to watch for `VolumeSnapshot` and `VolumeSnapshotContent` create/update/delete events. It filters out these objects with `Snapshotter==<CSI driver name>` specified in the associated VolumeSnapshotClass object and then processes these events in workqueues with exponential backoff.
 
-### Snapshotter
+## Compatibility
 
-Snapshotter talks to CSI over socket (/run/csi/socket by default, configurable by -csi-address). The snapshotter then:
+This information reflects the head of this branch.
 
-* Discovers the supported snapshotter name by `GetDriverName` call.
+| Compatible with CSI Version                                                                | Container Image             | Min K8s Version |
```
msau42: Do we also need a column for what versions of the snapshot CRDs it supports?

README.md
```diff
@@ -1,66 +1,145 @@
...
-for i in $(find deploy/kubernetes -name '*.yaml'); do kubectl create -f $i; done
-```
+Install CSI Driver:
+* Follow instructions provided by your CSI Driver vendor.
+* Here is an example to install the sample hostpath CSI driver
+  * kubectl create -f deploy/kubernetes/csi-snapshotter
+  * https://github.com/kubernetes-csi/external-snapshotter/tree/master/deploy/kubernetes/csi-snapshotter
 
-### Running with Leader Election
+### Important changes in the snapshot beta APIs
```
msau42: clarify from alpha1v1-> beta1v1

README.md
```diff
@@ -1,66 +1,145 @@
...
+* `--leader-election`: Enables leader election. This is useful when there are multiple replicas of the same snapshot controller running for the same Kubernetes deployment. Only one of them may be active (=leader). A new leader will be re-elected when current leader dies or becomes unresponsive for ~15 seconds.
+
+* `--leader-election-namespace <namespace>`: The namespace where the leader election resource exists. Defaults to the pod namespace if not set.
+
+#### Other recognized arguments
+* `--kubeconfig <path>`: Path to Kubernetes client configuration that the snapshot controller uses to connect to Kubernetes API server. When omitted, default token provided by Kubernetes will be used. This option is useful only when the snapshot controller does not run as a Kubernetes pod, e.g. for debugging.
+
+* `--resync-period <duration>`: Internal resync interval when the snapshot controller re-evaluates all existing `VolumeSnapshot` instances and tries to fulfill them, i.e. create / delete corresponding snapshots. It does not affect re-tries of failed calls! It should be used only when there is a bug in Kubernetes watch logic.
```
msau42: is there a default?

README.md
```diff
@@ -1,66 +1,145 @@
...
+
+* `--create-snapshotcontent-retrycount`: Number of retries when we create a snapshot content object for a snapshot. Default is 5.
```
msau42: does that mean we give up entirely after 5? I thought we changed it to retry forever until user cancels the operation

README.md
msau42: should this go into changelog instead?

README.md
```diff
@@ -1,66 +1,145 @@
...
+
+* `--leader-election`: Enables leader election. This is useful when there are multiple replicas of the same external-snapshotter running for one CSI driver. Only one of them may be active (=leader). A new leader will be re-elected when current leader dies or becomes unresponsive for ~15 seconds.
+
+* `--leader-election-namespace <namespace>`: The namespace where the leader election resource exists. Defaults to the pod namespace if not set.
+
+* `--timeout <duration>`: Timeout of all calls to CSI driver. It should be set to value that accommodates majority of `CreateSnapshot`, `DeleteSnapshot, and `ListSnapshots` calls. 1 minute is used by default.
+
+* `snapshot-name-prefix`: Prefix to apply to the name of a created snapshot. Default is `snapshot`.
+
+* `snapshot-name-uuid-length`: Length in characters for the generated uuid of a created snapshot. Defaults behavior is to NOT truncate.
```
msau42: Do we make sure the length doesn't exceed the CSI spec max? especially when prefix is added?

README.md
```diff
@@ -1,66 +1,145 @@
...
+### CSI external snapshotter sidecar command line options
+
+#### Important optional arguments that are highly recommended to be used
+* `--csi-address <path to CSI socket>`: This is the path to the CSI driver socket inside the pod that the external-snapshotter container will use to issue CSI operations (`/run/csi/socket` is used by default).
+
+* `--leader-election`: Enables leader election. This is useful when there are multiple replicas of the same external-snapshotter running for one CSI driver. Only one of them may be active (=leader). A new leader will be re-elected when current leader dies or becomes unresponsive for ~15 seconds.
+
+* `--leader-election-namespace <namespace>`: The namespace where the leader election resource exists. Defaults to the pod namespace if not set.
+
+* `--timeout <duration>`: Timeout of all calls to CSI driver. It should be set to value that accommodates majority of `CreateSnapshot`, `DeleteSnapshot, and `ListSnapshots` calls. 1 minute is used by default.
```
msau42: closing quote missing after DeleteSnapshot

README.md
```diff
@@ -1,66 +1,145 @@
...
+
+* All glog / klog arguments are supported, such as `-v <log level>` or `-alsologtostderr`.
+
+
+## Upgrade from Alpha to Beta
+
+The change from Alpha to Beta snapshot APIs is not backward compatible.
+
+If you have already deployed Alpha snapshot APIs and controller and want to upgrade to Beta, you need to do the following:
+* Delete snapshots created using Alpha snapshot CRDs and controller.
```
msau42: what does "controller" refer to in the alpha sense?

README.md
```diff
@@ -1,66 +1,145 @@
...
---leader-election=true
-```
 
 ## Testing
 
 Running Unit Tests:
 
 ```bash
-go test -timeout 30s  github.com/kubernetes-csi/external-snapshotter/pkg/controller
+go test -timeout 30s  github.com/kubernetes-csi/external-snapshotter/pkg/common-controller
```
msau42: does "make test" accomplish the same?

README.md
xing-yang: This retry count is used for creating VolumeSnapshotContent API objects.  We didn't talk about retrying this one forever.

The logic we are trying to change is for creating a physical snapshot on the storage system itself.  Grant started a PR which followed what Jan has done for the external-provisioner.  Due to the beta splitting controller work, that work is delayed.  That involves big controller logic changes which I do not want to add on top of the splitting controller changes in this release.  We can revisit that after we cut this release.

msau42: If API server is temporarily down, then we will give up on trying to bind or create a snapshot? I'm not sure we want that? Also keep in mind especially now that this is beta, we need to support the argument according to the deprecation policy.

xing-yang: I think we copied it from the PV controller. Looks like PV controller retries creating PV object 5 times by default, but createProvisionedPVRetryCount is a const defined inside the PV controller.  Should we get rid of these command line options and use a const in the snapshot controller instead?

https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/persistentvolume/pv_controller.go#L131
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/persistentvolume/pv_controller.go#L1489

msau42: Yeah I think that would be better. It gives more flexibility to change the behavior later.

I think PV controller scenario is slightly different than snapshot. For snapshots, I think we have stronger idempotency/naming guarantees, so it can retry the calls with the same id, which helps (but not completely) avoids the leaking issue. So I'm not sure how necessary the retries in snapshot controller should be. cc @jsafrane

msau42: Actually, does the scenario apply at all? Now we create the Content object before calling CSI. This avoids the leaking scenario right?

xing-yang: Yes, you are right.

README.md
xing-yang: Sure.  I can add that.  I actually have that in my first draft of the README but removed them when I submitted the PR because I thought maybe the workflow that @yuxiangqian is working on can better address this.

README.md
xing-yang: "controller" here refers to the alpha version external-snapshotter sidecar controller.

README.md
xing-yang: Sure, I'll add this to the changelog, but I still want to highlight some of these in this doc.  Let me rephrase a little bit.

README.md
xing-yang: sure

README.md
xing-yang: Yes, 60 sec

README.md
xing-yang: Removed these from README.  Will submit a PR that removes them from cmd line options and uses a const internally instead.

README.md
xing-yang: Fixed

README.md
xing-yang: The CSI spec says length of the name conforms to the length limit of an unicode string.  The code we have to handle the snapshot name in the snapshotter is similar to how the volume name is handled in the provisioner.

https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L319
https://github.com/kubernetes-csi/external-provisioner/blob/master/README.md#other-recognized-arguments

README.md
```diff
@@ -1,66 +1,150 @@
...
+## Upgrade from v1alpha1 to v1beta1
+
+The change from v1alpha1 to v1beta1 snapshot APIs is not backward compatible.
+
+If you have already deployed v1alpha1 snapshot APIs and external-snapshotter sidecar controller and want to upgrade to v1beta1, you need to do the following:
+* Delete volume snapshots created using v1alpha1 snapshot CRDs and external-snapshotter sidecar controller.
+* Uninstall v1alpha1 snapshot CRDs, external-snapshotter sidecar controller, and CSI driver.
+* Install v1beta1 snapshot CRDs, snapshot controller, CSI external-snapshotter sidecar and CSI driver.
+
+If you want to keep the physical snapshot resources created using the v1alpha1 version, here are suggested steps.
```
xing-yang: @msau42 @yuxiangqian please review this section to make sure it is okay.

README.md
xing-yang: "make test" will do more than just running unit tests.  It will also do go mod, go fmt, etc.

README.md
xing-yang: @bswartz Here is the upgrade section.  PTAL.

README.md
msau42: CSI Spec says 128 max: https://github.com/container-storage-interface/spec/blob/master/spec.md#size-limits.

Let's followup on this as a separate issue?

README.md
```diff
@@ -1,66 +1,150 @@
...
+Note: There is a possibility for data loss if something goes wrong during this process. Verify steps 1 and 2 works for a sample snapshot before moving forward. There was a bug in the pre-provisioned snapshots logic that was fixed in release [v1.2.1](https://github.com/kubernetes-csi/external-snapshotter/blob/release-1.2/CHANGELOG-1.2.md#notable-changes-1). Make sure you are using v1.2.1 or later before upgrading to v1beta1 if you want to try the following steps.
+1. Change deletionPolicy from Delete to Retain in the volume snapshot content.
+2. Delete volume snapshots created using v1alpha1 snapshot CRDs and external-snapshotter sidecar controller. Verify that volume snapshot content and the physical snapshot resource remain.
+3. Make sure above steps work for a sample snapshot before moving forward. This is to make sure that deletion of a volume snapshot object will not trigger the deletion of the volume snapshot content object and its associated physical snapshot resource on the storage system when the deletionPolicy is Retain. It is very important to verify this step to make sure there is no data loss.
```
msau42: Maybe put this note as 1), since you want them to do this before they actually try the steps

README.md
```diff
@@ -1,66 +1,150 @@
...
+4. Uninstall v1alpha1 snapshot CRDs, external-snapshotter sidecar controller, and CSI driver. The v1alpha1 volume snapshot content and its associated physical snapshot resource will not be in the Kubernetes database any more.
+5. Install v1beta1 snapshot CRDs, snapshot controller, CSI external-snapshotter sidecar, and CSI drvier.
+6. Create a v1beta1 volume snapshot content API object for the physical snapshot resource.
+  * Note: Do not set the UID field of VolumeSnapshotRef! If you copied a volume snapshot content yaml file from your v1alpha1 setup, make sure to DELETE the UID field of VolumeSnapshotRef. This field will be populated by the controller automatically after a v1beta1 volume snapshot API object is created and bound to the volume snapshot content.
```
msau42: are there any other things like annotations that they should also remove?

xing-yang:
> CSI Spec says 128 max: https://github.com/container-storage-interface/spec/blob/master/spec.md#size-limits.

> Let's followup on this as a separate issue?

Sure.  I'll open an issue to track.

README.md
xing-yang: Will remove this section as I have not tested it myself.

README.md
xing-yang: Sure

xing-yang: Opened an issue here: https://github.com/kubernetes-csi/external-snapshotter/issues/216

PR #210, opened 2019-12-14
sidecar controller update, fix potential snapshot leaking

### LINKED ISSUE #209: Potential leaking of snapshot on underlying storage system when deletion fails
In sidecar controller [here](https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/sidecar-controller/snapshot_controller.go#L72), it's possible that the scheduled CSI deletion operation in another goroutine fails. The content finalizer should not be removed until the CSI deletion operation success.

yuxiangqian:
**What type of PR is this?**
/kind cleanup
/kind bug

clean up sidecar logic, fix potential snapshot leaking issue, added more test cases

Fixes #209 

NONE
```release-note
None
```

### DIFF
--- a/pkg/sidecar-controller/snapshot_controller.go
+++ b/pkg/sidecar-controller/snapshot_controller.go
@@ -23,7 +23,7 @@ import (
 
 	crdv1 "github.com/kubernetes-csi/external-snapshotter/pkg/apis/volumesnapshot/v1beta1"
 	"github.com/kubernetes-csi/external-snapshotter/pkg/utils"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
 	"k8s.io/kubernetes/pkg/util/goroutinemap"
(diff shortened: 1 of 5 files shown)


yuxiangqian:
cc @jingxu97

pkg/sidecar-controller/snapshot_controller.go
```diff
@@ -57,73 +57,50 @@ func (ctrl *csiSnapshotSideCarController) syncContent(content *crdv1.VolumeSnaps
...
-				// Remove the content finalizer.
-				klog.V(5).Infof("syncContent: Content [%s] is a deletion candidate. Remove finalizer.", content.Name)
-				return ctrl.removeContentFinalizer(content)
-			}
-
-		case crdv1.VolumeSnapshotContentDelete:
-			klog.V(4).Infof("VolumeSnapshotContent[%s]: policy is Delete. Delete physical snapshot", content.Name)
-			err = ctrl.deleteCSISnapshot(content)
-			if err != nil {
-				return err
```
xing-yang: If deleteCSISnapshot fails here, we return error and finalizer won't be removed.  It won't be leaking?

yuxiangqian: deleteCSISnapshot spawns a goroutine to carry on deletion, and this is non-blocking. so yes, if scheduling deletion operation fails, it will not remove the finalizer,
but once the operation has been scheduled successfully, it will remove the finalizer regardless the execution of deletion. thats when it causes leaking

pkg/sidecar-controller/snapshot_controller.go
```diff
@@ -358,6 +335,7 @@ func (ctrl *csiSnapshotSideCarController) deleteCSISnapshotOperation(content *cr
 
 	_, snapshotterCredentials, err := ctrl.getCSISnapshotInput(content)
 	if err != nil {
+		ctrl.eventRecorder.Event(content, v1.EventTypeWarning, "SnapshotDeleteError", "Failed to get snapshot class")
```
xing-yang: getCSISnapshotInput could fail other than cannot get snapshot class.

s/Failed to get snapshot class/Failed to get snapshot deletion input.

pkg/sidecar-controller/snapshot_controller.go
```diff
@@ -511,8 +527,6 @@ func (ctrl csiSnapshotSideCarController) removeContentFinalizer(content *crdv1.V
 	if err != nil {
 		klog.Errorf("failed to update content store %v", err)
 	}
-
-	klog.V(5).Infof("Removed protection finalizer from volume snapshot content %s", content.Name)
```
xing-yang: Why moving this to the beginning when finalizer is not removed yet?

xing-yang:
Have you tested this?  Is there a visible delay in content deletion?

pkg/sidecar-controller/snapshot_controller.go
yuxiangqian: right, will put it back. Merge issue I think

yuxiangqian:
> Have you tested this? Is there a visible delay in content deletion?

It's kinda of hard to reproduce the leak issue. deletion via CSI happens really quick in real drivers like PD. I have, though, added unit testing to mimic driver error, and saw the finalizer is still there not removed.

PR #211, opened 2019-12-17
Remove createSnapshotContentRetryCount and createSnapshotContentInterval

xing-yang:
**What type of PR is this?**
/kind cleanup

Removes createSnapshotContentRetryCount and createSnapshotContentInterval
from command line options.

Fixes #

This was brought up during the review of the README PR (https://github.com/kubernetes-csi/external-snapshotter/pull/206) that these parameters should just be internal variables, not command line options.

```release-note
Removes createSnapshotContentRetryCount and createSnapshotContentInterval
from command line options.
```

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -337,10 +337,16 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
 		}
 
 		// update snapshot status
-		klog.V(5).Infof("syncUnreadySnapshot [%s]: trying to update snapshot status", utils.SnapshotKey(snapshot))
-		_, err = ctrl.updateSnapshotStatus(snapshot, newContent)
-		if err != nil {
+		for i := 0; i < ctrl.createSnapshotContentRetryCount; i++ {
+			klog.V(5).Infof("syncUnreadySnapshot [%s]: trying to update snapshot status", utils.SnapshotKey(snapshot))
+			_, err = ctrl.updateSnapshotStatus(snapshot, newContent)
+			if err == nil {
+				break
+			}
 			klog.V(4).Infof("failed to update snapshot %s status: %v", utils.SnapshotKey(snapshot), err)
+		}
+
+		if err != nil {
 			// update snapshot status failed
 			ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotStatusUpdateFailed", fmt.Sprintf("Snapshot status update failed, %v", err))
 			return err
@@ -393,10 +399,16 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
 				}
 
 				// Update snapshot status with BoundVolumeSnapshotContentName
-				klog.V(5).Infof("syncUnreadySnapshot [%s]: trying to update snapshot status", utils.SnapshotKey(snapshot))
-				_, err = ctrl.updateSnapshotStatus(snapshot, content)
-				if err != nil {
+				for i := 0; i < ctrl.createSnapshotContentRetryCount; i++ {
+					klog.V(5).Infof("syncUnreadySnapshot [%s]: trying to update snapshot status", utils.SnapshotKey(snapshot))
+					_, err = ctrl.updateSnapshotStatus(snapshot, content)
+					if err == nil {
+						break
+					}
 					klog.V(4).Infof("failed to update snapshot %s status: %v", utils.SnapshotKey(snapshot), err)
+				}
+
+				if err != nil {
 					// update snapshot status failed
 					ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotStatusUpdateFailed", fmt.Sprintf("Snapshot status update failed, %v", err))
 					return err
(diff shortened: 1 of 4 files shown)


pkg/common-controller/snapshot_controller_base.go
```diff
@@ -42,6 +42,12 @@ import (
 	"k8s.io/kubernetes/pkg/util/goroutinemap"
 )
 
+// Number of retries when we create a snapshot content object for a snapshot.
+const createSnapshotContentRetryCount = 5
```
msau42: my higher level question is why do we need to retry within the workqueue func? Can we just re-add the snapshot to the queue for retry whenever there's a failure, and let the workqueue properly throttle the retries?

xing-yang: I thought this is recommended practice for API update?  PV controller also retries when PV object fails to be created.  I remember when I first started to work on snapshot in external-storage repo, I was asked to add this type of retry as a bug fix.

msau42: I believe PV controller does it because of the possibility of leaking the volume. I don't think we have the same issue here since we are creating the content object first before calling CSI. Cc @jsafrane

jsafrane: > I don't think we have the same issue here since we are creating the content object first before calling CSI

+1, in PV controller, a volume is already created on storage backend when the controller saves a PV. And it really needs to save it, otherwise the volume may be leaked.

In snapshots, nothing is leaked if SnapshotContent is not created, right?

xing-yang: Yes, in the beta version, we changed the behavior and create the VolumeSnapshotContent first before creating the physical snapshot resource.

We are also doing retries when update snapshot fails.  In general, is it recommended practice to do retries when update an API object fails?

msau42: The general pattern is to requeue it on any failure and let workqueue schedule the retry.

xing-yang: Removed the RetryCount.

pkg/common-controller/snapshot_controller.go
```diff
@@ -337,16 +337,10 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
...
-			klog.V(4).Infof("failed to update snapshot %s status: %v", utils.SnapshotKey(snapshot), err)
-		}
-
+		klog.V(5).Infof("syncUnreadySnapshot [%s]: trying to update snapshot status", utils.SnapshotKey(snapshot))
+		_, err = ctrl.updateSnapshotStatus(snapshot, newContent)
 		if err != nil {
+			klog.V(4).Infof("failed to update snapshot %s status: %v", utils.SnapshotKey(snapshot), err)
 			// update snapshot status failed
 			ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotStatusUpdateFailed", fmt.Sprintf("Snapshot status update failed, %v", err))
 			return err
```
msau42: Looking at the caller, I see that we don't requeue the object and instead rely on resync period to retry. I think we should change the behavior to requeue, but we should also make sure the workqueue has proper throttling/backoff as well.

xing-yang: Use queue.AddRateLimited(key)?

xing-yang: Can you take a look of this? https://github.com/kubernetes-csi/external-snapshotter/pull/211/commits/10017c2bdccc95268596aefdcbdc769bc5b2dc3c

pkg/common-controller/snapshot_controller_base.go
```diff
@@ -365,8 +362,10 @@ func (ctrl *csiSnapshotCommonController) updateSnapshot(snapshot *crdv1.VolumeSn
 			klog.V(3).Infof("could not sync claim %q: %+v", utils.SnapshotKey(snapshot), err)
```
msau42: we don't want to return err in this case either?

pkg/common-controller/snapshot_controller.go
msau42: Yes, I like this direction. Also cc @yuxiangqian and @jsafrane for more eyes.  This will cause snapshotter to potentially retry more frequently than it has in the past, although it should be rate limited.

xing-yang: I wonder if we should delay this change.  This changes snapshot controller behavior a lot.  In the past we don't do retries when create snapshot fails because snapshot is time sensitive (other than a few retries when creating/updating API objects). 

Can we revisit this after the 2.0 release?  We have a pending fix for the create snapshot timeout issue as well that we should go back to after 2.0.

msau42: Ok let's open an issue to track it.

xing-yang: Sure

xing-yang: Opened an issue.  Let me remove this re-queue logic from this PR.

yuxiangqian: A little concerned about removing retrying logic for status update and snapshot content creation from application perspective. Do we have a rough idea how often retry would resolve API server failures? Application could end up based of the existence of status field of a snapshot to decide whether or not to unquiesce. With RateLimited requeue, the waiting time could be significantly longer?

xing-yang: Removed snapshotQueue.AddRateLimited().

What is conclusion here?  Should I just add back createSnapshotContentRetryCount() as a internal const or is the current PR okay?  @msau42

msau42: I don't like having retries within retries. It makes timing hard to reason about (metrics look strange), and we can end up overloading the API server since there's two different throttling behaviors that need to be controlled.  We previously had issues in external-provisioner due to it having multiple retries within retries and I'd like to avoid that here too. We can tune the rate limiting as necessary.

But I agree this is a large, fundamental change to how the snapshot controller has operated, so I'm ok deferring it to the next release.

msau42: Yes, let's add the retry count back for now so the only change is not making the retry count configurable.

xing-yang: Sure.

PR #217, opened 2019-12-20
Add a warning msg in the upgrade process

xing-yang:
**What type of PR is this?**
/kind documentation

This PR adds a warning message in the upgrade process that the underlying snapshots will be deleted.

Fixes #

```release-note
NOTE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,9 +124,10 @@ Install CSI Driver:
 The change from v1alpha1 to v1beta1 snapshot APIs is not backward compatible.
 
 If you have already deployed v1alpha1 snapshot APIs and external-snapshotter sidecar controller and want to upgrade to v1beta1, you need to do the following:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -125,6 +125,7 @@ The change from v1alpha1 to v1beta1 snapshot APIs is not backward compatible.
 
 If you have already deployed v1alpha1 snapshot APIs and external-snapshotter sidecar controller and want to upgrade to v1beta1, you need to do the following:
 * Delete volume snapshots created using v1alpha1 snapshot CRDs and external-snapshotter sidecar controller.
+  * <strong>Note: The underlying snapshots on the storage system will be deleted after this step!!!</strong>
```
msau42: I think it's also important to point out that if they skip this step, the next step will still delete all the objects and potentially the snapshots too

xing-yang: I'll just move the warning to the beginning to say that snapshots will be deleted to be safe.

PR #226, opened 2019-12-30
Small fixes to snapshot-controller YAML

bswartz:
**What type of PR is this?**
/kind cleanup

Remove "sidecar" language for the controller.
Update image tag to 2.0.0-rc4 (the latest)

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
+++ b/deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
@@ -85,7 +85,7 @@ spec:
               mountPath: /csi
         - name: csi-snapshotter
           # NOTE: replace with official image when released: quay.io/k8scsi/csi-snapshotter:v2.0.0
-          image: quay.io/k8scsi/csi-snapshotter:v2.0.0-rc2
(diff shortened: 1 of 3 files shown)


xing-yang:
Please also change the csi-snapshotter image to rc4.  Also change the imagePullPolicy for both csi-snapshotter and snapshot-controller from "IfNotPresent" to "Always".

bswartz:
> Please also change the csi-snapshotter image to rc4. Also change the imagePullPolicy for both csi-snapshotter and snapshot-controller from "IfNotPresent" to "Always".

Why change the imagePullPolicy? IMO, always pulling is wasteful of resources (and slows down the system) and only makes sense when using an image tag that will be recycled (like latest/canary, but not 2.0.0-rc4).

xing-yang:
> Why change the imagePullPolicy? IMO, always pulling is wasteful of resources (and slows down the system) and only makes sense when using an image tag that will be recycled (like latest/canary, but not 2.0.0-rc4).

This is to be consistent with other image pulling policies.  I changed it from "Always" to "IfNotPresent" during testing and forgot to change it back.  This is to make sure the image is always cleanly downloaded from the official repo.  You are change it to "IfNotPresent" during your testing, but the checked-in version should be "Always".

Issue #225, opened 2019-12-30
Random delay before readyToUse=true


bswartz:
My CSI plugin always returns readyToUse=true, because it simply blocks in CreateSnapshot() until the snapshot is created (typically 1 second or less). Usually the volumesnapshot object in k8s reflects readyToUse=true immediately, but with some randomness it sometimes shows up as readyToUse=false, and only get corrected after about a minute.

Here is a log that illustrates this happening:
[external-snapshotter.log](https://github.com/kubernetes-csi/external-snapshotter/files/4009023/external-snapshotter.log)
 
Notice at 05:10:00, the CreateSnapshot() RPC returns success with readyToUse=true. However there's an error on line 96 of the log:

snapshot_controller.go:325] error updating volume snapshot content status for snapshot snapcontent-d7f6b159-fd33-4f57-9084-21c9a12a691b: snapshot controller failed to update snapcontent-d7f6b159-fd33-4f57-9084-21c9a12a691b on API server: Operation cannot be fulfilled on volumesnapshotcontents.snapshot.storage.k8s.io "snapcontent-d7f6b159-fd33-4f57-9084-21c9a12a691b": the object has been modified; please apply your changes to the latest version and try again.

48 seconds later, the controller retries, and successfully updates the object. 

I have 2 issues with this behavior (1) why was readyToUse _ever_ set to false if the CreateSnapshot() RPC returned readyToUse=true on the first try? And (2) it seems that the long wait time before retrying is unneeded because it's just an API race with something else modifying the same snapshotcontent object. We could just retry the update right after the error, or requeue the operation for very soon after instead of waiting. 48 seconds is a long time to wait in an automated sequence of steps that's waiting for the snapshot to be usable.

xing-yang:
Is this tested with my fix https://github.com/kubernetes-csi/external-snapshotter/pull/224?
Can you provide corresponding logs from snapshot-controller?

xing-yang:
Can you also check the status of content: "kubectl describe volumesnapshotcontent"?  Is ReadyToUse in content already set to "true" while ReadyToUse in volumesnapshot is still "false"?  Is it a long delay until they are in sync?  I noticed this behavior early on and fixed it.  If it still happens to you, I need to see logs from both sidecar and snapshot controller.

bswartz:
[...]

PR #230, opened 2020-01-02
Fix the requeue logic

xing-yang:
**What type of PR is this?**
/kind bug

Add requeue to the sidecar controller.

Part of #225

```release-note
This fixes the re-queue logic so a failed snapshot operation will be added back to a rate limited queue for retries.
```

### DIFF
--- a/pkg/sidecar-controller/framework_test.go
+++ b/pkg/sidecar-controller/framework_test.go
@@ -409,24 +409,6 @@ func (r *snapshotReactor) getChangeCount() int {
 	return r.changedSinceLastSync
 }
 
-// waitForIdle waits until all tests, controllers and other goroutines do their
-// job and no new actions are registered for 10 milliseconds.
-func (r *snapshotReactor) waitForIdle() {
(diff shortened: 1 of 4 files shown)


humblec:
unless I missed something the logic here looks good to me.  However we will wait for @bswartz as the issue is reproducible in the setup and can have some real validation. considering this is on cache and sync code path, its better to get validated.

bswartz:
I downloaded this patch and ran it in my torture tests and I don't see a single slow snapshot anymore. Unfortunately my torture test eventually crashed after 35 snapshots due to bugs in other components, but I am confident that this patch solves my problem.

Thanks @xing-yang !

xing-yang:
Thanks @bswartz for testing this!

bswartz:
Okay I fixed the bug in my torture test, and re-ran it with this patch. I got up to 278 snapshots with no hangs and no crashes, before I decided to end the test.

xing-yang:
That's awesome.  Thanks @bswartz!

PR #232, opened 2020-01-03
Remove PVC finalizer when no longer in use

### LINKED ISSUE #231: PVC Finalizer should be removed when not in use
This is a bug introduced by the latest code refactoring.  PVC finalizer should not be tied to the deletionTimestamp on the snapshot.  It should always be removed when there are no snapshots actively using it as a source.

xing-yang:
PVC Finalizer is intended to prevent PVC from being deleted when there is a snapshot actively being created from it.  After the snapshot is created successfully and in ReadyToUse status, and if there are no other snapshots actively being created from this PVC, the PVC finalizer should be removed.

This is a bug introduced by the latest code refactoring. Without this fix, the current code only removes the PVC finalizer when the snapshot has a deletionTimestamp.  The fix here is to move the code that removes the PVC finalizer outside of the function that handles the snapshot with a deletionTimestamp in syncSnapshot().  This way PVC finalizer will always be removed when it is not needed actively as a source.

**What type of PR is this?**
/kind bug

Fixes #231 

```release-note
NONE
```


ggriffiths:
Looks good! 
/lgtm

Issue #196, opened 2019-11-21
Improve code coverage for pkg/common-controller


ggriffiths:
Need to improve code coverage after the split controller refactor.
/assign @ggriffiths 
/cc @xing-yang 
/kind cleanup

ggriffiths:
Still need to revisit and update the following:
- pkg/common-controller/snapshot_create_test.go (a few need to be re-visited)
- pkg/common-controller/snapshot_delete_test.go (most need to be re-written)
- pkg/common-controller/snapshot_update_test.go (a few need to be re-visited)

PR #254, opened 2020-02-12
Refactor common controller tests to use withXYZ functions and add tests

### LINKED ISSUE #253: Refactor common-controller tests to use test model similar to in-tree PV tests
/assign @ggriffiths

ggriffiths:

**What type of PR is this?**
/kind cleanup

refactor finalizer-specific test to use withSnapshotFinalizers function

Fixes #253 
Part of #196 

n/a

```release-note
NONE
```

### DIFF
--- a/pkg/common-controller/framework_test.go
+++ b/pkg/common-controller/framework_test.go
@@ -164,11 +164,13 @@ type reactorError struct {
 	error    error
 }
 
-func withSnapshotFinalizer(snapshot *crdv1.VolumeSnapshot) *crdv1.VolumeSnapshot {
-	snapshot.ObjectMeta.Finalizers = append(snapshot.ObjectMeta.Finalizers, utils.VolumeSnapshotContentFinalizer)
-	snapshot.ObjectMeta.Finalizers = append(snapshot.ObjectMeta.Finalizers, utils.VolumeSnapshotAsSourceFinalizer)
-	snapshot.ObjectMeta.Finalizers = append(snapshot.ObjectMeta.Finalizers, utils.VolumeSnapshotBoundFinalizer)
-	return snapshot
+func withSnapshotFinalizers(snapshots []*crdv1.VolumeSnapshot, finalizers ...string) []*crdv1.VolumeSnapshot {
+	for i := range snapshots {
+		for _, f := range finalizers {
+			snapshots[i].ObjectMeta.Finalizers = append(snapshots[i].ObjectMeta.Finalizers, f)
+		}
+	}
+	return snapshots
 }
 
 func withContentFinalizer(content *crdv1.VolumeSnapshotContent) *crdv1.VolumeSnapshotContent {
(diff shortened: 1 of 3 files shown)


ggriffiths:
Also just added more tests as well. Coverage for snapshot_controller.go is now 74.1%, up from 68.2%.

pkg/common-controller/framework_test.go
```diff
@@ -863,7 +876,7 @@ func newContentWithUnmatchDriverArray(contentName, boundToSnapshotUID, boundToSn
 func newSnapshot(
 	snapshotName, snapshotUID, pvcName, targetContentName, snapshotClassName, boundContentName string,
 	readyToUse *bool, creationTime *metav1.Time, restoreSize *resource.Quantity,
-	err *crdv1.VolumeSnapshotError, nilStatus bool, withFinalizer bool, deletionTimestamp *metav1.Time) *crdv1.VolumeSnapshot {
+	err *crdv1.VolumeSnapshotError, nilStatus bool, withAllFinalizers bool, deletionTimestamp *metav1.Time) *crdv1.VolumeSnapshot {
```
xing-yang: I thought the test framework in pv/pvc controller will allow you to create a basic newSnapshot() without passing in the withAllFinalizers flag, and then we can have another function withAllFinalizers() which takes a created snapshot object as input and apply finalizers on top of it.  So that means we shouldn't need "withAllFinalizers bool" as an input parameter for newSnapshot().

pkg/common-controller/snapshot_delete_test.go
```diff
@@ -18,13 +18,14 @@ package common_controller
 
 import (
 	//"errors"
+
```
xing-yang: Remove //"errors"

pkg/common-controller/snapshot_update_test.go
```diff
@@ -18,9 +18,11 @@ package common_controller
 
 import (
 	//"errors"
+	"errors"
```
xing-yang: Remove //"errors" on line 20

pkg/common-controller/framework_test.go
ggriffiths: I left "withFinalizers" as "withAllFinalizers" for the purpose of reducing the refactor required. Some tests require all finalizers while others require no finalizers, so pretty much every test would have the "withSnapFinalizers" wrapper. 

I mainly introduced "func withSnapshotFinalizers()" to fine-tune which finalizers are added. i.e. for cases where only one finalizer is wanted.

pkg/common-controller/framework_test.go
```diff
@@ -904,18 +917,18 @@ func newSnapshot(
 			VolumeSnapshotContentName: &targetContentName,
 		}
 	}
-	if withFinalizer {
-		return withSnapshotFinalizer(&snapshot)
+	if withAllFinalizers {
+		return withSnapshotFinalizers([]*crdv1.VolumeSnapshot{&snapshot}, utils.VolumeSnapshotContentFinalizer, utils.VolumeSnapshotAsSourceFinalizer, utils.VolumeSnapshotBoundFinalizer)[0]
```
xing-yang: withSnapshotFinalizers() only adds finalizers on VolumeSnapshot API object, right?  Why is utils.VolumeSnapshotContentFinalizer needed here?  utils.VolumeSnapshotContentFinalizer is added on VolumeSnapshotContent only.

pkg/common-controller/framework_test.go
```diff
@@ -164,16 +164,17 @@ type reactorError struct {
 	error    error
 }
 
-func withSnapshotFinalizer(snapshot *crdv1.VolumeSnapshot) *crdv1.VolumeSnapshot {
-	snapshot.ObjectMeta.Finalizers = append(snapshot.ObjectMeta.Finalizers, utils.VolumeSnapshotContentFinalizer)
```
xing-yang: Ok, I see that utils.VolumeSnapshotContentFinalizer is added to VolumeSnapshot API object here.  This is wrong.  We only need utils.VolumeSnapshotContentFinalizer for VolumeSnapshotContent, not for VolumeSnapshot.

ggriffiths: ok, removed the contentFinalizer.

pkg/common-controller/framework_test.go
ggriffiths: ok, removed the contentFinalizer.

PR #265, opened 2020-03-04
doc: add workaround for hack/update-generated-code.sh

### LINKED ISSUE #239: Incorrect repository layour for v2 Go Modules
It is impossible to use repository with Go Modules because it breaks the mandatory requirements for `v2`.

It works with `v1.2.2` but not with `v2.0.0`.
https://sum.golang.org/lookup/github.com/kubernetes-csi/external-snapshotter@v1.2.2

```bash
$ go get -v github.com/kubernetes-csi/external-snapshotter@v2.0.0
go: finding github.com/kubernetes-csi/external-snapshotter v2.0.0
go: finding github.com/kubernetes-csi/external-snapshotter v2.0.0
go: errors parsing go.mod
go.mod:10: require github.com/kubernetes-csi/external-snapshotter: version "v2.0.0" invalid: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

Details and requirements:
- https://golang.org/cmd/go/#hdr-Module_compatibility_and_semantic_versioning
- https://blog.golang.org/v2-go-modules

alex1989hu:
****What type of PR is this?**
/kind documentation

Add missing workaround to hack/update-generated-code.sh

Fixes #239

As per reviewer's request https://github.com/kubernetes-csi/external-snapshotter/issues/239#issuecomment-577985400, added the steps to let users to use update-generated-code.sh..
```release-note
NONE
```

### DIFF
--- a/hack/README.md
+++ b/hack/README.md
@@ -10,7 +10,35 @@ This is the script to update clientset/informers/listers and API deepcopy code u
 
 Make sure to run this script after making changes to /pkg/apis/volumesnapshot/v1beta1/types.go.
 
-To run this script, simply run: ./hack/update-generated-code.sh from the project root directory.
+To run this script, you have to patch it:
+```patch
+diff --git a/hack/update-generated-code.sh b/hack/update-generated-code.sh
+--- a/hack/update-generated-code.sh
++++ b/hack/update-generated-code.sh
+@@ -27,7 +27,7 @@ CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${SCRIPT_ROOT}; ls -d -1 ./vendor/k8s.io/code-ge
+ #                  k8s.io/kubernetes. The output-base is needed for the generators to output into the vendor dir
+ #                  instead of the $GOPATH directly. For normal projects this can be dropped.
+ bash ${CODEGEN_PKG}/generate-groups.sh "deepcopy,client,informer,lister" \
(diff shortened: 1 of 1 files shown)


yuxiangqian:
some nits and lgtm.
@alex1989hu what would be the path forward? should we re-org the whole repo and copy files into v2?

alex1989hu:
> some nits and lgtm.
> @alex1989hu what would be the path forward? should we re-org the whole repo and copy files into v2?

The long term goal is to drop `Go dep` compatibility - the reason why we had to create a `v2` directory is to being compatible with. If we drop the compatibility, then we can remove `v2` directory at the same time.

hack/README.md
```diff
@@ -10,7 +10,35 @@ This is the script to update clientset/informers/listers and API deepcopy code u
 
 Make sure to run this script after making changes to /pkg/apis/volumesnapshot/v1beta1/types.go.
 
-To run this script, simply run: ./hack/update-generated-code.sh from the project root directory.
+To run this script, you have to patch it:
```
xing-yang: Can you add some words here to explain why we have to go thru these steps to generate code and reference this issue https://github.com/kubernetes-csi/external-snapshotter/issues/239.

xing-yang: I'm going to merged this PR.
@alex1989hu can you submit a followup PR to address this comment.

PR #275, opened 2020-03-13
Allows VolumeSnapshot to be deleted if the class isn't found

huffmanca:
**What type of PR is this?**
/kind flake

There were some instances where a namespace remained after a test due to a VolumeSnapshot persisted after the associated objects (VolumeSnapshotContent/PVC) and VolumeSnapshotClass had been deleted. If the cluster entered into this state, then the namespace would remain until the VolumeSnapshot was manually removed.

This PR allows a VolumeSnapshot that's been scheduled for deletion to be successfully deleted if the VolumeSnapshotClass cannot be found.

N/A

```release-note
NONE
```

### DIFF
--- a/pkg/common-controller/snapshot_controller_base.go
+++ b/pkg/common-controller/snapshot_controller_base.go
@@ -26,7 +26,7 @@ import (
 	storagelisters "github.com/kubernetes-csi/external-snapshotter/v2/pkg/client/listers/volumesnapshot/v1beta1"
 	"github.com/kubernetes-csi/external-snapshotter/v2/pkg/utils"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -215,8 +215,11 @@ func (ctrl *csiSnapshotCommonController) snapshotWorker() {
 			// The volume snapshot still exists in informer cache, the event must have
 			// been add/update/sync
 			newSnapshot, err := ctrl.checkAndUpdateSnapshotClass(snapshot)
-			if err == nil {
-				klog.V(5).Infof("passed checkAndUpdateSnapshotClass for snapshot %q", key)
+			if err == nil || (newSnapshot.ObjectMeta.DeletionTimestamp != nil && errors.IsNotFound(err)) {
+				// If the VolumeSnapshotClass is not found, we still need to process an update
+				// so that syncSnapshot can delete the snapshot, should it still exist in the
+				// cluster after it's been removed from the informer cache
+				klog.V(5).Infof("updating snapshot %q; snapshotClass may have already been removed", key)
 				ctrl.updateSnapshot(newSnapshot)
 			}
 			return false
@@ -243,7 +246,10 @@ func (ctrl *csiSnapshotCommonController) snapshotWorker() {
 			return false
 		}
 		newSnapshot, err := ctrl.checkAndUpdateSnapshotClass(snapshot)
-		if err == nil {
+		if err == nil || errors.IsNotFound(err) {
+			// We should still handle deletion events even if the VolumeSnapshotClass
+			// is not found in the cluster
+			klog.V(5).Infof("deleting snapshot %q; snapshotClass may have already been removed", key)
 			ctrl.deleteSnapshot(newSnapshot)
 		}
 		return false
@@ -329,7 +335,8 @@ func (ctrl *csiSnapshotCommonController) checkAndUpdateSnapshotClass(snapshot *c
 		if err != nil {
 			klog.Errorf("checkAndUpdateSnapshotClass failed to getSnapshotClass %v", err)
 			ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "GetSnapshotClassFailed", fmt.Sprintf("Failed to get snapshot class with error %v", err))
-			return nil, err
+			// we need to return the original snapshot even if the class isn't found, as it may need to be deleted
+			return newSnapshot, err
 		}
 	} else {
 		klog.V(5).Infof("checkAndUpdateSnapshotClass [%s]: SetDefaultSnapshotClass", snapshot.Name)
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -23,7 +23,7 @@ import (
 
 	crdv1 "github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
 	"github.com/kubernetes-csi/external-snapshotter/v2/pkg/utils"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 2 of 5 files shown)


huffmanca:
@xing-yang - this should be ready for review

pkg/common-controller/snapshot_controller_base.go
```diff
@@ -218,6 +218,15 @@ func (ctrl *csiSnapshotCommonController) snapshotWorker() {
 			if err == nil {
 				klog.V(5).Infof("passed checkAndUpdateSnapshotClass for snapshot %q", key)
 				ctrl.updateSnapshot(newSnapshot)
+			} else if newSnapshot.ObjectMeta.DeletionTimestamp != nil && errors.IsNotFound(err) {
+				// If the VolumeSnapshotClass is not found, we still need to process an update
```
xing-yang: If this is a delete event, can we let the logic on line 256 to handle it?
https://github.com/kubernetes-csi/external-snapshotter/pull/275/files#diff-77f790654ab434b7aac6ae9b88ecd4b2R256

huffmanca: The issue that I've seen here is that the VolumeSnapshot would be queued for deletion, but the VolumeSnapshotClass would be deleted before all the associated objects (PVC/VolumeSnapshotContent) are removed from the cluster.

In this event, the snapshot still exists in the cluster, so it falls into the condition at line 214. There was an error retrieving the class from line 218, and then we exited the loop at line 231 only to repeat this again. I updated the conditions at line 255 so that this could be handled if it's caught in time, and if not it'll get deleted during the next update.

xing-yang: How about add a check for `if newSnapshot.ObjectMeta.DeletionTimestamp != nil && errors.IsNotFound(err)`.  Only `return false` on line 231 if the check is false.  This way it can be handled by the logic on line 256.

huffmanca: I updated the current check to include this. This should still work, and I haven't been able to reproduce it after running ~20 times (previously it would manifest ~5-10% of the time).

huffmanca:
I also removed the second commit and submitted it as a separate PR in https://github.com/kubernetes-csi/external-snapshotter/pull/278 .

pkg/common-controller/snapshot_controller_base.go
```diff
@@ -215,9 +215,16 @@ func (ctrl *csiSnapshotCommonController) snapshotWorker() {
...
+			if err == nil || (newSnapshot.ObjectMeta.DeletionTimestamp != nil && errors.IsNotFound(err)) {
+				// If the VolumeSnapshotClass is not found, we still need to process an update
+				// so that syncSnapshot can delete the snapshot, should it still exist in the
+				// cluster after it's been removed from the informer cache
 				klog.V(5).Infof("passed checkAndUpdateSnapshotClass for snapshot %q", key)
 				ctrl.updateSnapshot(newSnapshot)
+			} else if newSnapshot.ObjectMeta.DeletionTimestamp != nil && errors.IsNotFound(err) {
+
+				klog.V(5).Infof("snapshotClass not found and snapshot %q is scheduled for deletion", key)
+				ctrl.updateSnapshot(newSnapshot)
```
xing-yang: This is already handled by the code block at 218-223.

huffmanca: You're correct. I accidentally left this in; it should now be removed.

pkg/common-controller/snapshot_controller_base.go
```diff
@@ -243,7 +246,7 @@ func (ctrl *csiSnapshotCommonController) snapshotWorker() {
 			return false
 		}
 		newSnapshot, err := ctrl.checkAndUpdateSnapshotClass(snapshot)
-		if err == nil {
+		if err == nil || errors.IsNotFound(err) {
```
xing-yang: Can you add a note here to say that we still need to delete snapshot if snapshot class is not found?

huffmanca: I've included a note here. Should there be a log message printed as well? Debug level?

xing-yang: Yes

pkg/common-controller/snapshot_controller_base.go
```diff
@@ -215,7 +215,10 @@ func (ctrl *csiSnapshotCommonController) snapshotWorker() {
 			// The volume snapshot still exists in informer cache, the event must have
 			// been add/update/sync
 			newSnapshot, err := ctrl.checkAndUpdateSnapshotClass(snapshot)
-			if err == nil {
+			if err == nil || (newSnapshot.ObjectMeta.DeletionTimestamp != nil && errors.IsNotFound(err)) {
+				// If the VolumeSnapshotClass is not found, we still need to process an update
+				// so that syncSnapshot can delete the snapshot, should it still exist in the
+				// cluster after it's been removed from the informer cache
```
xing-yang: The message on line 222 should be updated.

huffmanca: This message has been updated to also include the possibility that the VolumeSnapshotClass has been deleted.

pkg/common-controller/snapshot_controller_base.go
huffmanca: I've broken this conditional into two separate statements and provided different messages for each.

pkg/common-controller/snapshot_controller_base.go
```diff
@@ -244,6 +247,12 @@ func (ctrl *csiSnapshotCommonController) snapshotWorker() {
 		}
 		newSnapshot, err := ctrl.checkAndUpdateSnapshotClass(snapshot)
 		if err == nil {
+			klog.V(5).Infof("deleting snapshot %q", key)
+			ctrl.deleteSnapshot(newSnapshot)
+		} else if errors.IsNotFound(err) {
+			// We should still handle deletion events even if the VolumeSnapshotClass
+			// is not found in the cluster
+			klog.V(5).Infof("snapshot %q has been scheduled for deletion and the snapshotClass has already been removed", key)
```
xing-yang: Can you combine these two messages too?

huffmanca: Messages have been combined into a single conditional block.

pkg/common-controller/snapshot_controller_base.go
```diff
@@ -243,7 +246,10 @@ func (ctrl *csiSnapshotCommonController) snapshotWorker() {
 			return false
 		}
 		newSnapshot, err := ctrl.checkAndUpdateSnapshotClass(snapshot)
-		if err == nil {
+		if err == nil || errors.IsNotFound(err) {
+			// We should still handle deletion events even if the VolumeSnapshotClass
+			// is not found in the cluster
+			klog.V(5).Infof("deleting snapshot %q ; snapshot has either been scheduled for deletion, or snapshotClass has already been removed", key)
```
xing-yang: I think "snapshot has either been scheduled for deletion" can be removed as both condition will go thru "deleteSnapshot".
Suggest change to the following:
deleting snapshot %q; snapshotClass may have already been removed

huffmanca: This has been updated to use the suggested message.

pkg/common-controller/snapshot_controller_base.go
```diff
@@ -215,8 +215,11 @@ func (ctrl *csiSnapshotCommonController) snapshotWorker() {
 			// The volume snapshot still exists in informer cache, the event must have
 			// been add/update/sync
 			newSnapshot, err := ctrl.checkAndUpdateSnapshotClass(snapshot)
-			if err == nil {
-				klog.V(5).Infof("passed checkAndUpdateSnapshotClass for snapshot %q", key)
+			if err == nil || (newSnapshot.ObjectMeta.DeletionTimestamp != nil && errors.IsNotFound(err)) {
+				// If the VolumeSnapshotClass is not found, we still need to process an update
+				// so that syncSnapshot can delete the snapshot, should it still exist in the
+				// cluster after it's been removed from the informer cache
+				klog.V(5).Infof("passed checkAndUpdateSnapshotClass for snapshot %q, or snapshot needs to be deleted and snapshotClass doesn't exist", key)
```
xing-yang: Suggest change to the following:
updating snapshot %q; snapshotClass may have already been removed

huffmanca: This has been changed to use the suggested message.

PR #278, opened 2020-03-18
Include a nil check for the snapshotContent

huffmanca:
/kind bug

It's currently possible for the snapshotController to crash when checking snapshot Finalizers if the VolumeSnapshotContent is nil. This PR introduces a simple nil check to prevent the panic.

```release-note
NONE
```

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -23,7 +23,7 @@ import (
 
 	crdv1 "github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
 	"github.com/kubernetes-csi/external-snapshotter/v2/pkg/utils"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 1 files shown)


pkg/common-controller/snapshot_controller.go
```diff
@@ -266,8 +266,10 @@ func (ctrl *csiSnapshotCommonController) checkandRemoveSnapshotFinalizersAndChec
 		}
 
 		if !inUse {
-			klog.V(5).Infof("checkandRemoveSnapshotFinalizersAndCheckandDeleteContent: Set VolumeSnapshotBeingDeleted annotation on the content [%s]", content.Name)
-			ctrl.setAnnVolumeSnapshotBeingDeleted(content)
+			if content != nil {
+				klog.V(5).Infof("checkandRemoveSnapshotFinalizersAndCheckandDeleteContent: Set VolumeSnapshotBeingDeleted annotation on the content [%s]", content.Name)
+				ctrl.setAnnVolumeSnapshotBeingDeleted(content)
```
xing-yang: This is called in multiple places.  Can you add a nil check inside setAnnVolumeSnapshotBeingDeleted()?
I actually just ran into this yesterday so I added it here:
https://github.com/kubernetes-csi/external-snapshotter/pull/261/files#diff-214a2fc68717c170415a3578536d3b74R1281

I'll remove my change once your PR is merged.

huffmanca: I've updated this to also include the check inside of `setAnnVolumeSnapshotBeingDeleted()`. I was hitting the panic on line 270 above, when we were printing out the debug statement.

I can remove the name of the content, since it should print this out within `setAnnVolumeSnapshotBeingDeleted` immediately after it checks for nil now.

xing-yang: This nil check on line 269 can be removed now since you moved the check inside setAnnVolumeSnapshotBeingDeleted()

huffmanca: I removed the nil check here, but I also updated the debug logging to not print the content name. That's where I was hitting the panic

pkg/common-controller/snapshot_controller.go
```diff
@@ -1281,6 +1281,9 @@ func (ctrl *csiSnapshotCommonController) getSnapshotFromStore(snapshotName strin
 func (ctrl *csiSnapshotCommonController) setAnnVolumeSnapshotBeingDeleted(content *crdv1.VolumeSnapshotContent) error {
 	// Set AnnVolumeSnapshotBeingDeleted if it is not set yet
 	if !metav1.HasAnnotation(content.ObjectMeta, utils.AnnVolumeSnapshotBeingDeleted) {
+		if content == nil {
+			return nil
+		}
```
xing-yang: This check should be moved before the if block.  Otherwise we'll still get a nil exception on line 1283 as it references the content.  I'm going to revert this PR.  Can you please submit a different PR?

xing-yang: Fixed in https://github.com/kubernetes-csi/external-snapshotter/pull/279

PR #280, opened 2020-03-23
add metric to common controller - basic metrics utilities

### LINKED ISSUE #142: Add Metrics to the snapshot controller
Add Metrics to snapshot controller.  Need to be consistent with Metrics in PV/PVC controller.

@jingxu97 to check if Shawn (?) can work on this.

yuxiangqian:
**What type of PR is this?**
/kind feature

This PR is to provide basic metrics utilities which later could be used in common controller for end to end operation metrics recording.
Fixes #142 

This PR only contains the basic metrics utility functions needed for end to end operational metrics recording in snapshot controller. Another PR will be based on this to add metrics recording in snapshot controller code. ATM, it should NOT have any user-facing changes.

```release-note
Adds the basic metrics utility functions needed for end to end operational metrics recording in snapshot controller.
```

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -1,0 +1,177 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"fmt"
+	"net"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"k8s.io/apimachinery/pkg/types"
+	k8smetrics "k8s.io/component-base/metrics"
+	"k8s.io/klog"
+)
+
+const (
+	opStatusUnknown      = "Unknown"
+	labelDriverName      = "driver_name"
+	labelOperationName   = "operation_name"
+	labelOperationStatus = "operation_status"
+	subSystem            = "snapshot_controller"
+	metricName           = "operation_total_seconds"
+	metricHelpMsg        = "Total number of seconds spent by the controller on an operation from end to end"
+)
+
+// OperationStatus is the interface type for representing an operation's execution
+// status, with the nil value representing an "Unknown" status of the operation.
+type OperationStatus interface {
+	String() string
+}
+
+var metricBuckets = []float64{0.1, 0.25, 0.5, 1, 2.5, 5, 10, 15, 30, 60, 120, 300, 600}
+
+type MetricsManager interface {
+	// StartMetricsEndpoint starts the metrics endpoint at the specified addr/pattern for
+	// metrics managed by this MetricsManager. It spawns a goroutine to listen to
+	// and serve HTTP requests received on addr/pattern.
+	// If the "pattern" is empty (i.e., ""), no endpoint will be started.
+	// An error will be returned if there is any.
+	StartMetricsEndpoint(pattern, addr string, logger promhttp.Logger, wg *sync.WaitGroup) (*http.Server, error)
+
+	// OperationStart takes in an operation and caches its start time.
+	// if the operation already exists, it's an no-op.
+	OperationStart(op Operation)
+
+	// DropOperation removes an operation from cache.
+	// if the operation does not exist, it's an no-op.
+	DropOperation(op Operation)
+
+	// RecordMetrics records a metric point. Note that it will be an no-op if an
+	// operation has NOT been marked "Started" previously via invoking "OperationStart".
+	// Invoking of RecordMetrics effectively removes the cached entry.
+	// op - the operation which the metric is associated with.
+	// status - the operation status, if not specified, i.e., status == nil, an
+	//          "Unknown" status of the passed-in operation is assumed.
+	RecordMetrics(op Operation, status OperationStatus)
+}
+
+// Operation is a structure which holds information to identify a snapshot
+// related operation
+type Operation struct {
+	// the name of the operation, for example: "CreateSnapshot", "DeleteSnapshot"
+	Name string
+	// the name of the driver which executes the operation
+	Driver string
+	// the resource UID to which the operation has been executed against
+	ResourceID types.UID
+}
+
+type operationMetricsManager struct {
+	// cache is a concurrent-safe map which stores start timestamps for all
+	// ongoing operations.
+	// key is an Operation
+	// value is the timestamp of the start time of the operation
+	cache sync.Map
+
+	// registry is a wrapper around Prometheus Registry
+	registry k8smetrics.KubeRegistry
+
+	// opLatencyMetrics is a Histogram metrics
+	opLatencyMetrics *k8smetrics.HistogramVec
+}
+
+func NewMetricsManager() MetricsManager {
+	mgr := &operationMetricsManager{
+		cache: sync.Map{},
+	}
+	mgr.init()
+	return mgr
+}
+
+func (opMgr *operationMetricsManager) OperationStart(op Operation) {
+	opMgr.cache.LoadOrStore(op, time.Now())
+}
+
+func (opMgr *operationMetricsManager) DropOperation(op Operation) {
+	opMgr.cache.Delete(op)
+}
+
+func (opMgr *operationMetricsManager) RecordMetrics(op Operation, status OperationStatus) {
+	obj, exists := opMgr.cache.Load(op)
+	if !exists {
+		// the operation has not been cached, return directly
+		return
+	}
+	ts, ok := obj.(time.Time)
+	if !ok {
+		// the cached item is not a time.Time, should NEVER happen, clean and return
+		klog.Errorf("Invalid cache entry for key %v", op)
+		opMgr.cache.Delete(op)
+		return
+	}
+	strStatus := opStatusUnknown
+	if status != nil {
+		strStatus = status.String()
+	}
+	duration := time.Since(ts).Seconds()
+	opMgr.opLatencyMetrics.WithLabelValues(op.Driver, op.Name, strStatus).Observe(duration)
+	opMgr.cache.Delete(op)
+}
+
+func (opMgr *operationMetricsManager) init() {
+	opMgr.registry = k8smetrics.NewKubeRegistry()
+	opMgr.opLatencyMetrics = k8smetrics.NewHistogramVec(
+		&k8smetrics.HistogramOpts{
+			Subsystem: subSystem,
+			Name:      metricName,
+			Help:      metricHelpMsg,
+			Buckets:   metricBuckets,
+		},
+		[]string{labelDriverName, labelOperationName, labelOperationStatus},
+	)
+	opMgr.registry.MustRegister(opMgr.opLatencyMetrics)
+}
+
+func (opMgr *operationMetricsManager) StartMetricsEndpoint(pattern, addr string, logger promhttp.Logger, wg *sync.WaitGroup) (*http.Server, error) {
+	if addr == "" {
+		return nil, fmt.Errorf("metrics endpoint will not be started as endpoint address is not specified")
+	}
+	// start listening
+	l, err := net.Listen("tcp", addr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to listen on address[%s], error[%v]", addr, err)
+	}
+	mux := http.NewServeMux()
+	mux.Handle(pattern, k8smetrics.HandlerFor(
+		opMgr.registry,
+		k8smetrics.HandlerOpts{
+			ErrorLog:      logger,
+			ErrorHandling: k8smetrics.ContinueOnError,
+		}))
+	srv := &http.Server{Addr: l.Addr().String(), Handler: mux}
+	// start serving the endpoint
+	go func() {
+		defer wg.Done()
+		if err := srv.Serve(l); err != http.ErrServerClosed {
+			klog.Fatalf("failed to start endpoint at:%s/%s, error: %v", addr, pattern, err)
+		}
+	}()
+	return srv, nil
+}
--- a/pkg/metrics/metrics_test.go
+++ b/pkg/metrics/metrics_test.go
@@ -1,0 +1,556 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"reflect"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	cmg "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+var (
+	statusMap map[int]string = map[int]string{
+		0: "Success",
+		1: "Failure",
+		2: "Unknown",
+	}
+)
+
+const (
+	httpPattern = "/metrics"
+	addr        = "localhost:0"
+)
+
+type fakeOpStatus struct {
+	statusCode int
+}
+
+func (s *fakeOpStatus) String() string {
+	if str, ok := statusMap[s.statusCode]; ok {
+		return str
+	}
+	return "Unknown"
+}
+
+func initMgr() (MetricsManager, *sync.WaitGroup, *http.Server) {
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	mgr := NewMetricsManager()
+	srv, err := mgr.StartMetricsEndpoint(httpPattern, addr, nil, wg)
+	if err != nil {
+		log.Fatalf("failed to start serving [%v]", err)
+	}
+	return mgr, wg, srv
+}
+
+func shutdown(srv *http.Server, wg *sync.WaitGroup) {
+	if err := srv.Shutdown(context.Background()); err != nil {
+		panic(err)
+	}
+	wg.Wait()
+}
+
+func TestNew(t *testing.T) {
+	mgr, wg, srv := initMgr()
+	defer shutdown(srv, wg)
+	if mgr == nil {
+		t.Errorf("failed testing new")
+	}
+}
+
+func TestDropNonExistingOperation(t *testing.T) {
+	mgr, wg, srv := initMgr()
+	defer shutdown(srv, wg)
+	op := Operation{
+		Name:       "drop-non-existing-operation-should-be-noop",
+		Driver:     "driver",
+		ResourceID: types.UID("uid"),
+	}
+	mgr.DropOperation(op)
+}
+
+func TestRecordMetricsForNonExistingOperation(t *testing.T) {
+	mgr, wg, srv := initMgr()
+	srvAddr := "http://" + srv.Addr + httpPattern
+	defer shutdown(srv, wg)
+	op := Operation{
+		Name:       "no-metrics-should-be-recorded-as-operation-did-not-start",
+		Driver:     "driver",
+		ResourceID: types.UID("uid"),
+	}
+	mgr.RecordMetrics(op, nil)
+	rsp, err := http.Get(srvAddr)
+	if err != nil || rsp.StatusCode != http.StatusOK {
+		t.Errorf("failed to get response from server %v, %v", err, rsp)
+	}
+	r, err := ioutil.ReadAll(rsp.Body)
+	if err != nil {
+		t.Errorf("failed to read response body %v", err)
+	}
+	if strings.Contains(string(r), op.Name) {
+		t.Errorf("found metric should have been dropped for operation [%s] [%s]", op.Name, string(r))
+	}
+}
+
+func TestDropOperation(t *testing.T) {
+	mgr, wg, srv := initMgr()
+	srvAddr := "http://" + srv.Addr + httpPattern
+	defer shutdown(srv, wg)
+	op := Operation{
+		Name:       "should-have-been-dropped",
+		Driver:     "driver",
+		ResourceID: types.UID("uid"),
+	}
+	mgr.OperationStart(op)
+	mgr.DropOperation(op)
+	time.Sleep(300 * time.Millisecond)
+	rsp, err := http.Get(srvAddr)
+	if err != nil || rsp.StatusCode != http.StatusOK {
+		t.Errorf("failed to get response from server %v, %v", err, rsp)
+	}
+	r, err := ioutil.ReadAll(rsp.Body)
+	if err != nil {
+		t.Errorf("failed to read response body %v", err)
+	}
+	if strings.Contains(string(r), op.Name) {
+		t.Errorf("found metric should have been dropped for operation [%s] [%s]", op.Name, string(r))
+	}
+	// re-add with a different name
+	op.Name = "should-have-been-added"
+	mgr.OperationStart(op)
+	time.Sleep(300 * time.Millisecond)
+	opStatus := &fakeOpStatus{
+		statusCode: 0,
+	}
+	mgr.RecordMetrics(op, opStatus)
+	expected :=
+		`# HELP snapshot_controller_operation_total_seconds [ALPHA] Total number of seconds spent by the controller on an operation from end to end
+# TYPE snapshot_controller_operation_total_seconds histogram
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="0.1"} 0
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="0.25"} 0
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="0.5"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="1"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="2.5"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="5"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="10"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="15"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="30"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="60"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="120"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="300"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="600"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="+Inf"} 1
+snapshot_controller_operation_total_seconds_sum{driver_name="driver",operation_name="should-have-been-added",operation_status="Success"} 0.3
+snapshot_controller_operation_total_seconds_count{driver_name="driver",operation_name="should-have-been-added",operation_status="Success"} 1
+`
+
+	if err := verifyMetric(expected, srvAddr); err != nil {
+		t.Errorf("failed testing[%v]", err)
+	}
+}
+
+func TestUnknownStatus(t *testing.T) {
+	mgr, wg, srv := initMgr()
+	srvAddr := "http://" + srv.Addr + httpPattern
+	defer shutdown(srv, wg)
+	op := Operation{
+		Name:       "unknown-status-operation",
+		Driver:     "driver",
+		ResourceID: types.UID("uid"),
+	}
+	mgr.OperationStart(op)
+	// should create a Unknown data point with latency ~300ms
+	time.Sleep(300 * time.Millisecond)
+	mgr.RecordMetrics(op, nil)
+	expected :=
+		`# HELP snapshot_controller_operation_total_seconds [ALPHA] Total number of seconds spent by the controller on an operation from end to end
+# TYPE snapshot_controller_operation_total_seconds histogram
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="0.1"} 0
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="0.25"} 0
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="0.5"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="1"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="2.5"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="5"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="10"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="15"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="30"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="60"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="120"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="300"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="600"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown",le="+Inf"} 1
+snapshot_controller_operation_total_seconds_sum{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown"} 0.3
+snapshot_controller_operation_total_seconds_count{driver_name="driver",operation_name="unknown-status-operation",operation_status="Unknown"} 1
+`
+
+	if err := verifyMetric(expected, srvAddr); err != nil {
+		t.Errorf("failed testing[%v]", err)
+	}
+}
+
+func TestRecordMetrics(t *testing.T) {
+	mgr, wg, srv := initMgr()
+	srvAddr := "http://" + srv.Addr + httpPattern
+	defer shutdown(srv, wg)
+	// add an operation
+	op := Operation{
+		Name:       "op1",
+		Driver:     "driver1",
+		ResourceID: types.UID("uid1"),
+	}
+	mgr.OperationStart(op)
+	// should create a Success data point with latency ~ 1100ms
+	time.Sleep(1100 * time.Millisecond)
+	success := &fakeOpStatus{
+		statusCode: 0,
+	}
+	mgr.RecordMetrics(op, success)
+
+	// add another operation metric
+	op.Name = "op2"
+	op.Driver = "driver2"
+	op.ResourceID = types.UID("uid2")
+	mgr.OperationStart(op)
+	// should create a Failure data point with latency ~ 100ms
+	time.Sleep(100 * time.Millisecond)
+	failure := &fakeOpStatus{
+		statusCode: 1,
+	}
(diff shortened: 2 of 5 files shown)


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,190 @@
...
+
+const (
+	// InterimFailure is an operation state which means the controller will keep
+	// on retrying the operation upon encountering this type of error.
+	InterimFailure OperationState = "InterimFailure"
+
+	// TerminatingFailure is an operation state which means the controller has encountered
+	// an error which it's not able to resume the operation and has marked the
+	// permanently failure of the operation.
+	TerminatingFailure OperationState = "TerminatingFailure"
```
xing-yang: How do you differentiate a TerminatingFailure vs a InterimFailure?

yuxiangqian: in controller, some errors, for example controller failure or timeout when creating snapshot, are considered to be interim and controller will reconcile later to retry. Basically, if a controller encountered an error and will try to reconcile later should be considered as interim failure, otherwise should be terminating error. WDYT?

xing-yang: Sure.  We added a isFinalError() in the CreateSnapshot timeout PR, but that depends on gRPC error codes.  We can probably do something similar for the snapshot controller.

https://github.com/kubernetes-csi/external-snapshotter/pull/261/files#diff-de5dbd65778c167a3d05cd17929d6851R658-R681

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,190 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
msau42: 2020

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,190 @@
...
+	// InterimFailure is an operation state which means the controller will keep
+	// on retrying the operation upon encountering this type of error.
+	InterimFailure OperationState = "InterimFailure"
+
+	// TerminatingFailure is an operation state which means the controller has encountered
+	// an error which it's not able to resume the operation and has marked the
+	// permanently failure of the operation.
+	TerminatingFailure OperationState = "TerminatingFailure"
+
+	// Success states that the controller has successfully executed the operation.
```
msau42: Is this time to cut, or time to upload?

pkg/metrics/metrics.go
msau42: What is an example of a terminating failure?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// Operation is a structure which holds information to identifier a snapshot
+// related operation
+type Operation struct {
+	// the name of the operation, for example: "CreateSnapshot", "DeleteSnapshot"
+	Name string
+	// the name of the driver which executes the operation
+	Driver string
+	// the resource UID to which the operation has been executed against
```
msau42: In this case, it's the VolumeSnapshot UID?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,190 @@
...
+		opMgr.cache.Delete(key)
+		return
+	}
+	duration := time.Since(ts.startTime).Seconds()
+	opMgr.opLatencyMetric.WithLabelValues(op.Driver, op.Name, string(state)).Observe(duration)
+	switch state {
+	case Success, TerminatingFailure:
+		opMgr.cache.Delete(key)
+	case InterimFailure:
+		// do nothing
```
msau42: I don't think we should count retry loops in the same metric for total latency across retry loops. It's going to make the error rate calculations strange because the total count is number of retry loops instead of number of objects processed.

pkg/metrics/metrics.go
yuxiangqian: done

pkg/metrics/metrics.go
yuxiangqian: it's up to the caller to decide. For volume snapshot, it should be when a content has been marked by ready to use = uploading finished.

pkg/metrics/metrics.go
yuxiangqian: for this specific case yes.

pkg/metrics/metrics.go
msau42: Do you think we should have 2 different metrics? One for time to cut and one for time to upload?

pkg/metrics/metrics.go
yuxiangqian: for example when an operation timed out from CSI, controller failed to update status field of an object due to API error etc.

pkg/metrics/metrics.go
yuxiangqian: I am not sure about that. time to upload applies to only cloud providers. For those storage systems doing only inplace snapshots, it really does not apply. WDYT?

pkg/metrics/metrics.go
yuxiangqian: If error rate is the only concern, I think it's a solvable problem, it depends on how you define error rates.
if error rate is defined by total permanent failure / total requests, it can be calculated by get the sum with "TerminatingFailure" over the total number of requests.
if error rate is defined by total failures / total # operations, it can be calculated by sum(terminating failure + interim failure) / total number of operations.
count retrying loops have its value to capture interim status and does make sense to me. WDYT?

pkg/metrics/metrics.go
yuxiangqian: For example, CSI returned a failure which is not recoverable from the controller automatically.

msau42: Are there any error codes returned by the CSI driver that the controller will not retry?How does "user canceled" get counted?

pkg/metrics/metrics.go
msau42: We specifically call out creation vs readyToUse in the API. It would be good to be explicit what this metric is measuring.

pkg/metrics/metrics.go
msau42: How do you calculate # total requests vs # total operations? Currently this metric is counting # total operations.

I agree counting retry loops and counting # user requests are both valuable. I think it is confusing and difficult to count them both in one metric.

pkg/metrics/metrics.go
xing-yang: Snapshot controller does not directly get CSI driver error codes.  For create snapshot, maybe we can check the VolumeSnapshotBeingCreated annotation on the content?  If create snapshot failed and content does not have the annotation, that means it is a terminating failure.

xing-yang:
In the release note, can you rename "common controller" to "snapshot controller".

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,190 @@
...
+type OperationState string
+
+const (
+	// InterimFailure is an operation state which means the controller will keep
+	// on retrying the operation upon encountering this type of error.
+	InterimFailure OperationState = "InterimFailure"
+
+	// TerminatingFailure is an operation state which means the controller has encountered
+	// an error which it's not able to resume the operation and has marked the
+	// permanently failure of the operation.
```
xing-yang: s/permanently/permanent

pkg/metrics/metrics.go
msau42: But I thought we retry on any error until the object is deleted?

pkg/metrics/metrics.go
xing-yang: Seems good to show both.  We could decide if the snapshot is cut by checking if the creation timestamp is set, and use readyToUse to decide if it is ready to use.  I'd avoid using "upload" in the general metrics term as not every driver supports upload so it could cause confusion.

pkg/metrics/metrics.go
xing-yang: Yes.  I just wonder if we need to change the logic now that we are talking about terminating failure.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,190 @@
...
+	// op - the Operation which the metric is associated with.
+	// state - metrics manager operations differently to the cache based on the state
+	//         of the Operation:
+	//   InterimFailure: record a failure metric however keep operation's start time in cache
+	//   TerminatingFailure: record a failure metric and remove operation's start time from cache
+	//   Success: record a success metric and remove the operation's start time from cache
+	RecordMetrics(op Operation, state OperationState)
+}
+
+// Operation is a structure which holds information to identifier a snapshot
```
xing-yang: s/identifier/identify

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,190 @@
...
+
+type operationTs struct {
+	operation Operation
+	startTime time.Time
+}
+
+type operationMetricsManager struct {
+	// cache is a concurrent-safe map which stores start timestamps for all
+	// ongoing operations.
+	// key is
```
xing-yang: comments not complete?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,190 @@
...
+	// value is
```
xing-yang: Same here

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,190 @@
...
+	cache sync.Map
+
+	registry k8smetrics.KubeRegistry
```
xing-yang: Comments on what this is?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,190 @@
...
+
+	opLatencyMetric *k8smetrics.HistogramVec
```
xing-yang: Comments

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,202 @@
...
+		Driver:     "driver1",
+		ResourceID: types.UID("uid1"),
+	}
+	mgr.OperationStart(op)
+	// should create an InterimFailure data point with latency ~ 300ms
+	time.Sleep(300 * time.Millisecond)
+	mgr.RecordMetrics(op, InterimFailure)
+	// should create an InterimFailure data point with latency ~ 1100ms
+	time.Sleep(800 * time.Millisecond)
+	mgr.RecordMetrics(op, Success)
```
xing-yang: Comment says "InterimFailure" but it records a "Success".

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+	// add another operation metric
+	op.Name = "op2"
+	op.Driver = "driver2"
+	op.ResourceID = types.UID("uid2")
+	mgr.OperationStart(op)
+	// should create an InterimFailure data point with latency ~ 100ms
+	time.Sleep(100 * time.Millisecond)
+	mgr.RecordMetrics(op, TerminatingFailure)
```
xing-yang: Comment says "InterimFailure" but it records a "TerminatingFailure".

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,202 @@
...
+		// compare metric
+		gh := got.Metric[i].Histogram
+		eh := expected.Metric[i].Histogram
+		if gh == nil {
+			return false
+		}
+		if gh.SampleSum == nil || math.Abs(*(gh.SampleSum)-*(eh.SampleSum)) > maxDelta {
+			return false
+		}
+		if gh.SampleCount == nil || *(gh.SampleCount) != *(eh.SampleCount) {
```
xing-yang: do we need a nil check for eh.SampleCount

pkg/metrics/metrics.go
yuxiangqian: My take on that would be sidecar controller *could* update the content object with a permanent error in the status on those cases instead of utilizing annotation.

pkg/metrics/metrics.go
yuxiangqian: I buy the arguments of separating cut and uploading latency, my concern is uploading is specific to limited providers and does not really make much sense to report for the rest(a snapshot will be marked as read to use right after a success create snapshot call has been made). Thinking about how snapshot controller would try to report metrics, it reacts to update of a VolumeSnapshotContent status change, and update its corresponding VolumeSnapshot's status, and then reports a latency metric. For those providers who do not need an upload phase, the ReadyToUse flag will be true, and a metric should be reported but which one? do we report both cut/upload?

pkg/metrics/metrics.go
yuxiangqian: total # of operation is going to be "snapshot_controller_operation_total_seconds_count"
total # of request is going to be(approximately) "snapshot_controller_operation_total_seconds_count" however with labels == success || labels == Permanent failure filtering?

yuxiangqian:
> In the release note, can you rename "common controller" to "snapshot controller".

done. thnx

pkg/metrics/metrics_test.go
yuxiangqian: no I ignored nil check purposefully as the expected output should be parsed successfully with those information available.

pkg/metrics/metrics_test.go
yuxiangqian: done

pkg/metrics/metrics_test.go
yuxiangqian: done

pkg/metrics/metrics.go
yuxiangqian: or maybe we say we only record ready latency metrics for those providers which do not have an upload phase.

pkg/metrics/metrics.go
yuxiangqian: done

pkg/metrics/metrics.go
yuxiangqian: done

pkg/metrics/metrics.go
yuxiangqian: done

pkg/metrics/metrics.go
yuxiangqian: done

pkg/metrics/metrics.go
yuxiangqian: done

pkg/metrics/metrics.go
xing-yang: How do you determine this?  By checking if both creation timestamp and readyToUse are returned at the same time?

yuxiangqian: could be something as following:
if timestamp != nil {
   if readytouse {
       report readytouse metric
       return
   }
   report cut metric
}

msau42: My suggestion is to have 2 operations, one for cut and one for upload. For the providers that don't upload, then their upload latency will be 0 and always successful. That seems fine to me.

pkg/metrics/metrics.go
msau42: I'm confused. I thought we always retry no matter the error. So the only time we stop retrying is when the user deletes the VolumeSnapshot object.

pkg/metrics/metrics.go
msau42: The start time doesn't make sense for the single-retry loop case. Say the operation starts at time X. It fails 5 times, and then succeeds on the 6th. Say every loop takes 1 second.

The latency of the temporary failure is going to be recorded as X+1, X+2, X+3, X+4, X+5, instead of 1, 1, 1, 1, 1.

I really think it's easier to have two different metrics: one for e2e time, and one for a single reconciliation loop.

yuxiangqian: that's a very fair point.
I thought about that but IMO it might not necessary a problem as latency metric for an interim reconcile loop(failed one) is really not that important. When error happens, count is more important than how much time it has spent? WDYT?
if you feel strong about latency of a failed reconcile loop is valuable, another timestamp can be introduced to record the last visited time in the cache as well to have a relatively more accurate metrics. I am hesitate to introduce two similar metrics.

pkg/metrics/metrics.go
yuxiangqian: yeah, that's the approach I am looking to implement in controller

pkg/metrics/metrics.go
msau42: I think latency of a single loop is valuable to show if something is hanging/slow that causes back pressure on the controller. I'm not sure I understand the confusion around having two metrics, one to capture total e2e time, and one to capture a single reconciliation loop time. In fact, I think the workqueue struct already has generic metrics that we may be able to reuse.

pkg/metrics/metrics.go
xing-yang: Sounds good to me.

pkg/metrics/metrics.go
yuxiangqian: workqueue does have generic metrics support at [here](https://github.com/kubernetes/client-go/blob/master/util/workqueue/metrics.go), however it's not trivial to distinguish different operations and  add labels dynamically based on the type of operation is not easy. A customized [MetricsProvider](https://github.com/kubernetes/client-go/blob/master/util/workqueue/metrics.go#L173) is needed.
When I say two metrics, I meant two histograms, we can record use the same histogram as e2e time with some enhancement to the cache.

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,217 @@
...
+var (
+	mgr MetricsManager
+	srv *http.Server
+	wg  *sync.WaitGroup
+)
+
+const (
+	httpPattern = "/metrics"
+	addr        = ":12134"
+	httpAddr    = "http://localhost" + addr + httpPattern
```
xing-yang: This failed on my setup too:

```
--- FAIL: TestStartServing (0.00s)
    metrics_test.go:77: failed to get response from server Get http://localhost:12134/metrics: dial tcp 127.0.0.1:12134: connect: connection refused
```

xing-yang: Hmm, seems very flaky.  I ran the test a few times and finally it passed.

yuxiangqian: yes, its failing constantly, I will need to fix it.
could you share with me your set up? do you have a full travis setup?
I ran:
make -k all test
it worked fine.

xing-yang: This is just on my dev machine.  I go to the metrics directory and ran "go test".

xing-yang: Tried "make -k all test" but it gave me the same error.

yuxiangqian: this is now fixed. @oleksiys helped figured it out. The root cause was the call "ListenAndServe" does two major things: Listen on the provided address, and start serving, and it's done in a separate goroutine. When the test code tries to call "Get", the server might not necessarily have been initialized yet and causes connection refused error.

pkg/metrics/metrics.go
xing-yang: Yes.  Controller will always retry to create the snapshot no matter what kind of error it is.  Now I think we don't need to track the interim failure such as timeout.  We only need to track a terminating failure, i.e., driver finally finished the create snapshot call with a final error which is not a timeout.  Controller will still retry later until the user deletes the VolumeSnapshot object.  If the controller can never successfully create the snapshot, then the failure count will continue to increment with each retry.

xing-yang:
Ran unit test on my setup and it passed now.

pkg/metrics/metrics.go
msau42: Sorry I still have concerns that that combining the two different measurements into one histogram for the following reasons:
1) Total latency, and per-loop latency are intended for two different audiences. One is for end users, another is for developers. Combining the two measurements into one I think is confusing to explain to end users how to interpret. It requires that the end users understand how things are implemented.
2) I'm not convinced we're able to capture all the conditions we want. For example, can we tell the difference between success latency of a retry loop vs success latency of a total operation? We can potentially keep on adding more states for each of these corner cases, but I think that makes 1) worse.
3) Adding more special cases to the cache makes the cache more complicated, more specialized, and less generalizable to future components.

I still prefer that we limit this to only capture end-to-end latency for now and consider how to capture a per-loop metric separately.

yuxiangqian: removed per operation latency from the current implementation and only left end to end + error count.

yuxiangqian:
@xing-yang @msau42 ping, ready for review

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,180 @@
...
+
+	// Success states that the controller has successfully executed the operation.
+	Success OperationState = "Success"
+)
+
+var metricBuckets = []float64{0.1, 0.25, 0.5, 1, 2.5, 5, 10, 15, 30, 60, 120, 300, 600}
+
+type MetricsManager interface {
+	// StartServing starts the metrics endpoint at the specified addr/pattern for
+	// metrics managed by this MetricsManager. It spawn a goroutine to listen to
```
xing-yang, line 57: s/spawn/spawns

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,180 @@
...
+	// and serve HTTP requests received on addr/pattern
+	// if the "pattern" is empty (i.e., ""), no endpoint will be started. An error
+	// will be returned if there is any.
+	StartServing(pattern, addr string, logger promhttp.Logger, wg *sync.WaitGroup) (*http.Server, error)
+
+	// OperationStart takes in an operation and cache it's start time.
```
xing-yang, line 63: s/cache/caches
s/it's/its

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,180 @@
...
+func (opMgr *operationMetricsManager) OperationStart(op Operation) {
+	opMgr.cache.LoadOrStore(op, time.Now())
+}
+
+func (opMgr *operationMetricsManager) DropOperation(op Operation) {
+	opMgr.cache.Delete(op)
+}
+
+func (opMgr *operationMetricsManager) RecordMetrics(op Operation, state OperationState) {
+	//	key := op.key()
```
xing-yang, line 123: Remove commented out code

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	if strings.Contains(string(r), op.Name) {
+		t.Errorf("found metric should have been dropped for operation [%s] [%s]", op.Name, string(r))
+	}
+	// re-add
+	mgr.OperationStart(op)
+	time.Sleep(300 * time.Millisecond)
+	mgr.RecordMetrics(op, Success)
+	expected :=
+		`# HELP snapshot_controller_operation_total_seconds [ALPHA] Total number of seconds spent by the controller on an operation from end to end
```
xing-yang, line 110: What does "ALPHA" mean?

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,264 @@
...
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver2",operation_name="op2",operation_status="Failure",le="60"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver2",operation_name="op2",operation_status="Failure",le="120"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver2",operation_name="op2",operation_status="Failure",le="300"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver2",operation_name="op2",operation_status="Failure",le="600"} 1
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver2",operation_name="op2",operation_status="Failure",le="+Inf"} 1
+snapshot_controller_operation_total_seconds_sum{driver_name="driver2",operation_name="op2",operation_status="Failure"} 0.100174856
+snapshot_controller_operation_total_seconds_count{driver_name="driver2",operation_name="op2",operation_status="Failure"} 1
+`
+	if err := verifyMetric(expected); err != nil {
+		t.Errorf("failed testing[%v]", err)
```
xing-yang, line 193: space after "testing"

xing-yang:
In the release notes, can you just include the following as this will be included in the release notes in the future:
```
Adds the basic metrics utility functions needed for end to end operational metrics recording in snapshot controller.
``` 

And move the remaining message to the Special notes for the reviewer section:
Another PR will be based on this to add metrics recording in snapshot controller code. ATM, it should NOT have any user-facing changes.

pkg/metrics/metrics_test.go
yuxiangqian: This is a fixed format from k8s.io/component-base/metrics
when initializing a metrics (in this case a histogram), this flag shows the stability level of this metrics, and by default its ALPHA.
see [here](https://github.com/kubernetes/kubernetes/blob/d56aaf77b929f70e740aab9c965a9b932187d036/staging/src/k8s.io/component-base/metrics/opts.go#L57)

yuxiangqian:
@xing-yang comments addressed.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,179 @@
...
+	metricHelpMsg        = "Total number of seconds spent by the controller on an operation from end to end"
+)
+
+type OperationState string
+
+const (
+	// Failed is an operation state which means the controller has encountered
+	// an error which is not recoverable and the controller has marked a permanent
+	// failure of the operation.
+	Failure OperationState = "Failure"
```
msau42, line 47: In what situations does the controller not retry?

Do we also count user cancellation as a failure?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,179 @@
...
+	DropOperation(op Operation)
+
+	// RecordMetrics records a metric point. Note that it will be an no-op if an
+	// operation has not been marked "Started" previously via invoking "OperationStart".
+	// Invoking of RecordMetrics effectively removes the cached entry.
+	// op - the operation which the metric is associated with.
+	// state - the operation state
+	//   Failure: record a failure metric
+	//   Success: record a success metric
+	RecordMetrics(op Operation, state OperationState)
```
msau42, line 78: For failure, I think it would be good to support status_codes, so that we can distinguish between configuration errors, and system errors.

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+}
+
+func TestStartServing(t *testing.T) {
+	if _, err := http.Get(httpAddr); err != nil {
+		t.Errorf("failed to get response from server %v", err)
+	}
+}
+
+func TestDropOperation(t *testing.T) {
```
msau42, line 86: Can we also have some test cases like multiple operations at the same time for the same driver, in various states, some dropped, some failed, some success, some ongoing?

pkg/metrics/metrics.go
yuxiangqian: good point but that would require major changes in the controller code to be able to generate those return codes.

pkg/metrics/metrics.go
yuxiangqian: ATM when CSI failed permanently it would stop retrying to call CSI.
when you say user cancellation, are you referring to deletion of a VS?

msau42: What is failed permanently? I don't think we should stop retrying until user cancels by deleting the VS. It is a Kubernetes anti-pattern

pkg/metrics/metrics.go
msau42: Can we make the metrics future proof by having a status_code dimension instead? We did this for in-tree too, with the expectation that in the future we may add more status codes. See https://github.com/kubernetes/kubernetes/pull/75750/files

pkg/metrics/metrics.go
xing-yang: This failure refers to the failure of one try.  The controller should always re-try.  We do have a bug to be fixed - If create snapshot fails very early, before the request is sent to the CSI driver, there will be no retries.  But we do plan to fix it when fixing the requeue logic.

yuxiangqian: one concern is that the cache will grow linearly to the number of failed volumesnapshots over time, and when should we report an error metrics if there is no permanently failed VolumeSnapshot?

msau42: the only time this e2e metric should be recorded is:
* volume operation is successful
* volume operation was aborted

xing-yang: "volume operation was aborted" - create snapshot will only be aborted when user deletes the snapshot, but how is that a failure?

msau42: A couple scenarios I see are:
* There was an invalid configuration so user canceled the operation, fixed the config, and then retry
* There was a system error and the user gave up and deleted the object
* Snapshot was taking too long and the user got impatient and gave up.

pkg/metrics/metrics.go
yuxiangqian: the state code is an alias to string which itself is already future proof? are you asking to make the function to take in string as "state" type?

pkg/metrics/metrics.go
yuxiangqian: If reconcile will retry CSI calls regardless the CSI return code (which is not the case ATM), I feel having an error metrics for end to end operation is not necessary and maybe not that valuable. User cancellation/abort is a pretty random behavior, they could cancel anytime after a failure has been observed, which will in turn make the metrics latency value not that useful. On the other hand, the error count, could be potentially valuable, as it would give users an idea how many total cancellations have been observed.

msau42: The latency value could tell us information about how long users have waited before giving up. I think since count is included as part of the histogram anyway, it doesn't hurt to include the latency. Consumers can decide if they want the use the data or not.

msau42: (Also we should fix the controller to retry in all cases. This was one of the design changes we made for beta)

pkg/metrics/metrics.go
msau42: I was thinking if in the future we decide to support error codes, then having the input be an error code would be more flexible. Especially if we want to take this metrics library and reuse it for other controllers (ie put it in csi-lib-utils).

But if we only want to contain this for snapshot controller for now, then at least I think we should be more descriptive in the current failure string to leave room for future error types.

yuxiangqian: introduced a status interface to make it more extendable.

yuxiangqian:
@msau42 gental ping

xing-yang:
lgtm

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,520 @@
...
+
+func shutdown() {
+	if err := srv.Shutdown(context.Background()); err != nil {
+		panic(err)
+	}
+	wg.Wait()
+}
+
+func TestMain(m *testing.M) {
+	initMgr()
```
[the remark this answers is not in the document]
yuxiangqian: This is on purposely set up to avoid potential "port in use" errors. After a HTTP server has been shutdown, there is a short period of time for the OS to mark the port available, trying to bind to that port would cause failure.
"That will cause problems & flaky if there is any state that changes across test." could you be a bit more specific?

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,520 @@
...
+		t.Errorf("failed testing new")
+	}
+}
+
+func TestStartServing(t *testing.T) {
+	if _, err := http.Get(httpAddr); err != nil {
+		t.Errorf("failed to get response from server %v", err)
+	}
+}
+
```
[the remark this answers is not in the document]
yuxiangqian: adding one more to cover " recording when no operations have been started", the rest of the cases should have. already been covered in TestConcurrency

yuxiangqian:
> Hi Xiangqian,
> 
> Michelle asked me to help out with this review. I've added a few comments.
> 
> Thanks!
> 
> Matt

Thanks Matt for taking time on this.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,177 @@
...
+
+// OperationStatus is the interface type for representing an operation's execution
+// status, with the nil value representing an "Unknown" status of the operation.
+type OperationStatus interface {
+	String() string
+}
+
+var metricBuckets = []float64{0.1, 0.25, 0.5, 1, 2.5, 5, 10, 15, 30, 60, 120, 300, 600}
+
+type MetricsManager interface {
```
humblec, line 50: Do we need to rename this to SnapshotMetricsManager ?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,177 @@
...
+	// the resource UID to which the operation has been executed against
+	ResourceID types.UID
+}
+
+type operationMetricsManager struct {
+	// cache is a concurrent-safe map which stores start timestamps for all
+	// ongoing operations.
+	// key is an Operation
+	// value is the timestamp of the start time of the operation
+	cache sync.Map
```
humblec, line 91: @yuxiangqian just to confirm: this cache gets invalidated or dropped (completely) when the controller starts/reschduled, Isn't it?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,177 @@
...
+	)
+	opMgr.registry.MustRegister(opMgr.opLatencyMetrics)
+}
+
+func (opMgr *operationMetricsManager) StartServing(pattern, addr string, logger promhttp.Logger, wg *sync.WaitGroup) (*http.Server, error) {
+	if addr == "" {
+		return nil, fmt.Errorf("metrics endpoint will not be started as endpoint address is not specified")
+	}
+	// start listening
+	l, err := net.Listen("tcp", addr)
```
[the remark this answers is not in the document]
mattcary: Both good points! Thanks for the education.

pkg/metrics/metrics_test.go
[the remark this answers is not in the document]
mattcary: Global state across tests means that one test can change things so that another test fails. But that failing test will succeed when run in isolation, making debugging very hard. That's the main reason in my experience.

The second reason is that it makes it impossible to parallelize the tests, although that's more of a theoretical issue now.

To avoid the port-in-use errors pass a 0 port, which will cause a port to be auto-allocated. The address can be recovered from svr.Addr. 

Doing this is a good idea anyway in case there are other tests or things running on the system that are allocating a port; this should be changed in csi-lib-utils as well which is yet another argument for your comment above :)

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,520 @@
...
+snapshot_controller_operation_total_seconds_bucket{driver_name="driver5",operation_name="unknown",operation_status="Unknown",le="+Inf"} 1
+snapshot_controller_operation_total_seconds_sum{driver_name="driver5",operation_name="unknown",operation_status="Unknown"} 0.100328549
+snapshot_controller_operation_total_seconds_count{driver_name="driver5",operation_name="unknown",operation_status="Unknown"} 1
+`
+	if err := verifyMetric(expected); err != nil {
+		t.Errorf("failed testing [%v]", err)
+	}
+}
+
+func verifyMetric(expected string) error {
```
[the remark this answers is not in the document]
mattcary: I think exposing the registry is a small price to pay for not having to duplicate ~100 lines of code that will have to be maintained over any changes in prometheus.

pkg/metrics/metrics_test.go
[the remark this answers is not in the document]
mattcary: It would be nice to have simpler test cases for those simple corner cases rather than having to wade through a complicated TestConcurrency failure.

I'm thinking of the perspective from the future when some change elsewhere breaks these tests, and we no longer remember how any of this works :)

pkg/metrics/metrics.go
mattcary: eh, this is already scoped to the external-snapshotter repo, anyone important this would stutter the type, wouldn't they?

pkg/metrics/metrics.go
mattcary: This is a good question.

If the controller is restarted, any in-flight operations won't be recorded if somehow it is seen by whoever calls RecordMetric. Only if OperationStart is called will RecordMetric do anything.

So this prevents prometheus from getting polluted with operations whose start time is wrong, which is a good thing.

I don't think the corner cases of operations dropped when the controller restarts is a huge deal. I think the most we could do is add a counter for RecordMetrics being called on unknown operations, but that might just be unnecessary complication.

pkg/metrics/metrics.go
yuxiangqian: yeah, I agree with Matt. Also, this might get moved to csi-util later.

pkg/metrics/metrics.go
yuxiangqian: cache is in-memory, and you are right, if the controller restart/rescheduled, metrics already reported by calling "RecordMetrics" will be in prom, however for those still in cache, i.e., had called "OperationStart" but not "RecordMetrics" yet (the operation has started but have not finished), will be lost. In the later case, recording a metric for the operation does not make much sense.

pkg/metrics/metrics_test.go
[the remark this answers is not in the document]
yuxiangqian: My apologies I have not made myself clear.
While I complete agree with your points on global state could end up cause, but it's a bit special here for http Servers without customized Handler. I have made the code change to adapt the comments.
If http.Handle(....) is used to register request handlers and no handler is supplied to create a [Server](https://golang.org/pkg/net/http/?m=all#Server), the DefaultServeMux will be used which itself is global.

pkg/metrics/metrics_test.go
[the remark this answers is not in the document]
yuxiangqian: IMO, exposing an unnecessary public interface for the sake of testing seems to be worse than a 100 lines more testing code :). I think its better to keep it hidden unless is necessary, and up till that point, we could switch to use the testutil.
Also, the test workflow is slightly different here, unit tests here are tested around http endpoint rather than prom registry.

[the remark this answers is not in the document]
mattcary: Sorry, as a likely maintainer for this code I'll have to insist. Duplicating complicated test verification is going to be a burden to maintain in the future. The public interface isn't a big deal, especially if this is going to be merged eventually with csi-lib-util.

I think the difference in the workflow doesn't prevent you from using the common library?

pkg/metrics/metrics.go
mattcary: +1

pkg/metrics/metrics_test.go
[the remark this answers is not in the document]
yuxiangqian: added simple tests as suggested

pkg/metrics/metrics_test.go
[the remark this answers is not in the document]
yuxiangqian: chatted offline to resolve this, unlike csi-lib-util/metrics, unit tests can be deterministic as the latency will be passed in when recording a data point, the encapsulation of cache manages the latency calculation. It makes deterministic latency not easy in this case and thus not quite feasible to use GatherAndCompare.
however to @mattcary 's point, I have relaxed the latency checking from a fixed range to a greater than operation to avoid flake test resulted by latencies introduced from other component, i.e., prom lib.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,177 @@
...
+	metricHelpMsg        = "Total number of seconds spent by the controller on an operation from end to end"
+)
+
+// OperationStatus is the interface type for representing an operation's execution
+// status, with the nil value representing an "Unknown" status of the operation.
+type OperationStatus interface {
+	String() string
+}
+
+var metricBuckets = []float64{0.1, 0.25, 0.5, 1, 2.5, 5, 10, 15, 30, 60, 120, 300, 600}
```
msau42, line 48: Do we think 10 mins is enough? Are we capturing upload time too?

msau42:
LGTM! just one minor question about the max histogram bucket

pkg/metrics/metrics.go
yuxiangqian: Good question. From my observation using gce PD snapshot as example, a 4G PV roughly takes ~5-10s to do a snapshot (including uploading), assuming linear interpolation, 600s *should* be able to upload any snapshot for volumes of size between 240G to 480G. do you think its sufficient? GCE PD can have up to 64TB disks, I am a little concerned to have a such big bucket which seems to be a bit overkill for most of the cases. and the exact latency time is anyway captured by the sum metric. 
Snapshot latency, depending on storage vendor implementation, varies a lot. For storage vendors with inplace implementation, the latency should be really small like within 1 second, where as for those do uploading, it takes much longer and depends on a volume's size. WDYT?

pkg/metrics/metrics.go
humblec: Sure.. nw :)

humblec:
LGTM from me too. Thanks @yuxiangqian for this !

pkg/metrics/metrics.go
humblec: @msau42 @yuxiangqian  I believe we can't control this timeout from our end, considering the fact that, all the CSI drivers can have their own `timeout` configured with an arg to the sidecar, Isn't it? Either we have to use that timeout value or stick to a default value in general like 10 mins. Am I missing something here?

msau42: This isn't about a timeout value, this is about the maximum bucket range that we want to have for our histograms. This is probably fine for now, we can add more buckets if needed.

PR #282, opened 2020-03-30
Makes the number of worker threads configurable.

### LINKED ISSUE #276: Number of worker threads should be configurable
Number of worker threads are fixed for snapshot-controller and csi-snapshotter.  They should be configurable. 

https://github.com/kubernetes-csi/external-snapshotter/blob/master/cmd/snapshot-controller/main.go#L44
https://github.com/kubernetes-csi/external-snapshotter/blob/master/cmd/csi-snapshotter/main.go#L52

Example: https://github.com/kubernetes-csi/external-attacher/blob/master/cmd/csi-attacher/main.go#L59

huffmanca:
**What type of PR is this?**
/kind feature

Adjusts the worker threads to be configurable for the snapshot-controller and csi-snapshotter.

Fixes #276 

```release-note
The number of worker threads in the snapshot-controller and csi-snapshotter is now configurable via the `worker-threads` flag.
```

### DIFF
--- a/cmd/csi-snapshotter/main.go
+++ b/cmd/csi-snapshotter/main.go
@@ -48,9 +48,6 @@ import (
 )
 
 const (
-	// Number of worker threads
-	threads = 10
-
 	// Default timeout of short CSI calls like GetPluginInfo
 	defaultCSITimeout = time.Minute
 )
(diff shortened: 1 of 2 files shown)


cmd/csi-snapshotter/main.go
```diff
@@ -63,6 +60,7 @@ var (
 	snapshotNamePrefix     = flag.String("snapshot-name-prefix", "snapshot", "Prefix to apply to the name of a created snapshot")
 	snapshotNameUUIDLength = flag.Int("snapshot-name-uuid-length", -1, "Length in characters for the generated uuid of a created snapshot. Defaults behavior is to NOT truncate.")
 	showVersion            = flag.Bool("version", false, "Show version.")
+	threads                = flag.Int("threads", 10, "Number of worker threads.")
```
msau42: Can we use the same flag name as the other sidecars?

huffmanca: I've updated this to be `worker-threads` to match the external-attacher.

msau42:
/lgtm
also update the release note thanks!

huffmanca:
Done. Thank you for the release note reminder!

PR #283, opened 2020-04-01
Check if PVC is being used by a snapshot as source

xing-yang:
**What type of PR is this?**
/kind bug

This PR fixes a problem when the PVC finalizer is not removed when it is no longer used by a snapshot as source.

Fixes #

To reproduce this problem, create snapshot, immediately delete the snapshot before create snapshot has completed.  Snapshot will be created and then deleted, but the source PVC can't be deleted because the PVC finalizer is not removed.

This PR fixes the problem.

```release-note
 Fixes a problem of not removing the PVC finalizer when it is no longer used by a snapshot as source.
```


xing-yang:
@yuxiangqian This PR is ready for review now.

pkg/common-controller/snapshot_controller.go
```diff
@@ -190,6 +193,21 @@ func (ctrl *csiSnapshotCommonController) syncSnapshot(snapshot *crdv1.VolumeSnap
 	return nil
 }
 
+// Check if PVC is being used by the current snapshot as source
+func (ctrl *csiSnapshotCommonController) isPVCInUseByCurrentSnapshot(snapshot *crdv1.VolumeSnapshot) bool {
+	// Get snapshot source which is a PVC
+	pvc, err := ctrl.getClaimFromVolumeSnapshot(snapshot)
```
yuxiangqian, line 199: IMO, it might worth checking whether the PVC has the finalizer here as well. It does not make sense to block the deletion if there is no finalizer on PVC?

pkg/common-controller/snapshot_controller.go
```diff
@@ -190,6 +193,21 @@ func (ctrl *csiSnapshotCommonController) syncSnapshot(snapshot *crdv1.VolumeSnap
...
+	if err != nil {
+		klog.Infof("cannot get claim from snapshot [%s]: [%v] Claim may be deleted already.", snapshot.Name, err)
+		return false
+	}
+	if snapshot.Spec.Source.PersistentVolumeClaimName != nil && pvc.Name == *snapshot.Spec.Source.PersistentVolumeClaimName && !utils.IsSnapshotReady(snapshot) {
```
yuxiangqian, line 204: nit, first two checks are not needed. getClaimFromVolumeSnapshot. already did all checks. if !utils.IsSnapshotReady(...) should. be sufficient.

pkg/common-controller/snapshot_controller.go
xing-yang: Done

pkg/common-controller/snapshot_controller.go
xing-yang: Done

PR #284, opened 2020-04-01
Update Error in Snapshot Status

xing-yang:
**What type of PR is this?**
/kind bug

This PR updates the Error field in the snapshot status based on the status of the content.

Partial fix for #333

```release-note
Updates Error field in the snapshot status based on the status of the content.
```

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -943,6 +943,10 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
 	if content.Status != nil && content.Status.ReadyToUse != nil {
 		readyToUse = *content.Status.ReadyToUse
 	}
+	var volumeSnapshotErr *crdv1.VolumeSnapshotError
+	if content.Status != nil && content.Status.Error != nil {
+		volumeSnapshotErr = content.Status.Error.DeepCopy()
+	}
 
 	klog.V(5).Infof("updateSnapshotStatus: updating VolumeSnapshot [%+v] based on VolumeSnapshotContentStatus [%+v]", snapshot, content.Status)
 
@@ -964,6 +968,9 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
 		if size != nil {
 			newStatus.RestoreSize = resource.NewQuantity(*size, resource.BinarySI)
 		}
+		if volumeSnapshotErr != nil {
+			newStatus.Error = volumeSnapshotErr
+		}
 		updated = true
 	} else {
 		newStatus = snapshotObj.Status.DeepCopy()
@@ -984,6 +991,10 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
 		}
 		if (newStatus.RestoreSize == nil && size != nil) || (newStatus.RestoreSize != nil && newStatus.RestoreSize.IsZero() && size != nil && *size > 0) {
 			newStatus.RestoreSize = resource.NewQuantity(*size, resource.BinarySI)
+			updated = true
+		}
+		if (newStatus.Error == nil && volumeSnapshotErr != nil) || (newStatus.Error != nil && volumeSnapshotErr != nil && newStatus.Error.Time != nil && volumeSnapshotErr.Time != nil && &newStatus.Error.Time != &volumeSnapshotErr.Time) || (newStatus.Error != nil && volumeSnapshotErr == nil) {
+			newStatus.Error = volumeSnapshotErr
 			updated = true
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/common-controller/snapshot_controller.go
```diff
@@ -943,6 +943,18 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
...
+	var volumeSnapshotErr *crdv1.VolumeSnapshotError
+	if content.Status != nil && content.Status.Error != nil {
+		volumeSnapshotErr = &crdv1.VolumeSnapshotError{}
+		if content.Status.Error.Time != nil {
+			errTime := *content.Status.Error.Time
+			volumeSnapshotErr.Time = &errTime
+		}
+		if content.Status.Error.Message != nil {
+			msg := *content.Status.Error.Message
+			volumeSnapshotErr.Message = &msg
```
humblec, line 955: @xing-yang Dont we have to record this event once we set it on the VolumeSnapshot object?

humblec:
This looks good to me, however, got a question on the event submission.

pkg/common-controller/snapshot_controller.go
```diff
@@ -943,6 +943,18 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
 	if content.Status != nil && content.Status.ReadyToUse != nil {
 		readyToUse = *content.Status.ReadyToUse
 	}
+	var volumeSnapshotErr *crdv1.VolumeSnapshotError
+	if content.Status != nil && content.Status.Error != nil {
```
saikat-royc, line 947: To use this change for #333 we need to clear the error in the volume snapshot object, if the content has been created or bound successfully.
```
volumeSnapshotErr := &crdv1.VolumeSnapshotError{}
if content.status != nil && content.Status.Error != nil {
   // update the error time and message
}
```

Thoughts @xing-yang ?

pkg/common-controller/snapshot_controller.go
```diff
@@ -986,6 +1001,10 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
 			newStatus.RestoreSize = resource.NewQuantity(*size, resource.BinarySI)
 			updated = true
 		}
+		if (newStatus.Error == nil && volumeSnapshotErr != nil) || (newStatus.Error != nil && volumeSnapshotErr != nil && newStatus.Error.Time != nil && volumeSnapshotErr.Time != nil && &newStatus.Error.Time != &volumeSnapshotErr.Time) {
```
saikat-royc, line 1004: new status should be updated unconditionally? so that we can clear any error in the volume snapshot object?

pkg/common-controller/snapshot_controller.go
xing-yang: Yes, make sense.,

pkg/common-controller/snapshot_controller.go
xing-yang: Yes

pkg/common-controller/snapshot_controller.go
yuxiangqian: hmmm, isn't the following sufficient?
volumeSnapshotErr = content.Status.Error.DeepCopy()
I'd say just do it in the code blocks where assignment really happens and delete this whole block. WDYT?

pkg/common-controller/snapshot_controller.go
```diff
@@ -964,6 +976,9 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
 		if size != nil {
 			newStatus.RestoreSize = resource.NewQuantity(*size, resource.BinarySI)
 		}
+		if volumeSnapshotErr != nil {
```
yuxiangqian, line 979: i.e., here could be simply:
newStats.Error = content.Status.Error.DeepCopy()?

pkg/common-controller/snapshot_controller.go
yuxiangqian: I am not sure I understand this if statement here: "&newStatus.Error.Time != &volumeSnapshotErr.Time"
and what if newStats.Error != nil but volumeSnapshotErr is nil? in this case, should newStatus.Error be updated to nil as well?

xing-yang:
Review comments addressed.

pkg/common-controller/snapshot_update_test.go
```diff
@@ -327,6 +322,34 @@ func TestSync(t *testing.T) {
 			expectSuccess:    true,
 			test:             testSyncSnapshot,
 		},
+		{
```
saikat-royc, line 325: Thinking of some more test cases:
1. vs object status == nil, vs content object status error != nil -> error expected in volume snapshot object error status.
2. vs object status == nil,  vs content object status error == nil, no change expected in volume snapshot object error status.

xing-yang: Added two new test cases.
Test case 2 behaves like this: When both statuses are nil, updateSnapshotStatus will create an Status and set boundContentName and set readyToUse to false.

Issue #292, opened 2020-04-10
Not able to delete a VolumeSnapshot if it's corresponding VolumeSnapshotContent updates deletion policy from Delete to Retain


yuxiangqian:
Steps to reproduce the issue:
1. Create a dynamic VolumeSnapshot refer to a VolumeSnapshotClass with DeletionPolicy as Delete, this should trigger a VolumeSnapshotContent creation for the given PVC, and both VolumeSnapshot and VolumeSnapshotContent should have the following finalizer:
snapshot.storage.kubernetes.io/volumesnapshot-bound-protection
2. Update VolumeSnapshotContent's DeletionPolicy from Delete to Retain, i.e., kubectl edit <volume-snapshot-content-name> and apply.
3. kubectl delete <volume-snapshot-name>. This should effectively remove the VolumeSnapshot from API server with VolumeSnapshotContent untouched. However the VolumeSnapshot will not be deleted from API server due to the finalizer.

Root cause is here:
https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/common-controller/snapshot_controller.go#L298

The bound finalizer should have been removed from both VolumeSnapshot and VolumeSnapshotContent even if a content exists.

yuxiangqian:
/cc @xing-yang 
When delete a bi-directionally bound VolumeSnapshot and VolumeSnapshotContent with DeletionPolicy set to "Delete" on the content, the original idea is to wait until the VolumeSnapshotContent is removed from API server before removing the bound finalizer on VolumeSnapshot. However that does not apply to the situation when deletion policy is "Retain".

xing-yang:
The fix is merged.  Closing this.

Issue #291, opened 2020-04-09
Deletion of a VolumeSnapshot object can causes deletion of a unbound VolumeSnapshotContent


yuxiangqian:
[checkContentAndBoundStatus](https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/common-controller/snapshot_controller.go#L222) returns four values:
1. the content (NOT necessarily bound to the given VS)
2. content's deletion policy == Delete, a boolean
3. content bound to given snapshot, a boolean
4. any error

[processSnapshotWithDeletionTimestamp](https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/common-controller/snapshot_controller.go#L248) calls **checkContentAndBoundStatus**, then calls **checkandRemoveSnapshotFinalizersAndCheckandDeleteContent** without verifying binding

xing-yang:
Reproduced it:

- create a dynamic snapshot (VS1) from a PVC, this should lead to the controller to create a VolumeSnapshotContent( VSC1)
- Update VSC1's VolumeSnapshotRef such that it points to a different VolumeSnapshot with name VS2, UID left to be empty.
- create a pre-provisioned snapshot (VS2) which points VSC1, the two should bind.
- readyToUse from VS1 becomes False as VSC1 no longer points to VS1.
- delete VS1.  VSC1 will be deleted, however, it is bound to VS2 not VS1.

xing-yang:
The fix is merged.  Closing this.

PR #294, opened 2020-04-13
revamp find content logic to fix 290 291 292

### LINKED ISSUE #290: findContentFromStore in snapshot controller potentially allows rebinding a dynamically provisioned VSC to a pre-provisioned VS
to reproduce:
1. create a dynamic snapshot (VS1) from a PVC, this should lead to the controller to create a VolumeSnapshotContent( VSC1)
2. Update VSC1's VolumeSnapshotRef such that it points to a different VolumeSnapshot with name VS2, UID left to be empty.
3. create a pre-provisioned snapshot (VS2) which points VSC1, the two should bind.

yuxiangqian:
**What type of PR is this?**
/kind bug

fixes issue 290 291 292

Fixes #290 #291 #292 

NONE

```release-note
This PR fixes the following issues in snapshot controller:
Issue #290 : Disallow a pre-provisioned VolumeSnapshot pointing to a dynamically created VolumeSnapshotContent
Issue #291 : Verify VolumeSnapshot and VolumeSnapshotContent are bi-directional bound before initializing a deletion on a VolumeSnapshotContent which the to-be-deleted VolumeSnapshot points to.
Issue #292 : Allow deletion of a VolumeSnapshot/VolumeSnapshotContent when the VolumeSnapshotContent's DeletionPolicy has been updated from Delete to Retain
```


xing-yang:
Can you elaborate what is fixed in this PR in the release notes section as that will go to the changelog.

pkg/common-controller/snapshot_controller.go
```diff
@@ -171,27 +171,22 @@ func (ctrl *csiSnapshotCommonController) syncSnapshot(snapshot *crdv1.VolumeSnap
...
-		if !utils.IsSnapshotReady(snapshot) || !utils.IsBoundVolumeSnapshotContentNameSet(snapshot) {
-			return ctrl.syncUnreadySnapshot(snapshot)
-		}
-		return ctrl.syncReadySnapshot(snapshot)
+		return ctrl.processSnapshotWithDeletionTimestamp(snapshot)
 	}
-	return nil
+	klog.V(5).Infof("syncSnapshot[%s]: check if we should add finalizers on snapshot", utils.SnapshotKey(snapshot))
+	if err := ctrl.checkandAddSnapshotFinalizers(snapshot); err != nil {
+		klog.Errorf("error check and add Snapshot finalizers for snapshot [%s]: %v", snapshot.Name, errFinalizer)
```
xing-yang, line 178: Either use errFinalizer on line 177 or replace errFinalizer with err on line 178.

pkg/common-controller/snapshot_controller.go
```diff
@@ -404,91 +415,160 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
...
-		ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotContentMissing", "VolumeSnapshotContent is missing")
-		klog.V(4).Infof("synchronizing unready snapshot[%s]: snapshotcontent %q requested and not found, will try again next time", uniqueSnapshotName, contentName)
-		return nil, fmt.Errorf("snapshot %s is bound to a non-existing content %s", uniqueSnapshotName, contentName)
+	if content == nil {
+		// no matching content with the desired name has been found in cache
+		return nil, nil
+	}
+	// check whether the content represents a dynamically provisioned snapshot
+	if content.Spec.Source.VolumeHandle == nil {
+		ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotContentMismatch", "VolumeSnapshotContent is pre-provisioned while expecting a dynamically provisioned one")
```
xing-yang, line 539: Since both VolumeHandle and SnapshotHandle are optional fields, we should check SnapshotHandle is not nil as well.

yuxiangqian:
> Can you elaborate what is fixed in this PR in the release notes section as that will go to the changelog.

Done. PTAL

pkg/common-controller/snapshot_controller.go
yuxiangqian: really good catch. done

pkg/common-controller/snapshot_controller.go
```diff
@@ -404,91 +415,160 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
...
-			if err != nil {
+	if contentObj != nil {
+		klog.V(5).Infof("Found VolumeSnapshotContent object %s for snapshot %s", contentObj.Name, uniqueSnapshotName)
+		if contentObj.Spec.Source.SnapshotHandle != nil {
+			ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotHandleSet", fmt.Sprintf("Snapshot handle should not be set in content %s for dynamic provisioning", uniqueSnapshotName))
+			return fmt.Errorf("snapshotHandle should not be set in the content for dynamic provisioning for snapshot %s", uniqueSnapshotName)
+		}
+		newSnapshot, err := ctrl.bindandUpdateVolumeSnapshot(contentObj, snapshot)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
yuxiangqian: done

pkg/common-controller/snapshot_controller.go
yuxiangqian: Again we really need to have webhooks in-place to prevent this kind of situations from happening in the beginning.
The reason why I did not enforce SnapshotHandle == nil here is because its hard to define the behavior if both are specified.
maybe we fail the findContent and update the snapshot with error event, WDYT? I think thats a reasonable implementation as well, if you feel strong about it, I will
enforce that.

yuxiangqian: per discussion today, I will introduce a sanity check in the very beginning of syncSnapshot.

xing-yang:
> Issue #292 : Allow deletion of a VolumeSnapshot/VolumeSnapshotContent when the VolumeSnapshotContent's DeletionPolicy has been updated from Delete to Retain

Should just be "Allow deletion of a VolumeSnapshot when ..." as VolumeSnapshotContent should not be deleted with a Retain policy.

pkg/common-controller/snapshot_controller.go
```diff
@@ -421,6 +421,7 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
 	klog.V(5).Infof("getDynamicallyProvisionedContentFromStore for snapshot %s", uniqueSnapshotName)
 	contentObj, err := ctrl.getDynamicallyProvisionedContentFromStore(snapshot)
 	if err != nil {
+		klog.V(4).Infof("getDynamicallyProvisionedContentFromStore[%s]: error when get content %v", uniqueSnapshotName, err)
```
xing-yang, line 424: Change msg to "error when get content for snapshot" as you gave the snapshot name here.

yuxiangqian:
> > Issue #292 : Allow deletion of a VolumeSnapshot/VolumeSnapshotContent when the VolumeSnapshotContent's DeletionPolicy has been updated from Delete to Retain
> 
> Should just be "Allow deletion of a VolumeSnapshot when ..." as VolumeSnapshotContent should not be deleted with a Retain policy.
>
technically it should be both VolumeSnapshot and VolumeSnapshotContent? Without the fix, there is no way to delete VolumeSnapshot without remove finalizer on it manually. VolumeSnapshotContent deletion is not possible (kubectl delete vsc-name) without this fix as well.

pkg/common-controller/snapshot_controller.go
yuxiangqian: added validation of source in syncSnapshot.

pkg/common-controller/snapshot_controller.go
```diff
@@ -173,6 +171,19 @@ func (ctrl *csiSnapshotCommonController) syncSnapshot(snapshot *crdv1.VolumeSnap
 	if snapshot.ObjectMeta.DeletionTimestamp != nil && !ctrl.isPVCwithFinalizerInUseByCurrentSnapshot(snapshot) {
 		return ctrl.processSnapshotWithDeletionTimestamp(snapshot)
 	}
+
+	// TODO(xiangqian@): Putting this check in controller as webhook has not been implemented
+	//                   yet. Remove the source checking one issue #187 is resolved:
```
xing-yang, line 176: s/one/once

pkg/common-controller/snapshot_controller.go
```diff
@@ -173,6 +171,19 @@ func (ctrl *csiSnapshotCommonController) syncSnapshot(snapshot *crdv1.VolumeSnap
...
+	//                   https://github.com/kubernetes-csi/external-snapshotter/issues/187
+	klog.V(5).Infof("syncSnapshot[%s]: validate snapshot to make sure source has been correctly specified", utils.SnapshotKey(snapshot))
+	if (snapshot.Spec.Source.PersistentVolumeClaimName == nil && snapshot.Spec.Source.VolumeSnapshotContentName == nil) ||
+		(snapshot.Spec.Source.PersistentVolumeClaimName != nil && snapshot.Spec.Source.VolumeSnapshotContentName != nil) {
+		err := fmt.Errorf("Exactly one of PersistentVolumeClaimName and VolumeSnapshotContentName should be specified")
+		klog.Errorf("syncSnapshot[%s]: validation error, %s", utils.SnapshotKey(snapshot), err.Error())
+		ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotValidationError", err.Error())
+		return err
+	}
+
```
xing-yang, line 186: We should check for content source in syncContent too just in case it is modified by mistake by user?

xing-yang:
> technically it should be both VolumeSnapshot and VolumeSnapshotContent? Without the fix, there is no way to delete VolumeSnapshot without remove finalizer on it manually. VolumeSnapshotContent deletion is not possible (kubectl delete vsc-name) without this fix as well.

Technically it is correct but is sounds wrong because it is associated with Retain policy:).

pkg/common-controller/snapshot_update_test.go
```diff
@@ -142,128 +141,167 @@ func TestSync(t *testing.T) {
...
 		{
-			name:              "7-1 - fail to create snapshot with non-existing snapshot class",
-			initialContents:   nocontents,
-			expectedContents:  nocontents,
-			initialSnapshots:  newSnapshotArray("snap7-1", "snapuid7-1", "claim7-1", "", classNonExisting, "", &False, nil, nil, nil, false, true, nil),
-			expectedSnapshots: newSnapshotArray("snap7-1", "snapuid7-1", "claim7-1", "", classNonExisting, "", &False, nil, nil, newVolumeError("Failed to create snapshot content with error failed to get input parameters to create snapshot snap7-1: \"volumesnapshotclass.snapshot.storage.k8s.io \\\"non-existing\\\" not found\""), false, true, nil),
-			initialClaims:     newClaimArray("claim7-1", "pvc-uid7-1", "1Gi", "volume7-1", v1.ClaimBound, &classEmpty),
-			initialVolumes:    newVolumeArray("volume7-1", "pv-uid7-1", "pv-handle7-1", "1Gi", "pvc-uid7-1", "claim7-1", v1.VolumeBound, v1.PersistentVolumeReclaimDelete, classEmpty),
-			expectedEvents:    []string{"Warning SnapshotContentCreationFailed"},
+			name:              "2-10 - (static) do not bind content donest not point to the snapshot",
```
xing-yang, line 144: do not bind content that does not point to the snapshot

pkg/common-controller/snapshot_controller.go
```diff
@@ -215,117 +221,135 @@ func (ctrl *csiSnapshotCommonController) isPVCwithFinalizerInUseByCurrentSnapsho
...
-	// it is removed when it is not needed any more.
-	if utils.NeedToAddSnapshotAsSourceFinalizer(snapshot) {
-		addSourceFinalizer = true
-	}
-	if utils.NeedToAddSnapshotBoundFinalizer(snapshot) && snapshotBound && deleteContent {
-		// Add bound finalizer if snapshot is bound to content and deletion policy is delete
-		addBoundFinalizer = true
+	if content != nil {
+		// A bound finalizer is needed ONLY when all following conditions are satisfied:
+		// 1. the VolumeSnapshot is bound to some content
```
xing-yang, line 348: s/bound to some content/bound to a content

pkg/common-controller/snapshot_controller.go
yuxiangqian: Putting the check ONLY in syncContent is enough, any getXXXContent call should be verifying as syncSnapshot is the one conducting the binding.
I am open to have a check in syncContent and issue an event attach to the content if that helps. WDYT?

pkg/common-controller/snapshot_controller.go
```diff
@@ -155,13 +155,11 @@ func (ctrl *csiSnapshotCommonController) syncContent(content *crdv1.VolumeSnapsh
...
-
 	klog.V(5).Infof("syncSnapshot [%s]: check if we should remove finalizer on snapshot PVC source and remove it if we can", utils.SnapshotKey(snapshot))
 
 	// Check if we should remove finalizer on PVC and remove it if we can.
-	errFinalizer := ctrl.checkandRemovePVCFinalizer(snapshot)
-	if errFinalizer != nil {
-		klog.Errorf("error check and remove PVC finalizer for snapshot [%s]: %v", snapshot.Name, errFinalizer)
+	if err := ctrl.checkandRemovePVCFinalizer(snapshot); err != nil {
+		klog.Errorf("error check and remove PVC finalizer for snapshot [%s]: %v", snapshot.Name, err)
 		// Log an event and keep the original error from syncUnready/ReadySnapshot
```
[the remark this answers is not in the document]
yuxiangqian: noted, will have it fixed when squash.

pkg/common-controller/snapshot_controller.go
```diff
@@ -404,91 +436,162 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
...
+	// check whether the content represents a dynamically provisioned snapshot
+	if content.Spec.Source.VolumeHandle == nil {
+		ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotContentMismatch", "VolumeSnapshotContent is pre-provisioned while expecting a dynamically provisioned one")
+		klog.V(4).Infof("sync snapshot[%s]: snapshot content %s is pre-provisioned while expecting a dynamically provisioned one", utils.SnapshotKey(snapshot), contentName)
+		return nil, fmt.Errorf("snapshot %s expects a dynamically provisioned VolumeSnapshotContent %s but gets a pre-provisioned one", utils.SnapshotKey(snapshot), contentName)
 	}
-	content, ok := contentObj.(*crdv1.VolumeSnapshotContent)
+	// check whether the content points back to the passed in VolumeSnapshot
+	ref := content.Spec.VolumeSnapshotRef
+	if ref.Name != snapshot.Name || ref.Namespace != snapshot.Namespace || ref.UID != snapshot.UID {
```
xing-yang, line 568: s/ref.UID != snapshot.UID/(ref.UID != "" && ref.UID != snapshot.UID)

xing-yang: By the way, did you skip this "ref.UID != " check on purpose so this Misbound event can show up earlier?  If so, please add comments to clarify.

yuxiangqian: I do not think "ref.UID == """ is a legit case here. It's a dynamically created snapshot content, the UID should already have been be set when its created.  If UID == "", it must have
been updated manually which is no longer a legit case.

xing-yang: Can you add these comments before this check?

xing-yang:
@yuxiangqian Just some nits.  Please squash all commits after addressing comments.

pkg/common-controller/snapshot_controller.go
```diff
@@ -502,6 +502,8 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
 // If it found a content which does not point to the passed in VolumeSnapshot, it
 // updates the status of the snapshot with an event and returns an error.
 // Otherwise, the found content will be returned.
+// A content is considered to be a preprivioned one if its Spec.Source.SnapshotHandle
```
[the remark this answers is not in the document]
yuxiangqian: good eyes, thnx

pkg/common-controller/snapshot_controller.go
yuxiangqian: done

pkg/common-controller/snapshot_controller.go
```diff
@@ -559,7 +563,7 @@ func (ctrl *csiSnapshotCommonController) getDynamicallyProvisionedContentFromSto
 	}
 	// check whether the content represents a dynamically provisioned snapshot
 	if content.Spec.Source.VolumeHandle == nil {
-		ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotContentMismatch", "VolumeSnapshotContent is pre-provisioned while expecting a dynamically provisioned one")
+		ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotContentMismatch", "VolumeSnapshotContent "+contentName+" is pre-provisioned while expecting a dynamically provisioned one")
```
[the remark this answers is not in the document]
yuxiangqian: its go-coding styler formatting :).

yuxiangqian:
addressed comments and squashed. thanks Xing

pkg/common-controller/snapshot_controller.go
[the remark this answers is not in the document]
xing-yang: Did a search and found several issues opened about inconsistent gofmt around the + operator depending on context:).

PR #309, opened 2020-04-29
Update docs for controller namespace

jasonltaylor:
Make all namespace settings explicit and add more documentation around
setting appropriate namespace when installing.

**What type of PR is this?**

/kind cleanup

Improves documentation around appropriate namespace setting for the snapshot controller. Makes all namespace settings in YAML files explicit.

Fixes #

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
+++ b/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
@@ -1,8 +1,15 @@
 # RBAC file for the snapshot controller.
+#
+# The snapshot controller implements the control loop for CSI snapshot functionality.
+# It should be installed as part of the base Kubernetes distribution in an appropriate
+# namespace for components implementing base system functionality. For installing with
+# Vanilla Kubernetes, kube-system makes sense for the namespace.
+
 apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: snapshot-controller
+  namespace: default # TODO: replace with the namespace you want for your controller, e.g. kube-system
 
 ---
 kind: ClusterRole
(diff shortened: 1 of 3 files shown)


deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: snapshot-controller
+  namespace: kube-system # TODO: replace with the namespace you want for your controller
```
xing-yang, line 6: I just realized that we can't make this change here because this rbac file is actually used by the prow job.

Instead of changing the namespace to "kube-system", can you just comment it out and add a comment to explain why it makes sense to deploy it in a different namespace?

```
# namespace: kube-system # TODO: replace with the namespace you want for your controller
```

deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
```diff
@@ -44,8 +45,7 @@ metadata:
 subjects:
   - kind: ServiceAccount
     name: snapshot-controller
-    # replace with non-default namespace name
-    namespace: default
+    namespace: kube-system # TODO: replace with the namespace you want for your controller
```
xing-yang, line 48:
```
# replace with the namespace you want for your controller, i.e., kube-system
    namespace: default
```

deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
jasonltaylor: ok - i take it prow cannot override the namespace in the yaml? Should I also update the USAGE.md to direct users to set the namespace to something here:

```
If your cluster does not come pre-installed with the correct components, you may manually install these components by executing the following steps.
```

xing-yang: Prow will just use this rbac directly.

Sure, go ahead and add some comments in the Usage section to clarify about the namespace.

jasonltaylor: Updated README and YAML files with more info about namespace and reverted the namespace to default for use by prow job.

PR #327, opened 2020-06-29
Remove snapshot Finalizer if content deletionPolicy changes

huffmanca:
**What type of PR is this?**
/kind bug

Previously, if a VolumeSnapshot was created from a class with the Delete DeletionPolicy and the content was updated to have the Retain policy, then the VolumeSnapshot would persist as the finalizer remained on it indefinitely. Now, the VolumeSnapshot is updated to remove the finalizer if the associated content has its deletionPolicy updated.

Fixes #

```release-note
If the VolumeSnapshotContent's DeletionPolicy is modified to Retain, then the associated VolumeSnapshot can be successfully deleted.
```

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -135,6 +135,13 @@ func (ctrl *csiSnapshotCommonController) syncContent(content *crdv1.VolumeSnapsh
 			// right away so that it is in-sync with the content status
 			ctrl.snapshotQueue.Add(snapshotName)
 		}
+
+		// Check if we need to remove the content Finalizer from the snapshot because the deletionPolicy has changed
+		if snapshot != nil && utils.NeedToRemoveSnapshotBoundFinalizer(snapshot, content) {
+			// We have the volumesnapshot bound finalizer, but our deletionPolicy has changed to Retain
+			klog.V(5).Infof("syncContent [%s]: Removing Finalizer for VolumeSnapshot", snapshot.Name)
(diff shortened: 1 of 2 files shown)


pkg/common-controller/snapshot_controller.go
```diff
@@ -135,6 +135,13 @@ func (ctrl *csiSnapshotCommonController) syncContent(content *crdv1.VolumeSnapsh
 			// right away so that it is in-sync with the content status
 			ctrl.snapshotQueue.Add(snapshotName)
 		}
+
+		// Check if we need to remove the content Finalizer from the snapshot because the deletionPolicy has changed
+		if snapshot != nil && utils.NeedToRemoveSnapshotBoundFinalizer(snapshot, content) {
+			// We have the volumesnapshot bound finalizer, but our deletionPolicy has changed to Retain
+			klog.V(5).Infof("syncContent [%s]: Removing Finalizer for VolumeSnapshot", snapshot.Name)
+			return ctrl.removeContentFinalizer(snapshot)
```
xing-yang, line 143: I'm not convinced that we need to remove utils.VolumeSnapshotBoundFinalizer from VolumeSnapshot in syncContent.  We've already got this logic in syncSnapshot.

huffmanca: Looks like you're correct - I was testing with an older version of the controller, and it appears that this has been resolved as of v2.1.1.

Apologies. I'll close this PR.

huffmanca:
Closing, as this issue appears to already be addressed as of v2.1.1.

PR #329, opened 2020-07-01
Deletes snapshot if associated content has a deletion timestamp

huffmanca:
**What type of PR is this?**
/kind bug

When running the e2e tests, it seems that ~2% of the time some of the VolumeSnapshot and VolumeSnapshotContent resources remain behind. Both contain a DeletionTimestamp, but the bound finalizer prevents either from ultimately being deleted.

This PR adjusts the way we delete VolumeSnapshots, so that if the associated VolumeSnapshotContent exists but has a DeletionTimestamp, then we'll go ahead and proceed with deleting the VolumeSnapshot.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -327,7 +327,7 @@ func (ctrl *csiSnapshotCommonController) checkandRemoveSnapshotFinalizersAndChec
 	//       by snapshot sidecar controller.
 	//    c. If deletion will not cascade to the content, remove the finalizer on
 	//       the snapshot such that it can be removed from API server.
-	removeBoundFinalizer := !(content != nil && deleteContent)
+	removeBoundFinalizer := !((content != nil && content.ObjectMeta.DeletionTimestamp == nil) && deleteContent)
 	return ctrl.removeSnapshotFinalizer(snapshot, true, removeBoundFinalizer)
 }
 


msau42:
cc @AndiLi99 who is investigating the e2e tests potentially leaking volume and snapshot resouorces

pkg/common-controller/snapshot_controller.go
```diff
@@ -327,7 +327,7 @@ func (ctrl *csiSnapshotCommonController) checkandRemoveSnapshotFinalizersAndChec
 	//       by snapshot sidecar controller.
 	//    c. If deletion will not cascade to the content, remove the finalizer on
 	//       the snapshot such that it can be removed from API server.
-	removeBoundFinalizer := !(content != nil && deleteContent)
+	removeBoundFinalizer := !((content != nil && content.ObjectMeta.DeletionTimestamp == nil) && deleteContent)
```
xing-yang, line 330: The PR title says "Deletes snapshot if associated content has a deletion timestamp" but this code here checks if the DeletionTimestamp == nil?

xing-yang: Also I don't understand why this would fix the problem.  You are adding a condition here to make it stricter so the original code has already contained your change.

huffmanca: If the content exists and has a DeletionTimestamp, then removeBoundFinalizer becomes true since we negate the conditional. When I was testing this earlier, it does end up removing the VolumeSnapshots (which also allows the namespaces to be terminated); however, there may be a better way to go about this.

The state that I've seen is the VolumeSnapshotClass deleted successfully, and then both the VolumeSnapshot and VolumeSnapshotContent persist with the bound finalizer remaining indefinitely.

AndiLi99: I was also confused on this point. However note the !( .. ) which effectively flips the logic. I would suggest distributing the ! operator to make the intention more clear, and to remove the extraneous parenthesis. I think the logic as it is may be flawed, reading the comments above it seems we don't want to remove the finalize if the deleteContent == true, which is not what happens as it is written now. https://en.wikipedia.org/wiki/De_Morgan%27s_laws

huffmanca: This approach also has an issue of leaving the VolumeSnapshotContent, which should likely be deleted.

I think right now in this scenario (VolumeSnapshot & VolumeSnapshotContent remain, both with finalizers/deletionTimestamps, all other resources cleaned up) we're hitting https://github.com/kubernetes-csi/external-snapshotter/blob/4800ca72d4035acc7ae79821fc527b72202e6041/pkg/common-controller/snapshot_controller.go#L157 , which adds the annotation to the Content.

Forgive my ignorance, but is the sidecar-controller typically deployed as part of the e2e test? Might the issue be that this controller is being removed before it has time to clean up the content, which also results in the snapshot (and ultimately the namespace) lingering after the test runs?

Would it make sense to add a check in the common-controller at this line to remove the finalizer from the content if the following are all true:

* VolumeSnapshot has DeletionTimestamp set
* Associated VolumeSnapshotContent has DeletionTimestamp set
* VolumeSnapshotContent has Delete policy
* VolumeSnapshot has bound finalizer
* VolumeSnapshotContent has bound finalizer

Or does this tread too much on the sidecar-controller ?

xing-yang: Can you verify that both the snapshot-controller and the csi-snapshotter sidecar are still running before the snapshot resources are cleaned up?  We need to make sure both are still running, otherwise, we can't guarantee there are no leaked resources.

AndiLi99: > Forgive my ignorance, but is the sidecar-controller typically deployed as part of the e2e test?

As I understand it, there are two ways to run the storage e2e tests: internal and external drivers. The tests itself should be the same but the setup and driver installation methods differ. The internal way involves driver definitions hardcoded into the codebase at https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/drivers/csi.go, these are installed and uninstalled for every test. Prow only tests using the internal driver. Take a look at the raw build for this job and search for "deploying", you will see that the driver is installed for each test. https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/92555/pull-kubernetes-e2e-gce-storage-snapshot/1278407698437640195/.

>  Might the issue be that this controller is being removed before it has time to clean up the content, which also results in the snapshot (and ultimately the namespace) lingering after the test runs?

You might be right on this point and it is worth investigating.

The external driver (which is a stub for any CSI driver to plug into) found here https://github.com/kubernetes/kubernetes/tree/master/test/e2e/storage/external will install the driver once at the beginning of the test suite and delete it at the end.

huffmanca: @AndiLi99 ,

After doing a bit more research this does appear to be caused by the sidecar-controller being torn down before the VolumeSnapshotContents during the e2e tests.

The annotation is present to delete the VolumeSnapshotContent; however, without the sidecar-controller the content is never cleaned up, leaving the namespace present.

I've heard that you're working on addressing this (or a similar) issue in the e2e tests. Do you have an open PR that I can track?

AndiLi99: Please take a look at https://github.com/kubernetes/kubernetes/pull/92555. This is the PR I recently merged which organizes the creation and deletion of snapshot objects such as snapshot content. Take a look at the `SnapshotResource` struct and its cleanup and create methods. It waits until the snapshot objects are fully deleted before ending the tests, which should take care of the issue. However there is nothing to guarantee future tests will use it and so the resource leak may come back.

huffmanca: @AndiLi99 ,

This looks good to me. Thank you for resolving this issue in the existing e2e tests.

@xing-yang ,

Do we want to close this PR and trust that e2e tests should use Andi's pattern above, or do we want to include a check for this issue in the common-controller (in case the sidecar-controller no longer exists)? I'm fine with either approach, and don't mind closing this PR if we trust that the e2e tests are correctly configured going forward.

yuxiangqian: do we really need this?

xing-yang:
@huffmanca I think this PR can be closed now?

xing-yang:
I think this is addressed by other PRs.  Closing this PR.

PR #334, opened 2020-07-21
chore: get rid of k8s.io/kubernetes dep

makkes:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

The k8s.io/kubernetes dependency is problematic because of the many replaces it brings with it. This makes it hard to import external-snapshotter in 3rd party projects without having a lot of `replace` directives in those projects as well.

The only usage of that import was to get two very simple utility functions so I copied them over to this repo.

no issue

```release-note
NONE
```


pkg/utils/util.go
```diff
@@ -109,6 +108,41 @@ var SnapshotterListSecretParams = secretParamsMap{
 	secretNamespaceKey: PrefixedSnapshotterListSecretNamespaceKey,
 }
 
+// ContainsString checks if a given slice of strings contains the provided string.
+// If a modifier func is provided, it is called with the slice item before the comparation.
+func ContainsString(slice []string, s string, modifier func(s string) string) bool {
```
yuxiangqian, line 113: nit, can we remove the modifier? it appears its not used at all in this repo (i.e., all nils).

pkg/utils/util.go
```diff
@@ -109,6 +108,41 @@ var SnapshotterListSecretParams = secretParamsMap{
...
+		if modifier != nil && modifier(item) == s {
+			return true
+		}
+	}
+	return false
+}
+
+// RemoveString returns a newly created []string that contains all items from slice that
+// are not equal to s and modifier(s) in case modifier func is provided.
+func RemoveString(slice []string, s string, modifier func(s string) string) []string {
```
yuxiangqian, line 127: ditto

pkg/utils/util.go
makkes: done

pkg/utils/util.go
makkes: done

makkes:
Thanks, @yuxiangqian for reviewing. I have addressed your comments.

yuxiangqian:
lgtm,  will wait until all tests passed to approve.

makkes:
@yuxiangqian would you mind having a look, please? 🙇‍♂️

PR #339, opened 2020-07-27
Update controller images

### LINKED ISSUE #336: VolumeSnapshotContent with Retain policy cannot be deleted
I created a VolumeSnapshot dynamically with deletionPolicy as "Retain". Then I deleted the VolumeSnapshot, it got deleted but the corresponding VolumeSnapshotContent remained, which is expected.
Now I explicitly tried to delete the VolumeSnapshotContent but it never gets deleted.

One observation is that the "volumesnapshotcontent-bound-protection" finalizer is still there on the object. The code to remove this finalizer is never getting invoked in case of Retain policy. The culprit seems to be sidecar-controller/snapshot_controller.shouldDelete(). The only conditions when it can return true are-

        if content.Spec.Source.SnapshotHandle != nil && content.Spec.VolumeSnapshotRef.UID == "" 
and 

	if metav1.HasAnnotation(content.ObjectMeta, utils.AnnVolumeSnapshotBeingDeleted) 

Neither of these are true in case of the VolumeSnapshotContent in concern and hence the finalizer removal code is never invoked. However if I manually delete the finalizer by editing the object, the VolumeSnapshotContent gets deleted.

Snapshotter image version used: v2.1.1
Kubernetes version: v1.18.2

Kartik494:
**What type of PR is this?**

/kind bug

Fixes #336 

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
+++ b/deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
@@ -84,7 +84,7 @@ spec:
             - name: socket-dir
               mountPath: /csi
         - name: csi-snapshotter
-          image: quay.io/k8scsi/csi-snapshotter:v2.0.1
+          image: quay.io/k8scsi/csi-snapshotter:canary
           args:
             - "--v=5"
             - "--csi-address=$(ADDRESS)"
--- a/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml
+++ b/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml
@@ -25,7 +25,7 @@ spec:
       serviceAccount: snapshot-controller
       containers:
         - name: snapshot-controller
-          image: quay.io/k8scsi/snapshot-controller:v2.0.1
+          image: quay.io/k8scsi/snapshot-controller:canary
           args:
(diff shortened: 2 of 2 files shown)


deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
```diff
@@ -84,7 +84,7 @@ spec:
             - name: socket-dir
               mountPath: /csi
         - name: csi-snapshotter
-          image: quay.io/k8scsi/csi-snapshotter:v2.0.1
+          image: quay.io/k8scsi/csi-snapshotter:v2.1.1
```
xing-yang, line 87: In the master, can you just update the image to canary?
In the stable branches, can up submit separate PRs to update image to specific versions?

Kartik494: @xing-yang i have addressed the changes.PTAL

xing-yang:
@Kartik494 @ggriffiths we should use images from the new repos now:
k8s.gcr.io/sig-storage/csi-snapshotter:v2.1.1

xing-yang:
Never mind.  We don't have canary images in the new repos.  So for now, let's continue to use canary images in quay.io for master until we have something else.

PR #347, opened 2020-08-09
Emit event even if status update fails

### LINKED ISSUE #346: Failure events are not propagated to VolumeSnapshot or VolumeSnapshotContent
In the case when driver fails to create a snapshot and returns an appropriate GRPC error code, the same is not propagated to the VolumeSnapshot object. The VolumeSnapshot remains in ReadyToUse=False as expected-

```
# kubectl get volumesnapshot
NAME          READYTOUSE   SOURCEPVC   SOURCESNAPSHOTCONTENT   RESTORESIZE   SNAPSHOTCLASS   SNAPSHOTCONTENT                                    CREATIONTIME   AGE
snapdep5      false        pvcdep4                                           snapclass1      snapcontent-9c441220-ce0d-4b41-8d0a-9b0372082495                  10m
```

But describe on the object does not show any failure events, it continues to show "CreatingSnapshot" as below-

```
Events:
  Type    Reason            Age   From                 Message
  ----    ------            ----  ----                 -------
  Normal  CreatingSnapshot  12m   snapshot-controller  Waiting for a snapshot default/snapdep5 to be created by the CSI driver.
```

Similarly the corresponding VolumeSnapshotContent doesnt show any events.

**Env Details:**
Kubernetes v1.18.2
Snapshotter v2.2.0-rc2
Red Hat Enterprise Linux Server 7.6

xing-yang:
**What type of PR is this?**
/kind bug

This PR moves the event up so event will be emitted even if snapshot/content status update fails so user can see the error.

Fixes #346 

```release-note
Emit event even if status update fails.
```

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -741,13 +741,13 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotErrorStatusWithEvent(snap
 	snapshotClone.Status.ReadyToUse = &ready
 	newSnapshot, err := ctrl.clientset.SnapshotV1beta1().VolumeSnapshots(snapshotClone.Namespace).UpdateStatus(context.TODO(), snapshotClone, metav1.UpdateOptions{})
 
+	// Emit the event even if the status update fails so that user can see the error
+	ctrl.eventRecorder.Event(newSnapshot, eventtype, reason, message)
+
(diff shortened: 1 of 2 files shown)


xing-yang:
Discussed this at today's CSI meeting.  Event does not have to be tied to status update so it is fine to emit event even if status update fails.

PR #354, opened 2020-08-12
Check for non-CSI PVs and mark not in use as appropriate

huffmanca:
**What type of PR is this?**
/kind bug

Currently, if you specify a PVC and that PVC was created by an in-tree driver we never create the content and cannot delete the snapshot, as the source finalizer is never removed.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -819,8 +819,17 @@ func (ctrl *csiSnapshotCommonController) ensurePVCFinalizer(snapshot *crdv1.Volu
 		return newControllerUpdateError(pvc.Name, "cannot add finalizer on claim because it is being deleted")
 	}
 
+	// Check to see if the backing PV is created by a CSI driver. If not, then remove the Finalizer,
+	// as the snapshot can't be created from this PVC
+	pv, err := ctrl.getVolumeFromVolumeSnapshot(snapshot)
+	if err != nil {
+		klog.Infof("cannot get volume from snapshot [%s]: [%v]", snapshot.Name, err)
+		return newControllerUpdateError(snapshot.Name, "cannot get volume from claim")
+	}
+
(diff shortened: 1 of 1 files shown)


huffmanca:
This approach (checking to see if the backing volume is CSI before adding the finalizer to the PVC) has worked when I've manually tested it. The VolumeSnapshot can be deleted, as the PVC doesn't have the finalizer when the snapshot is scheduled for deletion.

pkg/common-controller/snapshot_controller.go
```diff
@@ -819,8 +819,17 @@ func (ctrl *csiSnapshotCommonController) ensurePVCFinalizer(snapshot *crdv1.Volu
 		return newControllerUpdateError(pvc.Name, "cannot add finalizer on claim because it is being deleted")
 	}
 
+	// Check to see if the backing PV is created by a CSI driver. If not, then remove the Finalizer,
+	// as the snapshot can't be created from this PVC
+	pv, err := ctrl.getVolumeFromVolumeSnapshot(snapshot)
```
xing-yang, line 824: This could fail for other reasons.  This method does not really check if PV is backed by a CSI driver.

xing-yang:
I have submitted this PR https://github.com/kubernetes-csi/external-snapshotter/pull/360 for the finalizer deletion problem.  Can you give a try?

For the problem that you reported, I think we should check early and fail early.

After the call to checkAndUpdateSnapshotClass in common-controller/snapshot_controller_base.go, we can add a small function with the following steps:
- If snapshotClass is found, get snapshotClass.
- find StorageClass from Snapshot’s source PVC.
- Compare storageClass.Provisioner with snapshotClass.Driver.
- If they are different, return error.

huffmanca:
@xing-yang ,

I tested your #360 PR, and it addresses this issue. I am able to successfully delete a VolumeSnapshot that is backed by a non-CSI PVC. The finalizer is also removed from the PVC.

I'll close this PR, and have commented on #360.

PR #375, opened 2020-09-15
Pass snapshot metadata to CSI driver

### LINKED ISSUE #341: Provide an option to pass snapshot metadata to csi driver
When external snapshotter sends a request to the csi-driver it doesn't contain any additional information about the snapshot name, snapshot namespace, or the snapshot content name, it would be nice if we can provide an option to do the same. currently the same feature is available in external-provisioner to pass the PVC metadata (PVC name, namespace, and PV name) as additional parameters in createvolume requests when `--extra-create-metadata`.

cc @xing-yang 

Let me know if it sounds good, I can work on it.

chrishenzie:
@msau42 @xing-yang @Madhu-1 

**What type of PR is this?**

/kind feature

This PR introduces an `--extra-create-metadata` flag to the CSI snapshotter.

When set to true, the VolumeSnapshot's name / namespace and the VolumeSnapshotContent's name will be passed as parameters to the CSI driver in `CreateSnapshotRequest`.

Fixes #341

In the CSI provisioner we have a `csi.storage.k8s.io/` prefix defined for metadata passed as parameters to the CSI driver:

https://github.com/kubernetes-csi/external-provisioner/blob/b1d315e4b8dbc7e4a948ee2b9b3d7a58892cfa57/pkg/controller/controller.go#L108-L111

I want to use this same prefix in the CSI snapshotter for consistency, however in this repo this prefix is intended for parameters to remove.

https://github.com/kubernetes-csi/external-snapshotter/blob/60202d2e3149648af816b16525e51a457e66c292/pkg/utils/util.go#L49-L53

I modified the existing method that prunes these prefixed parameters to exclude these few. Any suggestions on refactoring to improve readability / code quality here are appreciated.

Outside of unit tests, I deployed this in a real cluster and observed the following creating a snapshot (taken from AWS EBS CSI driver logs):
```
I0914 21:30:40.192053       1 controller.go:412] CreateSnapshot: called with args source_volume_id:"vol-057e3e0fc815fe8dd" name:"snapshot-8f42c931-29f0-4755-8e3e-4f49ceb94dcf" parameters:<key:"csi.storage.k8s.io/volumesnapshot/name" value:"snapshot" > parameters:<key:"csi.storage.k8s.io/volumesnapshot/namespace" value:"default" > parameters:<key:"csi.storage.k8s.io/volumesnapshotcontent/name" value:"snapcontent-8f42c931-29f0-4755-8e3e-4f49ceb94dcf" >
```

```release-note
The CSI snapshotter has a new `--extra-create-metadata` flag.

When set to true, the name and namespace of the source VolumeSnapshot, and the name of the source  VolumeSnapshotContent will be passed as parameters to the CSI driver in CreateSnapshotRequest.
```

### DIFF
--- a/cmd/csi-snapshotter/main.go
+++ b/cmd/csi-snapshotter/main.go
@@ -62,6 +62,7 @@ var (
 	showVersion            = flag.Bool("version", false, "Show version.")
 	threads                = flag.Int("worker-threads", 10, "Number of worker threads.")
 	csiTimeout             = flag.Duration("timeout", defaultCSITimeout, "The timeout for any RPCs to the CSI driver. Default is 1 minute.")
+	extraCreateMetadata    = flag.Bool("extra-create-metadata", false, "If set, add snapshot metadata to plugin snapshot requests as parameters.")
 
 	leaderElection          = flag.Bool("leader-election", false, "Enables leader election.")
 	leaderElectionNamespace = flag.String("leader-election-namespace", "", "The namespace where the leader election resource exists. Defaults to the pod namespace if not set.")
@@ -173,6 +174,7 @@ func main() {
 		*resyncPeriod,
 		*snapshotNamePrefix,
 		*snapshotNameUUIDLength,
+		*extraCreateMetadata,
 	)
 
 	run := func(context.Context) {
--- a/pkg/utils/util.go
+++ b/pkg/utils/util.go
@@ -49,7 +49,9 @@ const (
 	// CSI Parameters prefixed with csiParameterPrefix are not passed through
 	// to the driver on CreateSnapshotRequest calls. Instead they are intended
 	// to be used by the CSI external-snapshotter and maybe used to populate
-	// fields in subsequent CSI calls or Kubernetes API objects.
+	// fields in subsequent CSI calls or Kubernetes API objects. An exception
+	// exists for the volume snapshot and volume snapshot content keys, which are
+	// passed as parameters on CreateSnapshotRequest calls.
 	csiParameterPrefix = "csi.storage.k8s.io/"
 
 	PrefixedSnapshotterSecretNameKey      = csiParameterPrefix + "snapshotter-secret-name"      // Prefixed name key for DeleteSnapshot secret
@@ -57,6 +59,10 @@ const (
 
 	PrefixedSnapshotterListSecretNameKey      = csiParameterPrefix + "snapshotter-list-secret-name"      // Prefixed name key for ListSnapshots secret
 	PrefixedSnapshotterListSecretNamespaceKey = csiParameterPrefix + "snapshotter-list-secret-namespace" // Prefixed namespace key for ListSnapshots secret
+
+	PrefixedVolumeSnapshotNameKey        = csiParameterPrefix + "volumesnapshot/name"        // Prefixed VolumeSnapshot name key
+	PrefixedVolumeSnapshotNamespaceKey   = csiParameterPrefix + "volumesnapshot/namespace"   // Prefixed VolumeSnapshot namespace key
+	PrefixedVolumeSnapshotContentNameKey = csiParameterPrefix + "volumesnapshotcontent/name" // Prefixed VolumeSnapshotContent name key
 
 	// Name of finalizer on VolumeSnapshotContents that are bound by VolumeSnapshots
 	VolumeSnapshotContentFinalizer = "snapshot.storage.kubernetes.io/volumesnapshotcontent-bound-protection"
--- a/pkg/sidecar-controller/content_create_test.go
+++ b/pkg/sidecar-controller/content_create_test.go
@@ -41,6 +41,11 @@ func TestSyncContent(t *testing.T) {
 					snapshotName: "snapshot-snapuid1-1",
 					driverName:   mockDriverName,
 					snapshotId:   "snapuid1-1",
+					parameters: map[string]string{
+						utils.PrefixedVolumeSnapshotNameKey:        "snap1-1",
+						utils.PrefixedVolumeSnapshotNamespaceKey:   "default",
+						utils.PrefixedVolumeSnapshotContentNameKey: "content1-1",
+					},
 					creationTime: timeNow,
 					readyToUse:   true,
 				},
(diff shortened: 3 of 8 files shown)


yuxiangqian:
please see my comments in #341

pkg/utils/util.go
```diff
@@ -58,6 +60,10 @@ const (
 	PrefixedSnapshotterListSecretNameKey      = csiParameterPrefix + "snapshotter-list-secret-name"      // Prefixed name key for ListSnapshots secret
 	PrefixedSnapshotterListSecretNamespaceKey = csiParameterPrefix + "snapshotter-list-secret-namespace" // Prefixed namespace key for ListSnapshots secret
 
+	PrefixedVolumeSnapshotNameKey        = csiParameterPrefix + "volumesnapshot/name"        // Prefixed name key for CreateSnapshot
```
xing-yang, line 63: Prefixed name key -> Prefixed volumesnapshot name key

pkg/utils/util.go
```diff
@@ -58,6 +60,10 @@ const (
...
+	PrefixedVolumeSnapshotNamespaceKey   = csiParameterPrefix + "volumesnapshot/namespace"   // Prefixed namespace key for CreateSnapshot
```
xing-yang, line 64: Prefixed namespace key -> Prefixed volumesnapshot namespace key

pkg/utils/util.go
```diff
@@ -58,6 +60,10 @@ const (
...
+	PrefixedVolumeSnapshotContentNameKey = csiParameterPrefix + "volumesnapshotcontent/name" // Prefixed name key for CreateSnapshot
```
xing-yang, line 65: Prefixed name key -> Prefixed volumesnapshotcontent name key

pkg/utils/util.go
```diff
@@ -49,7 +49,9 @@ const (
 	// CSI Parameters prefixed with csiParameterPrefix are not passed through
 	// to the driver on CreateSnapshotRequest calls. Instead they are intended
 	// to be used by the CSI external-snapshotter and maybe used to populate
-	// fields in subsequent CSI calls or Kubernetes API objects.
+	// fields in subsequent CSI calls or Kubernetes API objects. An exception
+	// exists for the volume snapshot and volume snapshot content keys, which are
+	// passed as parameters on CreateSnapshotRequest calls.
```
yuxiangqian, line 54: I think we should use a different prefix according to the official document:
https://kubernetes.io/blog/2019/01/15/container-storage-interface-ga/#dynamic-provisioning
Reusing this prefix does not bring too much value but could easily cause confusion.

pkg/utils/util.go
```diff
@@ -460,7 +466,21 @@ func RemovePrefixedParameters(param map[string]string) (map[string]string, error
 	newParam := map[string]string{}
 	for k, v := range param {
 		if strings.HasPrefix(k, csiParameterPrefix) {
-			// Check if its well known
+			// Check if it should be included
```
yuxiangqian, line 469: this check seems to be redundant?

cmd/csi-snapshotter/main.go
```diff
@@ -62,6 +62,7 @@ var (
 	showVersion            = flag.Bool("version", false, "Show version.")
 	threads                = flag.Int("worker-threads", 10, "Number of worker threads.")
 	csiTimeout             = flag.Duration("timeout", defaultCSITimeout, "The timeout for any RPCs to the CSI driver. Default is 1 minute.")
+	extraCreateMetadata    = flag.Bool("extra-create-metadata", false, "If set, add snapshot metadata to plugin snapshot requests as parameters.")
```
yuxiangqian, line 65: IMO, this is a per-snapshot-class setting rather than controller level toggle, and here is why:
In a situation where there are more than one (say two) csi drivers in a k8s cluster.
CSI-A needs these three parameters, CSI-B does not need (or even worse it would fail if non-recognizable parameters have been passed in), then having a controller level toggle will cause CSI-B not working.

My take is to have predefined keys defined in controller, but require their existences in "parameters" of VolumeSnapshot class for the controller to add those extra parameters at runtime.

msau42: @yuxiangqian the csi-snapshotter is run per csi driver. I don't see why this needs to be per SnapshotClass.

pkg/utils/util.go
msau42: I think we should be consistent with the prefix we use in external-provisioner:
https://github.com/kubernetes-csi/external-provisioner/blob/0c340861ea33bd03464d978cbc0fd2d14dba251c/pkg/controller/controller.go#L109

pkg/utils/util.go
chrishenzie: Can you clarify how it is redundant? Are you referring to checking the prefix? From my understanding, we need this check to return an error if it's using the prefix but we don't know about it (if it doesn't match any of the cases in the second switch statement).

chrishenzie:
Are there any additional blockers, or is this suitable for LGTM?

pkg/utils/util.go
chrishenzie: Oh I see, good catch, yes this is redundant. Will remove.

chrishenzie: Undid the changes to this function, but updated the `util_test.go` to include the two other prefixed parameters that should be removed (weren't covered in tests).

cmd/csi-snapshotter/main.go
yuxiangqian: please ignore this comment as its irrelevant. I thought this flag was in the snapshot controller.

yuxiangqian:
/lgtm
@msau42 @xing-yang I am still concerned using this "csi.storage.k8s.io" as the prefix of the parameter names passed down to CSI driver. The current implementation would simply ignore any parameters in "csi.storage.k8s.io" namespace in the storage class, and the BETA documentation says differently. Either we should fix the controller([code](https://github.com/kubernetes-csi/external-snapshotter/blob/f705b23bbd2941da4ab4301539a96ec009121719/pkg/utils/util.go#L476)) such that it does not ignore parameters other than secret related ones, or we need to fix the document.

msau42:
> /lgtm
> Michelle Au Xing Yang I am still concerned using this "csi.storage.k8s.io" as the prefix of the parameter names passed down to CSI driver. The current implementation would simply ignore/error on any parameters in "csi.storage.k8s.io" namespace provided in storage class Parameters list, and the BETA documentation says differently. Either we should fix the controller([code](https://github.com/kubernetes-csi/external-snapshotter/blob/f705b23bbd2941da4ab4301539a96ec009121719/pkg/utils/util.go#L476)) such that it does not ignore parameters other than secret related ones, or we need to fix the document.

I think we should update the documentation since external-provisioner is already using this prefix.

chrishenzie:
See https://github.com/kubernetes-csi/docs/pull/372 for the documentation updates. This PR will need to be merged first and a release cut before the docs PR can be merged.

yuxiangqian:
@chrishenzie Thanks for the hard work!
/lgtm
/approve

PR #378, opened 2020-09-22
Fix image name for example webhook, and other nitpicks

bswartz:
**What type of PR is this?**
/kind cleanup

The image name for the webhook was wrong in the example.
Also fixed a few nitpicks

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
+++ b/deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
@@ -72,7 +72,7 @@ spec:
       serviceAccount: csi-snapshotter
       containers:
         - name: csi-provisioner
-          image: quay.io/k8scsi/csi-provisioner:v1.5.0
+          image: k8s.gcr.io/sig-storage/csi-provisioner:v2.0.2
           args:
             - "--v=5"
             - "--csi-address=$(ADDRESS)"
@@ -84,7 +84,7 @@ spec:
             - name: socket-dir
               mountPath: /csi
         - name: csi-snapshotter
-          image: quay.io/k8scsi/csi-snapshotter:canary
+          image: k8s.gcr.io/sig-storage/csi-snapshotter:v3.0.0
           args:
             - "--v=5"
             - "--csi-address=$(ADDRESS)"
@@ -97,7 +97,7 @@ spec:
             - name: socket-dir
               mountPath: /csi
         - name: hostpath
-          image: quay.io/k8scsi/hostpathplugin:v1.2.0
+          image: k8s.gcr.io/sig-storage/hostpathplugin:v1.4.0
           args:
             - "--v=5"
             - "--endpoint=$(CSI_ENDPOINT)"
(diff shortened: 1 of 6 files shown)


deploy/kubernetes/webhook-example/webhook.yaml
```diff
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
       - name: snapshot-validation
-        image: k8s.gcr.io/sig-storage/validation-webhook:v2.2.0 # change the image if you wish to use your own custom validation server image
+        image: k8s.gcr.io/sig-storage/snapshot-validation-webhook:v2.2.0 # change the image if you wish to use your own custom validation server image
```
xing-yang, line 20: This should be v3.0.0.
We didn't cut v2.2.0.  We cut v3.3.0.

xing-yang (commented):
Can you also update the images under external-snapshotter/deploy?
Images should be updated to the following:
```
k8s.gcr.io/sig-storage/snapshot-controller:v3.0.0
k8s.gcr.io/sig-storage/csi-snapshotter:v3.0.0
```

xing-yang:
We should update this one too:

https://github.com/kubernetes-csi/external-snapshotter/blob/master/deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml#L75
New image:
k8s.gcr.io/sig-storage/csi-provisioner:v2.0.2

We also need to update the hostpath driver when the new image is cut:
https://github.com/kubernetes-csi/external-snapshotter/blob/master/deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml#L100

deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
```diff
@@ -72,7 +72,7 @@ spec:
       serviceAccount: csi-snapshotter
       containers:
         - name: csi-provisioner
-          image: quay.io/k8scsi/csi-provisioner:v1.5.0
+          image: k8s.gcr.io/sig-storage/csi-provisioner:v1.5.0
```
xing-yang, line 75: The latest is v2.0.2

deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
```diff
@@ -97,7 +97,7 @@ spec:
             - name: socket-dir
               mountPath: /csi
         - name: hostpath
-          image: quay.io/k8scsi/hostpathplugin:v1.2.0
+          image: k8s.gcr.io/sig-storage/hostpathplugin:v1.2.0
```
xing-yang, line 100: This image does not exist.  Let's wait for the new csi hostpath driver image to be cut.

yuxiangqian: all images will go to k8s.gcr.io in the future?

msau42: You can use hostpath driver 1.4.0

msau42:
@xing-yang longer term, do you see a way to consolidate all these image examples so we don't have to remember to update all these places every time we cut new releases? I am thinking things like making the image version templatable and using git tags or something to get the latest released version.

PR #388, opened 2020-10-02
Upgrade csi-lib-utils to 0.8.1

Jiawei0227:
**What type of PR is this?**
/kind bug

Upgrade csi-lib-utils to version 0.8.1

```release-note
Cherry-pick from https://github.com/kubernetes-csi/external-snapshotter/pull/387. Add `process_start_time_seconds` metric so that we can properly extrapolate the beginning part of the histogram data.
```


xing-yang:
@Jiawei0227 what is the difference between this PR and this other PR that is already merged?  https://github.com/kubernetes-csi/external-snapshotter/pull/387

xing-yang:
I see.  Looks like https://github.com/kubernetes-csi/external-snapshotter/pull/387 is a commit to the master while this one commits to release-3.0.

So this one should be a cherry-pick then.

xing-yang:
By the way, this is actually a new feature.  We should not be cherry-picking new features into a stable branch.

/hold

msau42:
This is actually a bug fix. Without this process_start_time metric, we cannot properly extrapolate the beginning part of the histogram data.

PR #396, opened 2020-10-09
Fix the failing multiarch builds

namrata-ibm:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: As distroless image is now multiarch, image push job fails on s390x and ppc64le due to lack of support.

Fixes # https://github.com/kubernetes-csi/csi-release-tools/issues/105

**Special notes for your reviewer**: Workaround to allow builds to pass while we add support to distroless image.

```release-note
Changing distroless base image from multiarch to -amd64 to fix Prow builds
```


namrata-ibm:
cc @msau42

PR #409, opened 2020-10-28
Add snapshot controller metrics

### LINKED ISSUE #142: Add Metrics to the snapshot controller

ggriffiths:

**What type of PR is this?**
/kind feature

Adds snapshot controller metrics

Fixes #142 

Pending work:
~~1. Fix metric tests in pkg/metrics~~
~~2. emit metrics for non-success statuses~~

```release-note
The `--metrics-address` and `--metrics-path` flags have been added to support metrics for the snapshot-controller. To enable metrics, the user must configure a `--metrics-path` such as `:9102`. The default value for `--metrics-path` is the empty string, meaning this feature is disabled by default. The default value for `--metrics-path` is `/metrics`.
```

### DIFF
--- a/cmd/snapshot-controller/main.go
+++ b/cmd/snapshot-controller/main.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"sync"
 	"time"
 
 	"k8s.io/client-go/kubernetes"
@@ -32,6 +33,7 @@ import (
 
 	"github.com/kubernetes-csi/csi-lib-utils/leaderelection"
 	controller "github.com/kubernetes-csi/external-snapshotter/v3/pkg/common-controller"
+	"github.com/kubernetes-csi/external-snapshotter/v3/pkg/metrics"
 
 	clientset "github.com/kubernetes-csi/external-snapshotter/client/v3/clientset/versioned"
 	snapshotscheme "github.com/kubernetes-csi/external-snapshotter/client/v3/clientset/versioned/scheme"
@@ -48,6 +50,9 @@ var (
 
 	leaderElection          = flag.Bool("leader-election", false, "Enables leader election.")
 	leaderElectionNamespace = flag.String("leader-election-namespace", "", "The namespace where the leader election resource exists. Defaults to the pod namespace if not set.")
+
+	httpEndpoint = flag.String("http-endpoint", "", "The TCP network address where the HTTP server for diagnostics, including metrics, will listen (example: :8080). The default is empty string, which means the server is disabled.")
+	metricsPath  = flag.String("metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
 )
 
 var (
@@ -87,6 +92,28 @@ func main() {
 	factory := informers.NewSharedInformerFactory(snapClient, *resyncPeriod)
 	coreFactory := coreinformers.NewSharedInformerFactory(kubeClient, *resyncPeriod)
 
+	// Create and register metrics manager
+	metricsManager := metrics.NewMetricsManager()
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	if *httpEndpoint != "" {
+		srv, err := metricsManager.StartMetricsEndpoint(*metricsPath, *httpEndpoint, promklog{}, wg)
+		if err != nil {
+			klog.Errorf("Failed to start metrics server: %s", err.Error())
+			os.Exit(1)
+		}
+		defer func() {
+			err := srv.Shutdown(context.Background())
+			if err != nil {
+				klog.Errorf("Failed to shutdown metrics server: %s", err.Error())
+			}
+
+			klog.Infof("Metrics server successfully shutdown")
+			wg.Done()
+		}()
+		klog.Infof("Metrics server successfully started on %s, %s", *httpEndpoint, *metricsPath)
+	}
+
 	// Add Snapshot types to the default Kubernetes so events can be logged for them
 	snapshotscheme.AddToScheme(scheme.Scheme)
 
@@ -99,6 +126,7 @@ func main() {
 		factory.Snapshot().V1beta1().VolumeSnapshotContents(),
 		factory.Snapshot().V1beta1().VolumeSnapshotClasses(),
 		coreFactory.Core().V1().PersistentVolumeClaims(),
+		metricsManager,
 		*resyncPeriod,
 	)
 
@@ -141,4 +169,10 @@ func buildConfig(kubeconfig string) (*rest.Config, error) {
 		return clientcmd.BuildConfigFromFlags("", kubeconfig)
 	}
 	return rest.InClusterConfig()
+}
+
+type promklog struct{}
+
+func (pl promklog) Println(v ...interface{}) {
+	klog.Error(v...)
 }
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -32,6 +32,7 @@ import (
 	klog "k8s.io/klog/v2"
 
 	crdv1 "github.com/kubernetes-csi/external-snapshotter/client/v3/apis/volumesnapshot/v1beta1"
+	"github.com/kubernetes-csi/external-snapshotter/v3/pkg/metrics"
 	"github.com/kubernetes-csi/external-snapshotter/v3/pkg/utils"
 )
 
@@ -235,6 +236,20 @@ func (ctrl *csiSnapshotCommonController) syncSnapshot(snapshot *crdv1.VolumeSnap
 // 2. Call checkandRemoveSnapshotFinalizersAndCheckandDeleteContent() with information obtained from step 1. This function name is very long but the name suggests what it does. It determines whether to remove finalizers on snapshot and whether to delete content.
 func (ctrl *csiSnapshotCommonController) processSnapshotWithDeletionTimestamp(snapshot *crdv1.VolumeSnapshot) error {
 	klog.V(5).Infof("processSnapshotWithDeletionTimestamp VolumeSnapshot[%s]: %s", utils.SnapshotKey(snapshot), utils.GetSnapshotStatusForLogging(snapshot))
+	driverName, err := ctrl.getSnapshotDriverName(snapshot)
+	if err != nil {
+		klog.Errorf("failed to getSnapshotDriverName while recording metrics for snapshot %q: %v", utils.SnapshotKey(snapshot), err)
+	}
+
+	snapshotProvisionType := metrics.DynamicSnapshotType
+	if snapshot.Spec.Source.VolumeSnapshotContentName != nil {
+		snapshotProvisionType = metrics.PreProvisionedSnapshotType
+	}
+
+	// Processing delete, start operation metric
+	deleteOperationKey := metrics.NewOperationKey(metrics.DeleteSnapshotOperationName, snapshot.UID)
+	deleteOperationValue := metrics.NewOperationValue(driverName, snapshotProvisionType)
+	ctrl.metricsManager.OperationStart(deleteOperationKey, deleteOperationValue)
 
 	var contentName string
 	if snapshot.Status != nil && snapshot.Status.BoundVolumeSnapshotContentName != nil {
@@ -269,6 +284,7 @@ func (ctrl *csiSnapshotCommonController) processSnapshotWithDeletionTimestamp(sn
 	}
 
 	klog.V(5).Infof("processSnapshotWithDeletionTimestamp[%s]: delete snapshot content and remove finalizer from snapshot if needed", utils.SnapshotKey(snapshot))
+
 	return ctrl.checkandRemoveSnapshotFinalizersAndCheckandDeleteContent(snapshot, content, deleteContent)
 }
 
@@ -388,6 +404,7 @@ func (ctrl *csiSnapshotCommonController) syncReadySnapshot(snapshot *crdv1.Volum
 		// snapshot is bound but content is not pointing to the snapshot
 		return ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotMisbound", "VolumeSnapshotContent is not bound to the VolumeSnapshot correctly")
 	}
+
 	// everything is verified, return
 	return nil
 }
@@ -396,6 +413,28 @@ func (ctrl *csiSnapshotCommonController) syncReadySnapshot(snapshot *crdv1.Volum
 func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.VolumeSnapshot) error {
 	uniqueSnapshotName := utils.SnapshotKey(snapshot)
 	klog.V(5).Infof("syncUnreadySnapshot %s", uniqueSnapshotName)
+	driverName, err := ctrl.getSnapshotDriverName(snapshot)
+	if err != nil {
+		klog.Errorf("failed to getSnapshotDriverName while recording metrics for snapshot %q: %s", utils.SnapshotKey(snapshot), err)
+	}
+
+	snapshotProvisionType := metrics.DynamicSnapshotType
+	if snapshot.Spec.Source.VolumeSnapshotContentName != nil {
+		snapshotProvisionType = metrics.PreProvisionedSnapshotType
+	}
+
+	// Start metrics operations
+	if !utils.IsSnapshotCreated(snapshot) {
+		// Only start CreateSnapshot operation if the snapshot has not been cut
+		ctrl.metricsManager.OperationStart(
+			metrics.NewOperationKey(metrics.CreateSnapshotOperationName, snapshot.UID),
+			metrics.NewOperationValue(driverName, snapshotProvisionType),
+		)
+	}
+	ctrl.metricsManager.OperationStart(
+		metrics.NewOperationKey(metrics.CreateSnapshotAndReadyOperationName, snapshot.UID),
+		metrics.NewOperationValue(driverName, snapshotProvisionType),
+	)
 
 	// Pre-provisioned snapshot
 	if snapshot.Spec.Source.VolumeSnapshotContentName != nil {
@@ -403,13 +442,16 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
 		if err != nil {
 			return err
 		}
+
 		// if no content found yet, update status and return
 		if content == nil {
 			// can not find the desired VolumeSnapshotContent from cache store
 			ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotContentMissing", "VolumeSnapshotContent is missing")
 			klog.V(4).Infof("syncUnreadySnapshot[%s]: snapshot content %q requested but not found, will try again", utils.SnapshotKey(snapshot), *snapshot.Spec.Source.VolumeSnapshotContentName)
+
 			return fmt.Errorf("snapshot %s requests an non-existing content %s", utils.SnapshotKey(snapshot), *snapshot.Spec.Source.VolumeSnapshotContentName)
 		}
+
 		// Set VolumeSnapshotRef UID
 		newContent, err := ctrl.checkandBindSnapshotContent(snapshot, content)
 		if err != nil {
@@ -426,8 +468,10 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
 			ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotStatusUpdateFailed", fmt.Sprintf("Snapshot status update failed, %v", err))
 			return err
 		}
+
 		return nil
 	}
+
 	// snapshot.Spec.Source.VolumeSnapshotContentName == nil - dynamically creating snapshot
 	klog.V(5).Infof("getDynamicallyProvisionedContentFromStore for snapshot %s", uniqueSnapshotName)
 	contentObj, err := ctrl.getDynamicallyProvisionedContentFromStore(snapshot)
@@ -1094,10 +1138,30 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
 	if updated {
 		snapshotClone := snapshotObj.DeepCopy()
 		snapshotClone.Status = newStatus
+
+		// We need to record metrics before updating the status due to a bug causing cache entries after a failed UpdateStatus call.
+		// Must meet the following criteria to emit a successful CreateSnapshot status
+		// 1. Previous status was nil OR Previous status had a nil CreationTime
+		// 2. New status must be non-nil with a non-nil CreationTime
+		driverName := content.Spec.Driver
+		createOperationKey := metrics.NewOperationKey(metrics.CreateSnapshotOperationName, snapshot.UID)
+		if !utils.IsSnapshotCreated(snapshotObj) && utils.IsSnapshotCreated(snapshotClone) {
+			ctrl.metricsManager.RecordMetrics(createOperationKey, metrics.NewSnapshotOperationStatus(metrics.SnapshotStatusTypeSuccess), driverName)
+		}
+
+		// Must meet the following criteria to emit a successful CreateSnapshotAndReady status
+		// 1. Previous status was nil OR Previous status had a nil ReadyToUse OR Previous status had a false ReadyToUse
+		// 2. New status must be non-nil with a ReadyToUse as true
+		if !utils.IsSnapshotReady(snapshotObj) && utils.IsSnapshotReady(snapshotClone) {
+			createAndReadyOperation := metrics.NewOperationKey(metrics.CreateSnapshotAndReadyOperationName, snapshot.UID)
+			ctrl.metricsManager.RecordMetrics(createAndReadyOperation, metrics.NewSnapshotOperationStatus(metrics.SnapshotStatusTypeSuccess), driverName)
+		}
+
 		newSnapshotObj, err := ctrl.clientset.SnapshotV1beta1().VolumeSnapshots(snapshotClone.Namespace).UpdateStatus(context.TODO(), snapshotClone, metav1.UpdateOptions{})
 		if err != nil {
 			return nil, newControllerUpdateError(utils.SnapshotKey(snapshot), err.Error())
 		}
+
 		return newSnapshotObj, nil
 	}
 
@@ -1175,6 +1239,45 @@ func (ctrl *csiSnapshotCommonController) getSnapshotClass(className string) (*cr
 	}
 
 	return class, nil
+}
+
+// getSnapshotDriverName is a helper function to get snapshot driver from the VolumeSnapshot.
+// We try to get the driverName in multiple ways, as snapshot controller metrics depend on the correct driverName.
+func (ctrl *csiSnapshotCommonController) getSnapshotDriverName(vs *crdv1.VolumeSnapshot) (string, error) {
+	klog.V(5).Infof("getSnapshotDriverName: VolumeSnapshot[%s]", vs.Name)
+	var driverName string
+
+	// Pre-Provisioned snapshots have contentName as source
+	var contentName string
+	if vs.Spec.Source.VolumeSnapshotContentName != nil {
+		contentName = *vs.Spec.Source.VolumeSnapshotContentName
+	}
+
+	// Get Driver name from SnapshotContent if we found a contentName
+	if contentName != "" {
+		content, err := ctrl.contentLister.Get(contentName)
+		if err != nil {
+			klog.Errorf("getSnapshotDriverName: failed to get snapshotContent: %v", contentName)
+		} else {
+			driverName = content.Spec.Driver
+		}
+
+		if driverName != "" {
+			return driverName, nil
+		}
+	}
+
+	// Dynamic snapshots will have a snapshotclass with a driver
+	if vs.Spec.VolumeSnapshotClassName != nil {
+		class, err := ctrl.getSnapshotClass(*vs.Spec.VolumeSnapshotClassName)
+		if err != nil {
+			klog.Errorf("getSnapshotDriverName: failed to get snapshotClass: %v", *vs.Spec.VolumeSnapshotClassName)
+		} else {
+			driverName = class.Driver
+		}
+	}
+
+	return driverName, nil
 }
 
 // SetDefaultSnapshotClass is a helper function to figure out the default snapshot class.
--- a/pkg/common-controller/snapshot_controller_base.go
+++ b/pkg/common-controller/snapshot_controller_base.go
@@ -24,6 +24,7 @@ import (
 	clientset "github.com/kubernetes-csi/external-snapshotter/client/v3/clientset/versioned"
 	storageinformers "github.com/kubernetes-csi/external-snapshotter/client/v3/informers/externalversions/volumesnapshot/v1beta1"
 	storagelisters "github.com/kubernetes-csi/external-snapshotter/client/v3/listers/volumesnapshot/v1beta1"
+	"github.com/kubernetes-csi/external-snapshotter/v3/pkg/metrics"
 	"github.com/kubernetes-csi/external-snapshotter/v3/pkg/utils"
 
 	v1 "k8s.io/api/core/v1"
@@ -60,6 +61,8 @@ type csiSnapshotCommonController struct {
 	snapshotStore cache.Store
 	contentStore  cache.Store
 
+	metricsManager metrics.MetricsManager
+
 	resyncPeriod time.Duration
 }
 
@@ -71,6 +74,7 @@ func NewCSISnapshotCommonController(
 	volumeSnapshotContentInformer storageinformers.VolumeSnapshotContentInformer,
 	volumeSnapshotClassInformer storageinformers.VolumeSnapshotClassInformer,
 	pvcInformer coreinformers.PersistentVolumeClaimInformer,
+	metricsManager metrics.MetricsManager,
 	resyncPeriod time.Duration,
 ) *csiSnapshotCommonController {
 	broadcaster := record.NewBroadcaster()
@@ -80,14 +84,15 @@ func NewCSISnapshotCommonController(
 	eventRecorder = broadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: fmt.Sprintf("snapshot-controller")})
 
 	ctrl := &csiSnapshotCommonController{
-		clientset:     clientset,
-		client:        client,
-		eventRecorder: eventRecorder,
-		resyncPeriod:  resyncPeriod,
-		snapshotStore: cache.NewStore(cache.DeletionHandlingMetaNamespaceKeyFunc),
-		contentStore:  cache.NewStore(cache.DeletionHandlingMetaNamespaceKeyFunc),
-		snapshotQueue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "snapshot-controller-snapshot"),
-		contentQueue:  workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "snapshot-controller-content"),
+		clientset:      clientset,
+		client:         client,
+		eventRecorder:  eventRecorder,
+		resyncPeriod:   resyncPeriod,
+		snapshotStore:  cache.NewStore(cache.DeletionHandlingMetaNamespaceKeyFunc),
+		contentStore:   cache.NewStore(cache.DeletionHandlingMetaNamespaceKeyFunc),
+		snapshotQueue:  workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "snapshot-controller-snapshot"),
+		contentQueue:   workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "snapshot-controller-content"),
+		metricsManager: metricsManager,
 	}
 
 	ctrl.pvcLister = pvcInformer.Lister()
@@ -363,6 +368,7 @@ func (ctrl *csiSnapshotCommonController) updateSnapshot(snapshot *crdv1.VolumeSn
 	if !newSnapshot {
 		return nil
 	}
+
 	err = ctrl.syncSnapshot(snapshot)
 	if err != nil {
 		if errors.IsConflict(err) {
@@ -407,6 +413,13 @@ func (ctrl *csiSnapshotCommonController) updateContent(content *crdv1.VolumeSnap
 func (ctrl *csiSnapshotCommonController) deleteSnapshot(snapshot *crdv1.VolumeSnapshot) {
 	_ = ctrl.snapshotStore.Delete(snapshot)
 	klog.V(4).Infof("snapshot %q deleted", utils.SnapshotKey(snapshot))
+	driverName, err := ctrl.getSnapshotDriverName(snapshot)
+	if err != nil {
+		klog.Errorf("failed to getSnapshotDriverName while recording metrics for snapshot %q: %s", utils.SnapshotKey(snapshot), err)
+	} else {
+		deleteOperationKey := metrics.NewOperationKey(metrics.DeleteSnapshotOperationName, snapshot.UID)
+		ctrl.metricsManager.RecordMetrics(deleteOperationKey, metrics.NewSnapshotOperationStatus(metrics.SnapshotStatusTypeSuccess), driverName)
+	}
 
 	snapshotContentName := ""
 	if snapshot.Status != nil && snapshot.Status.BoundVolumeSnapshotContentName != nil {
@@ -416,6 +429,7 @@ func (ctrl *csiSnapshotCommonController) deleteSnapshot(snapshot *crdv1.VolumeSn
 		klog.V(5).Infof("deleteSnapshot[%q]: content not bound", utils.SnapshotKey(snapshot))
 		return
 	}
+
 	// sync the content when its snapshot is deleted.  Explicitly sync'ing the
 	// content here in response to snapshot deletion prevents the content from
 	// waiting until the next sync period for its Release.
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -30,13 +30,48 @@ import (
 )
 
 const (
-	opStatusUnknown      = "Unknown"
-	labelDriverName      = "driver_name"
-	labelOperationName   = "operation_name"
-	labelOperationStatus = "operation_status"
-	subSystem            = "snapshot_controller"
-	metricName           = "operation_total_seconds"
-	metricHelpMsg        = "Total number of seconds spent by the controller on an operation from end to end"
+	labelDriverName               = "driver_name"
+	labelOperationName            = "operation_name"
+	labelOperationStatus          = "operation_status"
+	labelSnapshotType             = "snapshot_type"
+	subSystem                     = "snapshot_controller"
+	operationLatencyMetricName    = "operation_total_seconds"
+	operationLatencyMetricHelpMsg = "Total number of seconds spent by the controller on an operation"
+	unknownDriverName             = "unknown"
+
+	// CreateSnapshotOperationName is the operation that tracks how long the controller takes to create a snapshot.
+	// Specifically, the operation metric is emitted based on the following timestamps:
+	// - Start_time: controller notices the first time that there is a new VolumeSnapshot CR to dynamically provision a snapshot
+	// - End_time:   controller notices that the CR has a status with CreationTime field set to be non-nil
+	CreateSnapshotOperationName = "CreateSnapshot"
+
+	// CreateSnapshotAndReadyOperationName is the operation that tracks how long the controller takes to create a snapshot and for it to be ready.
+	// Specifically, the operation metric is emitted based on the following timestamps:
+	// - Start_time: controller notices the first time that there is a new VolumeSnapshot CR(both dynamic and pre-provisioned cases)
+	// - End_time:   controller notices that the CR has a status with Ready field set to be true
+	CreateSnapshotAndReadyOperationName = "CreateSnapshotAndReady"
+
+	// DeleteSnapshotOperationName is the operation that tracks how long a snapshot deletion takes.
+	// Specifically, the operation metric is emitted based on the following timestamps:
+	// - Start_time: controller notices the first time that there is a deletion timestamp placed on the VolumeSnapshot CR and the CR is ready to be deleted. Note that if the CR is being used by a PVC for rehydration, the controller should *NOT* set the start_time.
+	// - End_time: controller removed all finalizers on the VolumeSnapshot CR such that the CR is ready to be removed in the API server.
+	DeleteSnapshotOperationName = "DeleteSnapshot"
+
+	// DynamicSnapshotType represents a snapshot that is being dynamically provisioned
+	DynamicSnapshotType = snapshotProvisionType("dynamic")
+	// PreProvisionedSnapshotType represents a snapshot that is pre-provisioned
+	PreProvisionedSnapshotType = snapshotProvisionType("pre-provisioned")
+
+	// SnapshotStatusTypeUnknown represents that the status is unknown
+	SnapshotStatusTypeUnknown snapshotStatusType = "unknown"
+	// Success and Cancel are statuses for operation time (operation_total_seconds) as seen by snapshot controller
+	// SnapshotStatusTypeSuccess represents that a CreateSnapshot, CreateSnapshotAndReady,
+	// or DeleteSnapshot has finished successfully.
+	// Individual reconciliations (reconciliation_total_seconds) also use this status.
+	SnapshotStatusTypeSuccess snapshotStatusType = "success"
+	// SnapshotStatusTypeCancel represents that a CreateSnapshot, CreateSnapshotAndReady,
+	// or DeleteSnapshot has been deleted before finishing.
+	SnapshotStatusTypeCancel snapshotStatusType = "cancel"
 )
 
 // OperationStatus is the interface type for representing an operation's execution
@@ -57,11 +92,11 @@ type MetricsManager interface {
 
 	// OperationStart takes in an operation and caches its start time.
 	// if the operation already exists, it's an no-op.
-	OperationStart(op Operation)
+	OperationStart(key OperationKey, val OperationValue)
 
 	// DropOperation removes an operation from cache.
 	// if the operation does not exist, it's an no-op.
-	DropOperation(op Operation)
+	DropOperation(op OperationKey)
 
 	// RecordMetrics records a metric point. Note that it will be an no-op if an
 	// operation has NOT been marked "Started" previously via invoking "OperationStart".
@@ -69,18 +104,47 @@ type MetricsManager interface {
 	// op - the operation which the metric is associated with.
 	// status - the operation status, if not specified, i.e., status == nil, an
 	//          "Unknown" status of the passed-in operation is assumed.
-	RecordMetrics(op Operation, status OperationStatus)
+	RecordMetrics(op OperationKey, status OperationStatus, driverName string)
 }
 
-// Operation is a structure which holds information to identify a snapshot
-// related operation
-type Operation struct {
-	// the name of the operation, for example: "CreateSnapshot", "DeleteSnapshot"
+// OperationKey is a structure which holds information to
+// uniquely identify a snapshot related operation
+type OperationKey struct {
+	// Name is the name of the operation, for example: "CreateSnapshot", "DeleteSnapshot"
 	Name string
-	// the name of the driver which executes the operation
-	Driver string
-	// the resource UID to which the operation has been executed against
+	// ResourceID is the resource UID to which the operation has been executed against
 	ResourceID types.UID
+}
+
+// OperationValue is a structure which holds operation metadata
+type OperationValue struct {
+	// Driver is the driver name which executes the operation
+	Driver string
+	// SnapshotType represents the snapshot type, for example: "dynamic", "pre-provisioned"
+	SnapshotType string
+
+	// startTime is the time when the operation first started
+	startTime time.Time
+}
+
+// NewOperationKey initializes a new OperationKey
+func NewOperationKey(name string, snapshotUID types.UID) OperationKey {
+	return OperationKey{
+		Name:       name,
+		ResourceID: snapshotUID,
+	}
+}
+
+// NewOperationValue initializes a new OperationValue
+func NewOperationValue(driver string, snapshotType snapshotProvisionType) OperationValue {
+	if driver == "" {
+		driver = unknownDriverName
+	}
+
+	return OperationValue{
+		Driver:       driver,
+		SnapshotType: string(snapshotType),
+	}
 }
 
 type operationMetricsManager struct {
@@ -93,10 +157,11 @@ type operationMetricsManager struct {
 	// registry is a wrapper around Prometheus Registry
 	registry k8smetrics.KubeRegistry
 
-	// opLatencyMetrics is a Histogram metrics
+	// opLatencyMetrics is a Histogram metrics for operation time per request
 	opLatencyMetrics *k8smetrics.HistogramVec
 }
 
+// NewMetricsManager creates a new MetricsManager instance
 func NewMetricsManager() MetricsManager {
 	mgr := &operationMetricsManager{
 		cache: sync.Map{},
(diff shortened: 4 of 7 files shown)


cmd/snapshot-controller/main.go
```diff
@@ -87,6 +92,25 @@ func main() {
 	factory := informers.NewSharedInformerFactory(snapClient, *resyncPeriod)
 	coreFactory := coreinformers.NewSharedInformerFactory(kubeClient, *resyncPeriod)
 
+	// Create and register metrics manager
+	metricsManager := metrics.NewMetricsManager()
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	if *metricsAddress != "" {
+		srv, err := metricsManager.StartMetricsEndpoint(*metricsPath, *metricsAddress, nil, wg)
```
ggriffiths, line 100: I wasn't sure how the logger was intended to be passed down to the StartMetricsEndpoint? Don't we just use `klog`? Wasn't sure how to pass an instance of klog to the metrics pkg.

yuxiangqian:
/cc. @msau42 @mattcary @xing-yang

xing-yang:
Do you need a rebase?  It seems that the unit test is hitting this bug: https://github.com/kubernetes-csi/external-snapshotter/pull/381?

ggriffiths:
pkg/metric tests seem to be flakey now. Looking into it now.

cmd/snapshot-controller/main.go
```diff
@@ -87,6 +92,25 @@ func main() {
...
+		if err != nil {
+			klog.Errorf("Failed to start metrics: %s", err.Error())
+			os.Exit(1)
+		}
+		defer func() {
+			err := srv.Close()
+			if err != nil {
+				klog.Errorf("fFailed to shutdown metrics server: %s", err.Error())
```
xing-yang, line 108: typo: fFailed -> Failed

pkg/common-controller/snapshot_controller.go
```diff
@@ -416,6 +428,7 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
 		if err != nil {
 			// snapshot is bound but content is not bound to snapshot correctly
 			ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotBindFailed", fmt.Sprintf("Snapshot failed to bind VolumeSnapshotContent, %v", err))
+			ctrl.RecordSnapshotMetrics(snapshot, createSnapshotOperationName, preProvisionedSnapshotType, err)
```
xing-yang, line 431: What if it returns at line 424?  Do we need to record metrics with err there?

pkg/common-controller/snapshot_controller.go
```diff
@@ -455,11 +479,14 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
 	// If we reach here, it is a dynamically provisioned snapshot, and the volumeSnapshotContent object is not yet created.
 	if snapshot.Spec.Source.PersistentVolumeClaimName == nil {
 		ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotPVCSourceMissing", fmt.Sprintf("PVC source for snapshot %s is missing", uniqueSnapshotName))
-		return fmt.Errorf("expected PVC source for snapshot %s but got nil", uniqueSnapshotName)
+		err := fmt.Errorf("expected PVC source for snapshot %s but got nil", uniqueSnapshotName)
```
xing-yang, line 482: Should we record something when it return nil at line 476?

pkg/common-controller/snapshot_controller.go
```diff
@@ -425,27 +438,38 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
...
+			ctrl.RecordSnapshotMetrics(snapshot, createSnapshotOperationName, preProvisionedSnapshotType, err)
 			return err
 		}
+
+		ctrl.RecordSnapshotMetrics(snapshot, createSnapshotOperationName, preProvisionedSnapshotType, nil)
 		return nil
 	}
+
 	// snapshot.Spec.Source.VolumeSnapshotContentName == nil - dynamically creating snapshot
+	ctrl.StartMetricsOperation(snapshot, snapshottingOperationName, dynamicSnapshotType)
```
xing-yang, line 450: Do we also need to record error for snapshottingOperationName?  It looks like this only records a completion?

yuxiangqian, line 450: I am not sure whether the timer should be started here.
line 451 to 477 is really to handle the situations when the controller crashed/restarted after a content has been created however binding has not been successfully executed.
Or is it your desire to only recode error counts? then at line 475 it should clean the operation, otherwise it would be memory leak?

pkg/common-controller/snapshot_controller.go
```diff
@@ -1099,6 +1127,19 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
 		if err != nil {
 			return nil, newControllerUpdateError(utils.SnapshotKey(snapshot), err.Error())
 		}
+
+		// If this is the first time we're setting creation time, we know snapshotting has finished
+		if snapshotObj.Status != nil && snapshotObj.Status.CreationTime == nil &&
```
yuxiangqian, line 1132: snapshotObj.Status can be nil
also, can use newStatus directly(it should not be nil) instead of checking snapshotClone.Status != nil

pkg/common-controller/snapshot_controller.go
```diff
@@ -1186,6 +1227,22 @@ func (ctrl *csiSnapshotCommonController) getSnapshotClass(className string) (*cr
 	return class, nil
 }
 
+// getSnapshotClass is a helper function to get snapshot class from the class name.
```
yuxiangqian, line 1230: comment does not match

pkg/common-controller/snapshot_metrics.go
```diff
@@ -0,0 +1,77 @@
+package common_controller
```
yuxiangqian, line 1: I am not convinced we need to have this wrapper. OperationStatus definitely needs an implementation, however, I am not sure about the "StartMetricsOperation" and "RecordSnapshotMetrics" calls.  Those are public methods and not called in anywhere else.

pkg/common-controller/snapshot_metrics.go
```diff
@@ -0,0 +1,77 @@
...
+			Error:  err.Error(),
+		}
+	}
+
+	return SnapshotOperationStatus{
+		Status: "success",
+	}
+}
+
+func (sos SnapshotOperationStatus) String() string {
```
yuxiangqian, line 44: this implementation will likely create a LOT of time series of latency metrics (one for each error message) which is problematic. see:
https://prometheus.io/docs/practices/naming/#labels

the status string is a label of the latency metrics, we definitely need to limit the cardinality 
https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/metrics/metrics.go#L134

pkg/metrics/metrics.go
```diff
@@ -95,6 +100,9 @@ type operationMetricsManager struct {
 
 	// opLatencyMetrics is a Histogram metrics
 	opLatencyMetrics *k8smetrics.HistogramVec
+
+	// opCountMetrics is a Counter metrics
+	opCountMetrics *k8smetrics.CounterVec
```
yuxiangqian, line 105: do we really need this? I thought HistogramVec already provides count? given that the metrics are labeled using status, a count can be easily achieved by doing a label = "success"/"error"?

pkg/metrics/metrics_test.go
```diff
@@ -165,7 +166,7 @@ snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operatio
 snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="300"} 1
 snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="600"} 1
 snapshot_controller_operation_total_seconds_bucket{driver_name="driver",operation_name="should-have-been-added",operation_status="Success",le="+Inf"} 1
-snapshot_controller_operation_total_seconds_sum{driver_name="driver",operation_name="should-have-been-added",operation_status="Success"} 0.3
+snapshot_controller_operation_total_seconds_sum{driver_name="driver",operation_name="should-have-been-added",operation_status="Success"} 0.1
```
yuxiangqian, line 169: curious why this is adjusted to lower?

yuxiangqian (commented):
Good job @ggriffiths

pkg/metrics/metrics_test.go
ggriffiths: Typo, will fix this.

pkg/metrics/metrics.go
ggriffiths: Good point, removed the counter and added status to the histogram.

pkg/common-controller/snapshot_metrics.go
ggriffiths: Removed these two public wrapper functions

pkg/common-controller/snapshot_controller.go
```diff
@@ -237,6 +237,9 @@ func (ctrl *csiSnapshotCommonController) syncSnapshot(snapshot *crdv1.VolumeSnap
 func (ctrl *csiSnapshotCommonController) processSnapshotWithDeletionTimestamp(snapshot *crdv1.VolumeSnapshot) error {
 	klog.V(5).Infof("processSnapshotWithDeletionTimestamp VolumeSnapshot[%s]: %s", utils.SnapshotKey(snapshot), utils.GetSnapshotStatusForLogging(snapshot))
 
+	// New snapshot CR, start operation metric
+	ctrl.StartMetricsOperation(snapshot, deleteSnapshotOperationName, "")
```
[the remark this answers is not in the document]
ggriffiths: Added type for deletion metrics

cmd/snapshot-controller/main.go
```diff
@@ -110,6 +135,7 @@ func main() {
 		go ctrl.Run(*threads, stopCh)
 
 		// ...until SIGINT
+		defer wg.Done()
```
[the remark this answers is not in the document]
ggriffiths: Moved to the defer func

pkg/common-controller/snapshot_controller.go
ggriffiths: I couldn't find any error cases for `snapshottingOperationName` that didn't overlap with createSnapshot. Will take another look.

ggriffiths:
Addressed most feedback @xing-yang @yuxiangqian ,

Pending work I'll do today:
- Rework status field to classify different error types, instead of exact error messages.
- Check for places to emit snapshotting operation failures

pkg/common-controller/snapshot_metrics.go
ggriffiths: Added a type `snapshotStatusType` with four types as per the design doc:
1. Success
2. Invalid request
3. Controller error
4. Storage system error

However, I couldn't find any examples of storage-system-error? Wouldn't that mainly be in the snapshotter sidecar?

msau42: For storage system error, are we able to get csi error code?

pkg/common-controller/snapshot_controller.go
```diff
@@ -252,6 +269,12 @@ func (ctrl *csiSnapshotCommonController) processSnapshotWithDeletionTimestamp(sn
 	// content has been found from content cache store
 	content, err := ctrl.getContentFromStore(contentName)
 	if err != nil {
+		ctrl.metricsManager.RecordMetrics(metrics.Operation{
```
msau42, line 272: In general, is it possible to have the `RecordMetrics` call as a defer right after Start? That way we don't need to remember to call RecordMetrics if we add more exit points in the future.

pkg/common-controller/snapshot_controller.go
```diff
@@ -236,6 +237,22 @@ func (ctrl *csiSnapshotCommonController) syncSnapshot(snapshot *crdv1.VolumeSnap
...
+	if err != nil {
+		klog.Errorf("failed to getSnapshotDriverName while recording metrics for snapshot %q: %s", utils.SnapshotKey(snapshot), err)
+	}
+
+	// Processing delete, start operation metric
+	snapshotProvisionType := dynamicSnapshotType
+	if snapshot.Spec.Source.VolumeSnapshotContentName != nil {
+		snapshotProvisionType = preProvisionedSnapshotType
+	}
+	ctrl.metricsManager.OperationStart(metrics.Operation{
```
msau42, line 250: Can we save the `metrics.Operation` struct in a variable and reuse it, so that we don't accidentally call `RecordMetrics` with the wrong parameters?

pkg/common-controller/snapshot_metrics.go
```diff
@@ -0,0 +1,37 @@
+package common_controller
+
+const (
+	createSnapshotOperationName = "CreateSnapshot"
+	snapshottingOperationName   = "Snapshotting"
```
msau42, line 5: What is the difference between snapshotting and createsnapshot?

pkg/common-controller/snapshot_controller.go
```diff
@@ -252,6 +269,12 @@ func (ctrl *csiSnapshotCommonController) processSnapshotWithDeletionTimestamp(sn
 	// content has been found from content cache store
 	content, err := ctrl.getContentFromStore(contentName)
 	if err != nil {
+		ctrl.metricsManager.RecordMetrics(metrics.Operation{
+			Name:         deleteSnapshotOperationName,
+			Driver:       driverName,
+			ResourceID:   snapshot.UID,
+			SnapshotType: string(snapshotProvisionType),
+		}, NewSnapshotOperationStatus(snapshotCodeControllerError))
```
msau42, line 277: Recording a metric completion with an error status is mixing the semantic of what this metric is supposed to measure, which is one operation end to end. Instead, what this logic is doing is recording a metric per sync loop, so for a single volume snapshot request, it's could potentially record multiple operations in the histogram for it.

If this is what we want, then we should change the description of the metric to reflect that and look into end to end metrics separately.

For end to end metrics, I'm wondering if actually all we need to do is compare with the object creation/deletion timestamps.

pkg/common-controller/snapshot_controller.go
ggriffiths: I think we could have some sort defer call to perform RecordMetrics, but we would still need to update the snapshot type before the defer is executed. We could initialize a snapshotStatusType and change it before the defer is executed, but we'd still need to remember to update that.

I think a solution could be to have a custom error type that implements in the error interface and requires a snapshotStatusType assigned along with it. This would allow us to emit the correct snapshotStatusType based on the error returned per function. That might be a little involved for the 1.20 code freeze though.

pkg/common-controller/snapshot_metrics.go
ggriffiths: CreateSnapshot is full end to end from VolumeSnapshot CR noticed until it's marked ready.
Snapshotting is defined as when the VolumeSnapshot CR is noticed until a VolumeSnapshotContent is created.

From the design doc:

Snapshotting:
* Start_time: controller notices the first time that there is a new VolumeSnapshot CR to dynamically provision a snapshot
* End_time: controller notices that the CR has a status with CreationTime field set to be non-nil

CreateSnapshot
* Start_time: controller notices the first time that there is a new VolumeSnapshot CR(both dynamic and pre-provisioned cases)
* End_time: controller notices that the CR has a status with Ready field set to be true

@yuxiangqian is it worth only having CreateSnapshot and DeleteSnapshot operations? What sort of use cases are there for needing the snapshotting operation?

pkg/common-controller/snapshot_controller.go
ggriffiths: Good idea, will make this change.

pkg/common-controller/snapshot_controller.go
ggriffiths: The goal of this metric is the following:
>Snapshot operation end to end duration in number of seconds. For example, the duration from a user creates a VolumeSnapshot to request taking a snapshot of a particular PVC to the VolumeSnapshot has a Status.Ready == true.

I like the idea of comparing `VolumeSnapshot.CreationTimestamp` with `VolumeSnapshot.Status.CreationTime` for the e2e operation length. 

We could have two metrics, one counter w/ status, and one e2e with the operation time. This is what I originally implemented. I'll let @xing-yang @yuxiangqian chime in on if just having the histogram is enough. I saved the counter metric code and easily can re-add it if needed.

pkg/common-controller/snapshot_metrics.go
ggriffiths: We could check the VolumeSnapshot.Status.Error, but it seems like that error is used for both storage-system and controller errors. It formats the message as a string, I believe losing the csi error code.
https://github.com/kubernetes-csi/external-snapshotter/blob/23b415b6aaa7e0eba402bc3984ae2b726b101a80/pkg/sidecar-controller/snapshot_controller.go#L141-L156

So I'm not sure we can get the csi error code from that error, unless you are talking about checking in another spot.

pkg/common-controller/snapshot_controller.go
msau42: Is it possible to start the timer after we have figured out if it's dynamic or preprovisioned? We should already have the api object with the source that tells us this.

I think it's fine to have a field for status that needs to be updated before returning. What about initializing it to error, and only setting it to success in the one case?

pkg/common-controller/snapshot_metrics.go
msau42: My main ask here is to make the names more obvious what they're measuring. What about something like: "CreateSnapshot" and "CreateSnapshotAndReady"?

pkg/common-controller/snapshot_controller.go
msau42: I like to think of it as two different things to measure, and each should have its own set of metrics.

1) Total end to end time of an request. This covers the time from an end user perspective. A controller could retry multiple times, but from the user this is only one request. Intermediate errors are not a concern here.
2) Time of a controller reconciliation loop. This covers just one loop and is more for internal tracking of how efficient this controller is, and can offer some data on intermittent/intermediate errors.

The metric for each scenario could be a histogram that covers status and latency together. There is latency and status of a single control loop, and there is latency and status of a total end to end request (although the status is just success/cancel)

pkg/common-controller/snapshot_controller.go
```diff
@@ -546,6 +622,7 @@ func (ctrl *csiSnapshotCommonController) getDynamicallyProvisionedContentFromSto
 	if content.Spec.Source.VolumeHandle == nil {
```
xing-yang, line 622: Should we record a snapshotStatusTypeInvalidRequest when it returns at line 619?

```
	if content == nil {
		// no matching content with the desired name has been found in cache
		return nil, nil
	}
```

pkg/common-controller/snapshot_controller.go
```diff
@@ -588,22 +666,34 @@ func (ctrl *csiSnapshotCommonController) getContentFromStore(contentName string)
...
+		Name:         createSnapshotOperationName,
+		Driver:       driverName,
+		ResourceID:   snapshot.UID,
+		SnapshotType: string(dynamicSnapshotType),
+	}
 
 	// If PVC is not being deleted and finalizer is not added yet, a finalizer should be added to PVC until snapshot is created
 	klog.V(5).Infof("createSnapshotContent: Check if PVC is not being deleted and add Finalizer for source of snapshot [%s] if needed", snapshot.Name)
-	err := ctrl.ensurePVCFinalizer(snapshot)
+	err = ctrl.ensurePVCFinalizer(snapshot)
```
xing-yang, line 682: Record a snapshotStatusTypeControllerError when err != nil?

pkg/common-controller/snapshot_controller.go
```diff
@@ -654,6 +744,7 @@ func (ctrl *csiSnapshotCommonController) createSnapshotContent(snapshot *crdv1.V
 		strerr := fmt.Sprintf("Error creating volume snapshot content object for snapshot %s: %v.", utils.SnapshotKey(snapshot), err)
 		klog.Error(strerr)
 		ctrl.eventRecorder.Event(snapshot, v1.EventTypeWarning, "CreateSnapshotContentFailed", strerr)
+		ctrl.metricsManager.RecordMetrics(createOperation, NewSnapshotOperationStatus(snapshotStatusTypeControllerError))
```
xing-yang, line 747: At line 701, record snapshotStatusTypeInvalidRequest?
```
	snapshotRef, err := ref.GetReference(scheme.Scheme, snapshot)
	if err != nil {
		return nil, err
	}
```

pkg/common-controller/snapshot_controller.go
```diff
@@ -663,6 +754,7 @@ func (ctrl *csiSnapshotCommonController) createSnapshotContent(snapshot *crdv1.V
 	// Update content in the cache store
 	_, err = ctrl.storeContentUpdate(updateContent)
 	if err != nil {
+		ctrl.metricsManager.RecordMetrics(createOperation, NewSnapshotOperationStatus(snapshotStatusTypeControllerError))
```
xing-yang, line 757: I wonder if we need to record an error here.  We don't even return err.

pkg/common-controller/snapshot_controller.go
```diff
@@ -1093,12 +1185,46 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
...
+
 		snapshotClone := snapshotObj.DeepCopy()
 		snapshotClone.Status = newStatus
 		newSnapshotObj, err := ctrl.clientset.SnapshotV1beta1().VolumeSnapshots(snapshotClone.Namespace).UpdateStatus(context.TODO(), snapshotClone, metav1.UpdateOptions{})
 		if err != nil {
 			return nil, newControllerUpdateError(utils.SnapshotKey(snapshot), err.Error())
 		}
+
+		// Must meet the following criteria to emit a successful snapshotting status
+		// 1. Previous status was nil OR Previous status had a nil
```
xing-yang, line 1213: Previous status had a nil CreationTime

pkg/common-controller/snapshot_metrics.go
ggriffiths: CreateSnapshotAndReady sounds good, that is more clear to me. @yuxiangqian does this sound good? 

In this case, we would actually rename Snapshotting to CreateSnapshot, and rename CreateSnapshot to CreateSnapshotAndReady:

CreateSnapshot:

* Start_time: controller notices the first time that there is a new VolumeSnapshot CR to dynamically provision a snapshot
* End_time: controller notices that the CR has a status with CreationTime field set to be non-nil

CreateSnapshotAndReady
* Start_time: controller notices the first time that there is a new VolumeSnapshot CR(both dynamic and pre-provisioned cases)
* End_time: controller notices that the CR has a status with Ready field set to be true

pkg/common-controller/snapshot_controller.go
ggriffiths: That sounds good to me. I'll change the PR soon to reflect the two different histograms.

pkg/common-controller/snapshot_metrics.go
xing-yang: Sounds good to me.

pkg/common-controller/snapshot_controller.go
ggriffiths: Refactored this. The repetitive code is mainly from all of the error metrics emitting, so I added defer calls that only emit metrics if `err != nil`. Success metrics are still manually emitted outside of the defer calls, this is because they are mostly deeper in the call stack.

I initialized snapshotStatusType to be `controller-error`, and manually changed to `invalid-request`/etc when needed.

pkg/common-controller/snapshot_metrics.go
ggriffiths: Changed the metric types to reflect the above.

pkg/common-controller/snapshot_controller.go
ggriffiths: Added return err and metric emit is handled by defer

pkg/common-controller/snapshot_controller.go
ggriffiths: I am not sure either. I thought this code block inside of syncUnreadySnapshot was the extent of pre-provisioned snapshot handling:
https://github.com/kubernetes-csi/external-snapshotter/pull/409/files#diff-1a3473f9016f7abcf5007414ff7141e39032aaf7aba079eeef262d04167208a2R423-R433

Is there another area where we handle this case?

ggriffiths:
Ready for re-review, addressed most concerns and had one question for @yuxiangqian about one of the comments.

pkg/common-controller/snapshot_controller.go
```diff
@@ -427,9 +458,25 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
 			ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotStatusUpdateFailed", fmt.Sprintf("Snapshot status update failed, %v", err))
 			return err
 		}
+
+		snapshotStatusType = metrics.SnapshotStatusTypeSuccess
 		return nil
 	}
-	// snapshot.Spec.Source.VolumeSnapshotContentName == nil - dynamically creating snapshot
```
xing-yang, line 432: Can you not to remove this comment?

pkg/common-controller/snapshot_controller.go
```diff
@@ -533,19 +584,36 @@ func (ctrl *csiSnapshotCommonController) getPreprovisionedContentFromStore(snaps
...
+
 	contentName := utils.GetDynamicSnapshotContentNameForSnapshot(snapshot)
 	content, err := ctrl.getContentFromStore(contentName)
 	if err != nil {
+		snapshotStatusType = metrics.SnapshotStatusTypeInvalidRequest
 		return nil, err
 	}
 	if content == nil {
 		// no matching content with the desired name has been found in cache
+		ctrl.metricsManager.RecordMetrics(createAndReadyOperation, metrics.NewSnapshotOperationStatus(metrics.SnapshotStatusTypeInvalidRequest), snapshot)
```
xing-yang, line 609: If you have `defer` above, you only need to set the snapshotStatusType?

pkg/common-controller/snapshot_controller.go
```diff
@@ -1093,12 +1179,36 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
...
+		createAndReadyOperation := metrics.NewOperation(metrics.CreateSnapshotAndReadyOperationName, driverName, snapshot)
+
 		snapshotClone := snapshotObj.DeepCopy()
 		snapshotClone.Status = newStatus
 		newSnapshotObj, err := ctrl.clientset.SnapshotV1beta1().VolumeSnapshots(snapshotClone.Namespace).UpdateStatus(context.TODO(), snapshotClone, metav1.UpdateOptions{})
 		if err != nil {
 			return nil, newControllerUpdateError(utils.SnapshotKey(snapshot), err.Error())
 		}
+
+		// Must meet the following criteria to emit a successful snapshotting status
```
xing-yang, line 1196: snapshotting -> CreateSnapshot

pkg/common-controller/snapshot_controller.go
```diff
@@ -1093,12 +1179,36 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
...
+		// 1. Previous status was nil OR Previous status had a nil CreationTime
+		// 2. New status must be non-nil with a non-nil CreationTime
+		if (snapshotObj.Status == nil || (snapshotObj.Status != nil && snapshotObj.Status.CreationTime == nil)) &&
+			(newStatus != nil && newStatus.CreationTime != nil) {
+			ctrl.metricsManager.RecordMetrics(createOperation, metrics.NewSnapshotOperationStatus(metrics.SnapshotStatusTypeSuccess), newSnapshotObj)
+		}
+
+		// Must meet the following criteria to emit a successful CreateSnapshot status
```
xing-yang, line 1204: CreateSnapshot -> CreateSnapshotAndReady

pkg/metrics/metrics.go
```diff
@@ -23,20 +23,38 @@ import (
...
+	labelSnapshotType                  = "snapshot_type"
+	subSystem                          = "snapshot_controller"
+	latencyReconciliationMetricName    = "reconciliation_total_seconds"
+	latencyReconciliationMetricHelpMsg = "Number of seconds spent by the controller on intermediate reconciliations"
+	latencyOperationMetricName         = "operation_total_seconds"
+	latencyOperationMetricHelpMsg      = "Total number of seconds spent by the controller on an operation"
+
+	CreateSnapshotOperationName         = "CreateSnapshot"
+	CreateSnapshotAndReadyOperationName = "CreateSnapshotAndReady"
+	DeleteSnapshotOperationName         = "DeleteSnapshot"
```
xing-yang, line 47: Can you add comments to explain these 3 operations

pkg/metrics/metrics.go
```diff
@@ -126,27 +168,71 @@ func (opMgr *operationMetricsManager) RecordMetrics(op Operation, status Operati
...
+
+	// record invidual reconciliation time
+	reconciliationDuration := time.Since(ts).Seconds()
+	opMgr.opReconciliationLatencyMetrics.WithLabelValues(op.Driver, op.Name, op.SnapshotType, strStatus).Observe(reconciliationDuration)
+
+	// record total operation durations if operation is successful
+	if strStatus == string(SnapshotStatusTypeSuccess) {
+		switch op.Name {
+		case CreateSnapshotAndReadyOperationName:
+			// time since user first created volumesnapshot
```
xing-yang, line 184: Add comments about "ready"

xing-yang:
I just have a few comments, otherwise it looks good overall.

@yuxiangqian @msau42 please take a look to see if you have more comments.

pkg/common-controller/snapshot_controller.go
ggriffiths: Oops, thought I added this when I added the metric. Just re-added.

pkg/common-controller/snapshot_controller.go
ggriffiths: The defer only handles cases where err != nil. Otherwise, the defer would create ControllerError metrics on every reconcile without an error (even though the operation might not be finished)

ggriffiths:
Addressed those @xing-yang. Also, I just remembered this requirement in the design doc:
> As a storage admin/provider, I want to the total size of all snapshots for capacity planning

I can get this value in as a label or new metric if we want (in this PR or a future one).

xing-yang:
> I just remembered this requirement in the design doc:
>> As a storage admin/provider, I want to the total size of all snapshots for capacity planning
> I can get this value in as a label or new metric if we want (in this PR or a future one).

I'm fine either way.  @yuxiangqian @msau42 what do you think?

pkg/common-controller/snapshot_controller.go
xing-yang: Got it.

xing-yang:
/lgtm

@yuxiangqian @msau42 can you take another look?

pkg/common-controller/snapshot_controller.go
```diff
@@ -236,6 +237,22 @@ func (ctrl *csiSnapshotCommonController) syncSnapshot(snapshot *crdv1.VolumeSnap
...
+	if err != nil {
+		klog.Errorf("failed to getSnapshotDriverName while recording metrics for snapshot %q: %v", utils.SnapshotKey(snapshot), err)
+	}
+
+	// Processing delete, start operation metric
+	deleteOperation := metrics.NewOperation(metrics.DeleteSnapshotOperationName, driverName, snapshot)
+	ctrl.metricsManager.OperationStart(deleteOperation)
+	snapshotStatusType := metrics.SnapshotStatusTypeControllerError
+	defer func() {
+		if err != nil {
```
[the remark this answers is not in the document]
ggriffiths: That's true. I'll add a `metrics.SnapshotStatusTypeInProgress`. This will be the default status, and we can update to ControllerError/etc as needed.

This way, we'll always record a metric, but won't have ControllerError by default.

pkg/metrics/metrics.go
```diff
@@ -23,20 +23,54 @@ import (
...
+	// - Start_time: controller notices the first time that there is a deletion timestamp placed on the VolumeSnapshot CR and the CR is ready to be deleted. Note that if the CR is being used by a PVC for rehydration, the controller should *NOT* set the start_time.
+	// - End_time: controller removed all finalizers on the VolumeSnapshot CR such that the CR is ready to be removed in the API server.
+	DeleteSnapshotOperationName = "DeleteSnapshot"
+
+	// DynamicSnapshotType represents a snapshot that is being dynamically provisioned
+	DynamicSnapshotType = snapshotProvisionType("dynamic")
+	// PreProvisionedSnapshotType represents a snapshot that is pre-provisioned
+	PreProvisionedSnapshotType = snapshotProvisionType("pre-provisioned")
+
+	SnapshotStatusTypeUnknown            = snapshotStatusType("unknown")
```
msau42, line 67: Can you add comments about which statuses are relevant to which metric?

pkg/metrics/metrics.go
```diff
@@ -126,27 +184,71 @@ func (opMgr *operationMetricsManager) RecordMetrics(op Operation, status Operati
...
-	opMgr.opLatencyMetrics.WithLabelValues(op.Driver, op.Name, strStatus).Observe(duration)
+
+	// record invidual reconciliation time
+	reconciliationDuration := time.Since(ts).Seconds()
+	opMgr.opReconciliationLatencyMetrics.WithLabelValues(op.Driver, op.Name, op.SnapshotType, strStatus).Observe(reconciliationDuration)
+
+	// record total operation durations if operation is successful
+	if strStatus == string(SnapshotStatusTypeSuccess) {
+		switch op.Name {
+		case CreateSnapshotAndReadyOperationName:
```
msau42, line 199: does tracking e2e time for CreateSnapshot also seem useful?

cmd/snapshot-controller/main.go
```diff
@@ -48,6 +50,9 @@ var (
 
 	leaderElection          = flag.Bool("leader-election", false, "Enables leader election.")
 	leaderElectionNamespace = flag.String("leader-election-namespace", "", "The namespace where the leader election resource exists. Defaults to the pod namespace if not set.")
+
+	metricsAddress = flag.String("metrics-address", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
```
msau42, line 54: We have a fix for the leader election library coming soon, and it's going to require a http endpoint for /healthz. Can we rename this field to be more generic like "http-endpoint"?

cc @verult who's working on the leader election fix

pkg/metrics/metrics.go
ggriffiths: good point, added this

pkg/metrics/metrics.go
```diff
@@ -23,20 +23,62 @@ import (
...
+	labelSnapshotType                  = "snapshot_type"
+	subSystem                          = "snapshot_controller"
+	latencyReconciliationMetricName    = "reconciliation_total_seconds"
+	latencyReconciliationMetricHelpMsg = "Number of seconds spent by the controller on intermediate reconciliations"
+	latencyOperationMetricName         = "operation_total_seconds"
+	latencyOperationMetricHelpMsg      = "Total number of seconds spent by the controller on an operation from end to end"
+
+	// CreateSnapshotOperationName is the operation that tracks how long the controller takes to create a snapshot.
+	// Specifically, the operation metric is emitted based on the following timestamps:
+	// - Start_time: controller notices the first time that there is a new VolumeSnapshot CR to dynamically provision a snapshot
```
msau42, line 46: The start time description is relevant to the e2e metric, but not the reconcilation one. Can we clarify these comments

pkg/metrics/metrics.go
```diff
@@ -23,20 +23,62 @@ import (
...
+	SnapshotStatusTypeControllerError    = snapshotStatusType("controller-error")
+	SnapshotStatusTypeStorageSystemError = snapshotStatusType("storage-system-error")
+
+	// Success and Cancel are statuses for e2e operation time (operation_total_seconds)
+	// SnapshotStatusTypeSuccess represents that a CreateSnapshot, CreateSnapshotAndReady,
+	// or DeleteSnapshot has finished successfully.
+	// Individual reconciliations (reconciliation_total_seconds) also use this status.
+	SnapshotStatusTypeSuccess = snapshotStatusType("success")
+	// SnapshotStatusTypeCancel represents that a CreateSnapshot, CreateSnapshotAndReady,
+	// or DeleteSnapshot has been deleted before finishing.
```
msau42, line 80: are we actually able to detect a canceled delete?

pkg/metrics/metrics.go
```diff
@@ -126,27 +192,76 @@ func (opMgr *operationMetricsManager) RecordMetrics(op Operation, status Operati
...
+			// check if this is a cancellation. It is a cancellation if the snapshot deletion finished without a status set
+			// or there is a status set with ReadyToUse set to false.
+			if snapshot.Status == nil || (snapshot.Status.ReadyToUse != nil && !(*snapshot.Status.ReadyToUse)) {
+				// get duration from creation time to delete/cancellation
+				cancelOperationDuration := time.Since(snapshot.CreationTimestamp.Time).Seconds()
+
+				// emit this metric as a custom cancellation
+				opMgr.opRequestLatencyMetrics.WithLabelValues(
+					op.Driver,
+					CreateSnapshotAndReadyOperationName,
```
msau42, line 227: Is there a cancel for CreateSnapshot?

msau42 (commented):
Thanks for working through my comments! I just have some minor ones remaining. 

@yuxiangqian please take another look

pkg/metrics/metrics.go
ggriffiths: The intention here was to send a cancel metric for the CreateSnapshotAndReady operation if a delete is issued prior to completion. 

In this case, we're sending a successful delete metric AND cancel CreateSnapshotAndReady metric.

pkg/metrics/metrics.go
ggriffiths: Added this

ggriffiths:
Thanks for the quick reviews! Addressed the recent comments.

pkg/metrics/metrics.go
msau42: not seeing it?

ggriffiths: Ah sorry, thought I did. Added now above

yuxiangqian:
@ggriffiths I was off for couple of days, will get back on this by EOD.

pkg/metrics/metrics.go
```diff
@@ -126,27 +192,91 @@ func (opMgr *operationMetricsManager) RecordMetrics(op Operation, status Operati
...
 	}
-	duration := time.Since(ts).Seconds()
-	opMgr.opLatencyMetrics.WithLabelValues(op.Driver, op.Name, strStatus).Observe(duration)
+
+	// record invidual reconciliation time
+	reconciliationDuration := time.Since(ts).Seconds()
+	opMgr.opReconciliationLatencyMetrics.WithLabelValues(op.Driver, op.Name, op.SnapshotType, strStatus).Observe(reconciliationDuration)
+
+	// record total operation durations if operation is successful
+	if strStatus == string(SnapshotStatusTypeSuccess) {
```
yuxiangqian, line 205: This switch case seemed to be a bit over complicated to me as each case would do the same piece of work to record an end to end latency metric, propose to make the following changes:
`if success`
`    duration = time.Since(cached.StartTime)`
`    opMgr.opRequestLatencyMetrics.With....()
`

pkg/metrics/metrics.go
```diff
@@ -126,27 +192,91 @@ func (opMgr *operationMetricsManager) RecordMetrics(op Operation, status Operati
...
+		switch op.Name {
+		case CreateSnapshotOperationName:
+			// time since user first created volumesnapshot until a volumesnapshot.Status.CreationTime is aded
+			operationDuration := time.Since(snapshot.CreationTimestamp.Time).Seconds()
```
yuxiangqian, line 209: it should be: `time.Since(cached-start-time)`. the creation timestamp of snapshot object does not mean the time when the controller sees it. In an extreme case, say when the snapshot controller has not been installed initiatively, the user creates a snapshot object, and x-hours later the snapshot controller got installed, we should *only* start counting when the controller gets to process the snapshot object.

pkg/metrics/metrics.go
```diff
@@ -126,27 +192,91 @@ func (opMgr *operationMetricsManager) RecordMetrics(op Operation, status Operati
...
+			opMgr.opRequestLatencyMetrics.WithLabelValues(op.Driver, op.Name, op.SnapshotType, strStatus).Observe(operationDuration)
+
+		case CreateSnapshotAndReadyOperationName:
+			// time since user first created volumesnapshot until ReadyToUse is true
+			operationDuration := time.Since(snapshot.CreationTimestamp.Time).Seconds()
```
yuxiangqian, line 214: ditto, should use cached time

pkg/metrics/metrics.go
```diff
@@ -126,27 +192,91 @@ func (opMgr *operationMetricsManager) RecordMetrics(op Operation, status Operati
...
+					op.Driver,
+					CreateSnapshotAndReadyOperationName,
+					op.SnapshotType,
+					string(SnapshotStatusTypeCancel),
+				).Observe(cancelOperationDuration)
+			}
+
+			// Still emit the delete operation, as we are both deleting the snapshot and cancelling it's creation.
+			// For this duration, we use the time since user marked volumesnapshot for deletion.
+			operationDuration := time.Since(snapshot.DeletionTimestamp.Time).Seconds()
```
yuxiangqian, line 250: ditto, should use cached timestamp

pkg/metrics/metrics.go
```diff
@@ -23,20 +23,62 @@ import (
...
+	// - Start_time: controller notices the first time that there is a deletion timestamp placed on the VolumeSnapshot CR and the CR is ready to be deleted. Note that if the CR is being used by a PVC for rehydration, the controller should *NOT* set the start_time.
+	// - End_time: controller removed all finalizers on the VolumeSnapshot CR such that the CR is ready to be removed in the API server.
+	DeleteSnapshotOperationName = "DeleteSnapshot"
+
+	// DynamicSnapshotType represents a snapshot that is being dynamically provisioned
+	DynamicSnapshotType = snapshotProvisionType("dynamic")
+	// PreProvisionedSnapshotType represents a snapshot that is pre-provisioned
+	PreProvisionedSnapshotType = snapshotProvisionType("pre-provisioned")
+
+	// The following metrics are used for the individual reconciliation metrics.
```
yuxiangqian, line 67: comment irrelevant?

pkg/metrics/metrics.go
```diff
@@ -126,27 +192,91 @@ func (opMgr *operationMetricsManager) RecordMetrics(op Operation, status Operati
...
+	strStatus := string(SnapshotStatusTypeUnknown)
 	if status != nil {
 		strStatus = status.String()
 	}
-	duration := time.Since(ts).Seconds()
-	opMgr.opLatencyMetrics.WithLabelValues(op.Driver, op.Name, strStatus).Observe(duration)
+
+	// record invidual reconciliation time
+	reconciliationDuration := time.Since(ts).Seconds()
+	opMgr.opReconciliationLatencyMetrics.WithLabelValues(op.Driver, op.Name, op.SnapshotType, strStatus).Observe(reconciliationDuration)
```
yuxiangqian, line 202: I am not sure this is the best place to implement reconciliation metrics.
the whole purpose of the cache is to reserve a starting point when there are multiple reconciliations. for a single reconciliation, we do not need any cached timestamp. If reconciliation latency is useful, then we should probably just do the duration calculation within controller logic. WDYT?

pkg/common-controller/snapshot_controller.go
```diff
@@ -425,27 +460,47 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
 			// update snapshot status failed
 			klog.V(4).Infof("failed to update snapshot %s status: %v", utils.SnapshotKey(snapshot), err)
 			ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotStatusUpdateFailed", fmt.Sprintf("Snapshot status update failed, %v", err))
+			snapshotStatusType = metrics.SnapshotStatusTypeControllerError
 			return err
 		}
+
+		snapshotStatusType = metrics.SnapshotStatusTypeSuccess
```
yuxiangqian, line 467: at this point, the VolumeSnapshot may or may not have "ready" set to true yet. This simply means the controller has successfully bound the volume snapshot to volume snapshot content. Sidecar will call CSI "ListSnapshots" to update  content status, and then common controller will resync in next reconciliation. setting status to success here effectively ends the e2e latency reporting which is wrong IMO.

pkg/metrics/metrics.go
msau42: I disagree here. I think the end to end metrics should reflect the user experience, which includes time spent in other components. If there's a bug in the snapshot controller such that it restarts frequently or is slow to process a snapshot request, we should capture that.

yuxiangqian: I agree slowness should be captured, but the point is not related here, it will be captured actually.
Crash-looping is a different case though. If controller is crash looping, the user should have been notified by other manner not the e2e metrics.
The point is that the controller is not installed at all when a snapshot CR is created, in this case, using the creation time stamp of a snapshot CR as the start time is not reasonable to me. it will create a weird spiky point in the time-series, and I suspect this is the experience a user would want.
setting up a meeting to discuss further tomorrow.

xing-yang: How is this kind of metric calculated for PVC creation?  Do we use creation timestamp of PVC as the the start time or do we use the time when the PV controller sees the PVC object?

msau42: I think there's value to have multiple metrics each capturing different levels of response, starting with the user experience and getting more fine grained to the controller, then the csi driver.

The higher level metrics can be used to define SLOs, whereas the finer grained metrics can be used to debug where something may be going wrong.

xing-yang: Sure.  I think we should keep this metric and modify the description to reflect what it measures.
As a followup, add another metric that calculates the time between the controller observed start time and readyToUse.

pkg/metrics/metrics.go
```diff
@@ -175,3 +305,25 @@ func (opMgr *operationMetricsManager) StartMetricsEndpoint(pattern, addr string,
...
+
+// snapshotProvisionType represents which kind of snapshot a metric is
+type snapshotProvisionType string
+
+// snapshotStatusType represents the type of snapshot status to report
+type snapshotStatusType string
+
+// SnapshotOperationStatus represents the status for a snapshot controller operation
+type SnapshotOperationStatus struct {
+	statusCode snapshotStatusType
```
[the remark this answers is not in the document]
ggriffiths: I think the string simplifies things when it comes to recording the metrics. Otherwise, we would need to have a switch statement to get the string version of an enum. I can switch to that model if needed though.

ggriffiths:
Ready for re-review. I think there's some bug which causes failed snapshot.UpdateStatus API calls to still update the local snapshot cache. As a result, metrics are never emitted after a failed snapshot.UpdateStatus API call. This is because we look at the cached Snapshot to see if a metric needs to be recorded.

As a workaround, I'm now recording metrics just before `snapshot.UpdateStatus`:
https://github.com/kubernetes-csi/external-snapshotter/pull/409/files#diff-1a3473f9016f7abcf5007414ff7141e39032aaf7aba079eeef262d04167208a2R1128-R1148

pkg/metrics/metrics.go
```diff
@@ -69,18 +109,39 @@ type MetricsManager interface {
...
 
 // Operation is a structure which holds information to identify a snapshot
 // related operation
 type Operation struct {
-	// the name of the operation, for example: "CreateSnapshot", "DeleteSnapshot"
+	// Name is the name of the operation, for example: "CreateSnapshot", "DeleteSnapshot"
 	Name string
-	// the name of the driver which executes the operation
+	// Driver is the driver name which executes the operation
 	Driver string
```
yuxiangqian, line 121: discussed offline.
Putting driver name into the key (operation) seems to bring quite some problems with little benefits.
The name of the operation + resource UID should already be unique enough.
same thing applies to snapshot type.
There is no guarantee that a valid VolumeSnapshotClass exists when a VolumeSnapshot CR is created for dynamic provisioning.

given that, I am proposing the following:

key ->
type Operation struct {
  operation name
  resource uid
}

value:
type val struct {
  start ts
  driver
  snapshot type
}

pkg/common-controller/snapshot_controller.go
```diff
@@ -664,6 +693,7 @@ func (ctrl *csiSnapshotCommonController) createSnapshotContent(snapshot *crdv1.V
 	_, err = ctrl.storeContentUpdate(updateContent)
 	if err != nil {
 		klog.Errorf("failed to update content store %v", err)
+		return nil, err
```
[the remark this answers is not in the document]
xing-yang: +1

pkg/common-controller/snapshot_controller.go
```diff
@@ -1186,6 +1237,49 @@ func (ctrl *csiSnapshotCommonController) getSnapshotClass(className string) (*cr
...
 }
 
+// getSnapshotDriverName is a helper function to get snapshot driver from the VolumeSnapshot.
+// We try to get the driverName in multiple ways, as snapshot controller metrics depend on the correct driverName.
+func (ctrl *csiSnapshotCommonController) getSnapshotDriverName(vs *crdv1.VolumeSnapshot) (string, error) {
+	klog.V(5).Infof("getSnapshotDriverName: VolumeSnapshot[%s]", vs.Name)
+
+	var driverName string
+	// Try to get from snapshot class first
+	if vs.Spec.VolumeSnapshotClassName != nil {
```
yuxiangqian, line 1247: I think we might need to swap the logic like following:

get driver from content.Spec.Driver
if content not found
get driver from snapshot class

There are two major reasons:
1. content.Spec.Driver is populated by controller in dynamic provisioning case. while snapshot class can be deleted and recreated, the new driver will not be propagated to the content, however in VolumeSnapshot it holds only a reference
2. in preprovisioned case, driver is a required field for content while there could be NO snapshot class.
WDYT?

yuxiangqian (commented):
thanks @ggriffiths for the work!

pkg/common-controller/snapshot_controller.go
ggriffiths: Agreed, moved this around to align with the above suggestion.

pkg/metrics/metrics.go
```diff
@@ -105,46 +171,95 @@ func NewMetricsManager() MetricsManager {
...
-func (opMgr *operationMetricsManager) DropOperation(op Operation) {
+// OperationStart drops an operation
+func (opMgr *operationMetricsManager) DropOperation(op OperationKey) {
 	opMgr.cache.Delete(op)
 }
 
-func (opMgr *operationMetricsManager) RecordMetrics(op Operation, status OperationStatus) {
-	obj, exists := opMgr.cache.Load(op)
+// RecordMetrics emits operation metrics
+func (opMgr *operationMetricsManager) RecordMetrics(opKey OperationKey, opStatus OperationStatus, snapshot *crdv1.VolumeSnapshot, driverName string) {
```
yuxiangqian, line 186: rm "snapshot"? its no longer needed?

pkg/metrics/metrics.go
```diff
@@ -105,46 +171,95 @@ func NewMetricsManager() MetricsManager {
...
+			// record a cancel metric if found
+			opMgr.recordCancelMetric(obj, createAndReadyKey, operationDuration)
+		}
+	}
+
+	opMgr.cache.Delete(opKey)
+}
+
+// recordCancelMetric records a metric for a create operation that hasn't finished
+func (opMgr *operationMetricsManager) recordCancelMetric(obj interface{}, key OperationKey, duration float64) {
```
yuxiangqian, line 236: I think we should simply have "opMgr.opLatencyMetrics.WithLabelValues(
		val.Driver,
		key.Name,
		val.SnapshotType,
		string(SnapshotStatusTypeCancel)," in the RecordMetrics call. Cache management will happen and after this call again?

yuxiangqian (commented):
some nits.
@msau42 PTAL, for per reconciliation metric, we will do a followup.

pkg/metrics/metrics.go
ggriffiths: if we're deleting this snapshot, won't that cause a leak?

When we're record a delete metric, we must
1. Load Delete operation key, record metric, and delete key from cache
2. Load Create operation key, record cancel metric, and delete key from cache
3. Load CreateAndReady operation key, record metric, and delete key from cache.

Since the operation name is part of the key, we have three unique keys to delete (if Create/CreateAndReady keys still exist)

yuxiangqian: oh, yeah you are right. I read it wrongly :)

pkg/common-controller/snapshot_controller_base.go
```diff
@@ -407,6 +413,13 @@ func (ctrl *csiSnapshotCommonController) updateContent(content *crdv1.VolumeSnap
 func (ctrl *csiSnapshotCommonController) deleteSnapshot(snapshot *crdv1.VolumeSnapshot) {
 	_ = ctrl.snapshotStore.Delete(snapshot)
 	klog.V(4).Infof("snapshot %q deleted", utils.SnapshotKey(snapshot))
+	driverName, err := ctrl.getSnapshotDriverName(snapshot)
+	if err != nil {
+		klog.Errorf("failed to getSnapshotDriverName while recording metrics for snapshot %q: %s", utils.SnapshotKey(snapshot), err)
+	} else {
+		deleteOperationKey := metrics.NewOperationKey(metrics.DeleteSnapshotOperationName, snapshot.UID)
+		ctrl.metricsManager.RecordMetrics(deleteOperationKey, metrics.NewSnapshotOperationStatus(metrics.SnapshotStatusTypeSuccess), driverName)
```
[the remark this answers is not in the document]
msau42: Similar question for the createsnapshot operations

cmd/snapshot-controller/main.go
```diff
@@ -48,6 +50,9 @@ var (
 
 	leaderElection          = flag.Bool("leader-election", false, "Enables leader election.")
 	leaderElectionNamespace = flag.String("leader-election-namespace", "", "The namespace where the leader election resource exists. Defaults to the pod namespace if not set.")
+
+	httpEndpoint = flag.String("http-endpoint", "", "The TCP network address where the http server will listen (example: `:8080`). The default is empty string, which means the http endpoint is disabled.")
```
verult, line 54: nit: could be more descriptive on what the server is, i.e:

"The TCP network address where the HTTP server for diagnostics, including metrics, will listen (example: `:8080`). The default is empty string, which means the server is disabled."

pkg/common-controller/snapshot_controller_base.go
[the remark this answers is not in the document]
ggriffiths: Yes, I'm not sure how we can address that short of occasionally checking and cleaning up keys for snapshots that no longer exist. We could run that as a scheduled goroutine. 

How do we handle this in the snapshotStore and contentStore caches @yuxiangqian? Are those cache records cleared on force deletes?

pkg/common-controller/snapshot_controller.go
```diff
@@ -1095,10 +1136,30 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
...
+
+		// We need to record metrics before updating the status due to a bug causing cache entries after a failed UpdateStatus call.
+		// Must meet the following criteria to emit a successful CreateSnapshot status
+		// 1. Previous status was nil OR Previous status had a nil CreationTime
+		// 2. New status must be non-nil with a non-nil CreationTime
+		driverName := content.Spec.Driver
+		if !utils.IsSnapshotCreated(snapshotObj) && utils.IsSnapshotCreated(snapshotClone) {
+			createOperationKey := metrics.NewOperationKey(metrics.CreateSnapshotOperationName, snapshot.UID)
+			ctrl.metricsManager.RecordMetrics(createOperationKey, metrics.NewSnapshotOperationStatus(metrics.SnapshotStatusTypeSuccess), driverName)
+		}
```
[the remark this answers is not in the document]
ggriffiths: Ah I see - yes this would cause a leak. For example, if the CreateSnapshot operation took 1 reconcile to record, and CreateSnapshotAndReady took 5 reconciles to record, we would have 4 leaked CreateSnapshot operations. We can drop the metric in this case if we determine the old and new snapshots are already created.

pkg/common-controller/snapshot_controller.go
```diff
@@ -1095,10 +1136,30 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
 	if updated {
 		snapshotClone := snapshotObj.DeepCopy()
 		snapshotClone.Status = newStatus
+
+		// We need to record metrics before updating the status due to a bug causing cache entries after a failed UpdateStatus call.
```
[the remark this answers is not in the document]
ggriffiths: I don't think so. Since this is the first time the controller has seen that the snapshot has been created or is ready, it is okay to record IMO.

The update status will go through on the next reconcile so that the end user sees the snapshot as ready. 

The problem here was that UpdateStatus failed, but the snapshotStore cache recorded readyToUse as true. On the next reconcile loop, this snapshot would hit syncReadySnapshot, and update status with readyToUse=true (without starting a new operation, because readyToUse=true).

cmd/snapshot-controller/main.go
ggriffiths: Thanks, added this in

pkg/common-controller/snapshot_controller.go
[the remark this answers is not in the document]
ggriffiths: Added a case to handle this.

pkg/common-controller/snapshot_controller_base.go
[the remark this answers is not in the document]
yuxiangqian: My understanding is even a volumesnapshot has been deleted by user forcefully, the controller would still receive an event from API server. The controller should still have the snapshot locally cached in 'snapshotStore' (controller crash will free any RAM allocated anyways), thus in this case there should be no leaking as line 420 will be executed?

@ggriffiths as I suggested previously to remove the "else" block for metric reporting though. otherwise if the snapshotclass has been deleted, it will be a leak.

pkg/common-controller/snapshot_controller.go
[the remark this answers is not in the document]
yuxiangqian: should we do this in "syncUnreadySnapshot" instead of here? i.e., if a snapshot has been cut, do not add the "createOperationKey" operation instead of an explicit dropping. I think that is a more clean solution. WDYT? @ggriffiths

pkg/common-controller/snapshot_controller_base.go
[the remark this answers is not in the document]
msau42: Do you think it would be worth adding some metrics into the operation cache so we could potentially detect leaks? I'm worried this way of metrics caching is prone to leaking especially if we introduce/change logic flow in the future.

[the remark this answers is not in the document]
ggriffiths: Yes, I think that could be a good idea. I'm thinking that might be a good followup issue, to keep this PR simple.

pkg/common-controller/snapshot_controller.go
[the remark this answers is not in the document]
ggriffiths: Moved this to start in syncUnreadySnapshot only when the snapshot has not been created yet.

pkg/common-controller/snapshot_controller_base.go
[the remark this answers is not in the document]
ggriffiths: Let me know if that is okay @msau42 and I'll open a separate issue/PR to do that work. If not, I can add it sometime this week in this PR.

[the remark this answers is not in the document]
msau42: Followup is fine, thanks!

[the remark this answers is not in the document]
ggriffiths: Sounds good, opened a new issue here: https://github.com/kubernetes-csi/external-snapshotter/issues/434

pkg/common-controller/snapshot_controller.go
```diff
@@ -1094,10 +1138,32 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
...
+		// Must meet the following criteria to emit a successful CreateSnapshot status
+		// 1. Previous status was nil OR Previous status had a nil CreationTime
+		// 2. New status must be non-nil with a non-nil CreationTime
+		driverName := content.Spec.Driver
+		createOperationKey := metrics.NewOperationKey(metrics.CreateSnapshotOperationName, snapshot.UID)
+		if !utils.IsSnapshotCreated(snapshotObj) && utils.IsSnapshotCreated(snapshotClone) {
+			ctrl.metricsManager.RecordMetrics(createOperationKey, metrics.NewSnapshotOperationStatus(metrics.SnapshotStatusTypeSuccess), driverName)
+		}
+
+		// If the old snapshot and new snapshot have both already been created, make sure to drop the CreateSnapshot operation.
```
yuxiangqian, line 1152: this comment is no longer needed

yuxiangqian (commented):
one nit and LGTM. thanks @ggriffiths

pkg/common-controller/snapshot_controller.go
ggriffiths: Removed this

PR #420, opened 2020-11-03
Switch to klog/v2

xing-yang:
**What type of PR is this?**
/kind cleanup

Switch remaining files to use klog/v2.

Fixes #

```release-note
Log output from klog/v2 at higher log levels is now available.
```


cmd/snapshot-validation-webhook/main.go
```diff
@@ -20,7 +20,7 @@ import (
 	"flag"
 
 	webhook "github.com/kubernetes-csi/external-snapshotter/v3/pkg/validation-webhook"
-	"k8s.io/klog"
+	klog "k8s.io/klog/v2"
```
pohly, line 23: nit: you don't need to rename the module on import. It's not wrong, just redundant. Did you add that manually or was that something that your editor did?

pohly (changes requested):
This is a bug fix, so the release notes should mention it.

In https://github.com/kubernetes-csi/external-provisioner/pull/518 I used "log output from klog/v2 at higher log levels is now available".

cmd/snapshot-validation-webhook/main.go
xing-yang: Done

xing-yang:
@pohly addressed your comments.  PTAL.

pkg/validation-webhook/config.go
```diff
@@ -19,7 +19,7 @@ package webhook
 import (
 	"crypto/tls"
 
-	"k8s.io/klog"
+	klog "k8s.io/klog/v2"
```
pohly, line 22: Still some redundant "klog" renaming. Also below.

xing-yang: Done

xing-yang:
@pohly addressed your comments.  PTAL.

vendor/modules.txt
```diff
@@ -531,9 +531,8 @@ k8s.io/client-go/util/workqueue
 k8s.io/component-base/metrics
 k8s.io/component-base/version
 # k8s.io/klog v1.0.0
-## explicit
 k8s.io/klog
-# k8s.io/klog/v2 v2.2.0
+# k8s.io/klog/v2 v2.3.0
```
pohly, line 535: k8s.io/klog is still being pulled in by `vendor/github.com/kubernetes-csi/csi-lib-utils/connection/connection.go`.

Jan is working on a new release of csi-lib-utils which uses klog/v2. We need to update to that upcoming release.

xing-yang: ok

xing-yang: Can we get this PR merged first?  Then a followup PR can be done to update csi-lib-utils?

pohly: If we merge this PR as-is, we loose log output from csi-lib-utils because it then logs via klog/v1 with default log level.

We really should switch everything at once.

msau42: I'm ok merging it in phases as long as we don't release without it.

xing-yang: I won't cut a release without the csi-lib-utils change.  I see that there are 2 PRs in external-attacher to complete this change as well.

PR #424, opened 2020-11-10
added labels to the resources to be used for collective operation

aayushrangwala:
**What type of PR is this?**
/kind feature

We need to add the common/generic labels to all the resources which are being deployed as the part of any csi driver. It is the part of this [PR](https://github.com/kubernetes-csi/csi-driver-host-path/pull/216)

Since there was no common labels to identify all the items in one set of resources installed, these labels can help us to identify resources and operate on them in a collective manner

NONE

```release note
NONE
```


pohly:
These labels cannot  be set here. They need to be added when deploying a specific CSI driver - see https://github.com/kubernetes-csi/csi-driver-host-path/pull/216#discussion_r520438174

PR #425, opened 2020-11-12
Update controller based on snapshot v1 apis

xing-yang:
**What type of PR is this?**
/kind feature

This PR updates controller code based on snapshot v1 apis

Fixes #

```release-note
Update snapshot-controller, CSI snapshotter sidecar, and validation webhook based on v1 apis.
```


pkg/validation-webhook/snapshot.go
```diff
@@ -20,7 +20,7 @@ import (
 	"fmt"
 	"reflect"
 
-	volumesnapshotv1beta1 "github.com/kubernetes-csi/external-snapshotter/client/v3/apis/volumesnapshot/v1beta1"
+	volumesnapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v3/apis/volumesnapshot/v1beta1"
```
msau42, line 23: this is odd? we call it v1 but it's actually v1beta1.

pkg/validation-webhook/snapshot.go
```diff
@@ -55,24 +55,24 @@ func admitSnapshot(ar v1.AdmissionReview) *v1.AdmissionResponse {
 	deserializer := codecs.UniversalDeserializer()
 	switch ar.Request.Resource {
 	case SnapshotV1Beta1GVR:
```
msau42, line 57: do we need new case statements for v1?

pkg/validation-webhook/snapshot.go
xing-yang: ok, I"ll change that.

xing-yang: Done

pkg/validation-webhook/snapshot.go
xing-yang: I think we only need to handle v1beta1 here as we use oneof in the v1 manifest.

msau42: The default case throws an error, is that going to be problematic?

xing-yang: Probably safer just to ignore the error and log a message?

msau42: or add a case statement but don't validate anything.

Is there anything that we validate for v1beta that still needs to be validated for v1?

xing-yang: For invalid v1 objects, we get the following message from the "oneOf" restriction:
```
kubectl create -f invalid-snapshot_v1.yaml
The VolumeSnapshot "invalid-snapshot-demo" is invalid: : Invalid value: "": "spec.source" must validate one and only one schema (oneOf). Found 2 valid alternatives
```

For invalid v1beta1 objects, we get the following message from the webhook:
```
kubectl create -f invalid-snapshot_beta.yaml
Error from server: error when creating "invalid-snapshot_beta.yaml": admission webhook "validation-webhook.snapshot.storage.k8s.io" denied the request: only one of Spec.Source.PersistentVolumeClaimName = pvc and Spec.Source.VolumeSnapshotContentName = vsc should be set
```

xing-yang: We do need to return a *v1.AdmissionResponse here.  So we can't just do nothing.   We need to specify Allowed to true or false.
```
        reviewResponse := &v1.AdmissionResponse{
                Allowed: true,
                Result:  &metav1.Status{},
        }
```

xing-yang: @yuxiangqian can you take a look?

msau42: Hm so under the webhook config, we only have rules for v1beta: https://github.com/kubernetes-csi/external-snapshotter/blob/master/deploy/kubernetes/webhook-example/admission-configuration-template#L9

I guess that means v1 objects will not go through the webhook, and actually at some point after v1beta deprecation, we don't need the webhook?

xing-yang: I think so.  @yuxiangqian can you confirm?

yuxiangqian: I am afraid we need to add support for v1 resources as well.
@msau42  to your point, currently the webhook server will **ONLY** receive v1beta1 object related admission requests, and its configured [here](https://github.com/kubernetes-csi/external-snapshotter/blob/master/deploy/kubernetes/webhook-example/admission-configuration-template#L9)

this would need to be extended to also cover v1 in this or a separated PR.
if only handle v1beta1 objects, we will not have problem with **OneOf** semantic as v1 schema guarantees that, however it might cause problems with other validation rules, i.e., immutable checks [here](https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/validation-webhook/snapshot.go#L158) will not apply to v1 resources.

yuxiangqian: I think it makes sense to introduce v1 admission along with v1beta1 -> v1(GA) in the validation webhook. The change should not be very big. For v1 objects, we can ignore **OneOf** checks and leave it to schema level validation, but to keep it consistent, immutable checks/other validation checks(for [example](https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/utils/util.go#L140)) should be implemented on v1 objects.

xing-yang: > The default case throws an error, is that going to be problematic?
@yuxiangqian Can you answer the above question from Michelle?

xing-yang:
@msau42 @yuxiangqian Comments are addressed.  PTAL.  Thanks.

pkg/validation-webhook/snapshot.go
```diff
@@ -77,15 +94,59 @@ func admitSnapshot(ar v1.AdmissionReview) *v1.AdmissionResponse {
...
+	// Enforce strict validation for CREATE requests. Immutable checks don't apply for CREATE requests.
+	// Enforce strict validation for UPDATE requests where old is valid and passes immutability check.
+	if err := utils.ValidateV1Beta1Snapshot(snapshot); err != nil {
+		reviewResponse.Allowed = false
+		reviewResponse.Result.Message = err.Error()
+	}
+	return reviewResponse
+}
+
+func decideSnapshotV1(snapshot, oldSnapshot *volumesnapshotv1.VolumeSnapshot, isUpdate bool) *v1.AdmissionResponse {
```
msau42, line 149: for v1 we don't need ratcheting validation check right? we should always check for immutability

xing-yang: For v1, the webhook does not check if there is only one source as that is validated by the schema already.
In utils.ValidateSnapshot, we do have a check to make sure the VolumeSnapshotClassName is not an empty string.

msau42: Yes, I meant, can we remove the block L155 below that checks if the original object is validate on an update?

xing-yang: Done

xing-yang:
@msau42 addressed your comments.  PTAL.

pkg/validation-webhook/snapshot.go
msau42: Thanks! will let @yuxiangqian provide final review

pkg/common-controller/snapshot_update_test.go
```diff
@@ -406,54 +406,6 @@ func TestSync(t *testing.T) {
...
 			test:              testSyncSnapshot,
 		},
-		// TODO(xiangqian@): remove test cases 7-2/7-3 when webhooks are in place
-		//                   https://github.com/kubernetes-csi/external-snapshotter/issues/187
-		{
-			name:              "7-2 - validation fail if neither VolumeSnapshotContentName nor PersistentVolumeClaimName has been specified in Snapshot.Spec.Source",
-			initialContents:   nocontents,
-			expectedContents:  nocontents,
-			initialSnapshots:  newSnapshotArray("snap7-2", "snapuid7-2", "", "", validSecretClass, "", &False, nil, nil, nil, false, true, nil),
-			expectedSnapshots: withSnapshotInvalidLabel(newSnapshotArray("snap7-2", "snapuid7-2", "", "", validSecretClass, "", &False, nil, nil, newVolumeError("Exactly one of PersistentVolumeClaimName and VolumeSnapshotContentName should be specified"), false, true, nil)),
```
yuxiangqian, line 416: should we reserve some of these test cases for labeling?

pkg/utils/util.go
```diff
@@ -129,6 +130,21 @@ func ValidateSnapshot(snapshot *crdv1.VolumeSnapshot) error {
 		return fmt.Errorf("VolumeSnapshot is nil")
 	}
 
+	vscname := snapshot.Spec.VolumeSnapshotClassName
```
yuxiangqian, line 133: I think we should have a followup PR to move all these validating logic into webhook folder directly. having them defined here makes utils package depending on both v1 and v1beta1 which seems to be unnecessary. note that utils is used in controllers. let's create an issue to follow up on this?

pkg/validation-webhook/snapshot.go
```diff
@@ -30,8 +31,12 @@ import (
 var (
 	// SnapshotV1Beta1GVR is GroupVersionResource for volumesnapshots
 	SnapshotV1Beta1GVR = metav1.GroupVersionResource{Group: volumesnapshotv1beta1.GroupName, Version: "v1beta1", Resource: "volumesnapshots"}
+	// SnapshotV1GVR is GroupVersionResource for volumesnapshots
```
yuxiangqian, line 34: how about make the comments more specific, or remove them as the names already suggested?
// SnapshotV1GVR is GroupVersionResource for v1 VolumeSnapshots
// SnapshotV1GVR is GroupVersionResource for v1beta1 VolumeSnapshots
// // SnapshotContentV1Beta1GVR is GroupVersionResource for v1beta1 VolumeSnapshotContents
etc.

pkg/validation-webhook/snapshot.go
```diff
@@ -97,12 +126,35 @@ func decideSnapshot(snapshot, oldSnapshot *volumesnapshotv1beta1.VolumeSnapshot,
...
+	// Enforce strict validation for CREATE requests. Immutable checks don't apply for CREATE requests.
+	// Enforce strict validation for UPDATE requests where old is valid and passes immutability check.
+	if err := utils.ValidateV1Beta1Snapshot(snapshot); err != nil {
+		reviewResponse.Allowed = false
+		reviewResponse.Result.Message = err.Error()
+	}
+	return reviewResponse
+}
 
+func decideSnapshotV1(snapshot, oldSnapshot *volumesnapshotv1.VolumeSnapshot, isUpdate bool) *v1.AdmissionResponse {
```
yuxiangqian, line 149: nit, either we remove "v1" for all v1 resource related functions or variables, or append "v1" for all to make it consistent. i.e., ValidateSnapshot

pkg/validation-webhook/snapshot.go
```diff
@@ -117,7 +169,30 @@ func decideSnapshot(snapshot, oldSnapshot *volumesnapshotv1beta1.VolumeSnapshot,
...
 
-func decideSnapshotContent(snapcontent, oldSnapcontent *volumesnapshotv1beta1.VolumeSnapshotContent, isUpdate bool) *v1.AdmissionResponse {
+func decideSnapshotContentV1beta1(snapcontent, oldSnapcontent *volumesnapshotv1beta1.VolumeSnapshotContent, isUpdate bool) *v1.AdmissionResponse {
+	reviewResponse := &v1.AdmissionResponse{
+		Allowed: true,
+		Result:  &metav1.Status{},
+	}
+
+	if isUpdate {
+		// if it is an UPDATE and oldSnapcontent is valid, check immutable fields
```
yuxiangqian, line 179: hmmm, this does not seem to be right.
there should be a logic similar to line 129.
if oldSnapshotContent is already invalid, the admission hook should allow the update.
while in "decideSnapshotContentV1", it does have that piece of logic, which is not needed

pkg/utils/util.go
xing-yang: Sure

pkg/validation-webhook/snapshot.go
xing-yang: hmm, looks like I removed the wrong block of code.  I meant to remove line 207.

pkg/common-controller/snapshot_update_test.go
xing-yang: Does that mean we need to keep the logic that checks whether there are two fields in the source for V1?  Otherwise, when we validate v1 snapshots, we are only checking if snapshotClassName is not empty string.

xing-yang: Discussed about this.  Removed these tests.

xing-yang:
@yuxiangqian Your comments are addressed.  PTAL.  Thanks.

yuxiangqian (commented):
LGTM

PR #428, opened 2020-11-17
using csi driver from PV instead of from StorageClass

### LINKED ISSUE #404: Use driver name from PV instead of from StorageClass to validate
In current implementation, [SetDefaultSnapshotClass](https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/common-controller/snapshot_controller.go#L1189) relies on PVC's StorageClass to find the driver name. Instead its more appropriate to rely on PV's driver name as StorageClass can be updated after a volume has been provisioned.

Per discussion in https://github.com/kubernetes/kubernetes/issues/94312

yuxiangqian:
**What type of PR is this?**
>

/kind cleanup

In current implementation, SetDefaultSnapshotClass relies on PVC's StorageClass to find the driver name. Instead it's more appropriate to rely on PV's driver name as StorageClass can be updated after a volume has been provisioned.

This also removes the dependency on storage/v1

Fixes #404

```release-note
NONE

```

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -23,7 +23,6 @@ import (
 	"time"
 
 	v1 "k8s.io/api/core/v1"
-	storagev1 "k8s.io/api/storage/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


xing-yang:
@chrishenzie can you take a look since you opened this issue https://github.com/kubernetes/kubernetes/issues/94312

pkg/common-controller/snapshot_controller.go
```diff
@@ -1149,28 +1148,20 @@ func (ctrl *csiSnapshotCommonController) isVolumeBoundToClaim(volume *v1.Persist
...
-	if len(storageclassName) == 0 {
-		return nil, fmt.Errorf("cannot figure out the snapshot class automatically, please specify one in snapshot spec")
+		return "", err
 	}
-	storageclass, err := ctrl.client.StorageV1().StorageClasses().Get(context.TODO(), storageclassName, metav1.GetOptions{})
-	if err != nil {
-		return nil, err
+	// supports ONLY CSI volumes
+	if pv.Spec.PersistentVolumeSource.CSI == nil {
+		return "", fmt.Errorf("snapshot an non-CSI volume is not supported, snapshot:%s/%s", snapshot.Namespace, snapshot.Name)
```
chrishenzie, line 1162: nit: snapshotting non-CSI volumes are not supported*

pkg/common-controller/snapshot_controller.go
```diff
@@ -1197,21 +1190,23 @@ func (ctrl *csiSnapshotCommonController) SetDefaultSnapshotClass(snapshot *crdv1
...
+	list, err := ctrl.classLister.List(labels.Everything())
 	if err != nil {
 		return nil, nil, err
 	}
-	// Find default snapshot class if available
-	list, err := ctrl.classLister.List(labels.Everything())
+
+	pvDriver, err := ctrl.pvDriverFromSnapshot(snapshot)
 	if err != nil {
+		klog.Errorf("failed to get pv csi driver from snapshot %s-%s: %q", snapshot.Namespace, snapshot.Name, err)
```
chrishenzie, line 1201: nit: It may be worth having consistent formatting of snapshot strings (`%s/%s` above vs. `%s-%s` here).

chrishenzie (commented):
Other than two small nits this LGTM

pkg/common-controller/snapshot_controller.go
yuxiangqian: done

pkg/common-controller/snapshot_controller.go
yuxiangqian: done

yuxiangqian:
@xing-yang @chrishenzie PTAL, thnx

PR #429, opened 2020-11-18
Add certwatcher to webhook server

### LINKED ISSUE #422: snapshot-validation-webhook should reload server certificate on update
If the server certificate for snapshot-validation-webhook is updated, we should begin using the new certificate instead of the old.

This is required if we are to support credential rotation. For example, an external service may update the k8s secret with this pod's certificate.

hat tip to @msau42 for pointing out that [controller runtime][1] already has this feature.

[1]: https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/webhook/internal/certwatcher

ggriffiths:

**What type of PR is this?**
/kind bug

* Adds the certwatcher implementation from the internal pkg https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/internal/certwatcher/certwatcher.go
* Using github.com/kubernetes-sigs/controller-runtime/pkg/webhook directly is more involved at the moment and would require a breaking change (controller-runtime assumes the Cert/Key files are in the same directory, which external-snapshotter currently does not assume).

Fixes #422  

n/a

```release-note
The validation-webhook server will now watch for changes to --tls-cert-file and --tls-private-key-file and update the server's TLS config accordingly.
```

### DIFF
--- a/pkg/validation-webhook/certwatcher.go
+++ b/pkg/validation-webhook/certwatcher.go
@@ -1,0 +1,164 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhook
+
+import (
+	"context"
+	"crypto/tls"
+	"sync"
+
+	"github.com/fsnotify/fsnotify"
+	"k8s.io/klog"
+)
+
+// This file originated from github.com/kubernetes-sigs/controller-runtime/pkg/webhook/internal/certwatcher.
+// We cannot import this package as it's an internal one. In addition, we cannot yet easily integrate
+// with controller-runtime/pkg/webhook directly, as it would require extensive rework:
+// https://github.com/kubernetes-csi/external-snapshotter/issues/422
+
+// CertWatcher watches certificate and key files for changes.  When either file
+// changes, it reads and parses both and calls an optional callback with the new
+// certificate.
+type CertWatcher struct {
+	sync.Mutex
+
+	currentCert *tls.Certificate
+	watcher     *fsnotify.Watcher
+
+	certPath string
+	keyPath  string
+}
+
+// NewCertWatcher returns a new CertWatcher watching the given certificate and key.
+func NewCertWatcher(certPath, keyPath string) (*CertWatcher, error) {
+	var err error
+
+	cw := &CertWatcher{
+		certPath: certPath,
+		keyPath:  keyPath,
+	}
+
+	// Initial read of certificate and key.
+	if err := cw.ReadCertificate(); err != nil {
+		return nil, err
+	}
+
+	cw.watcher, err = fsnotify.NewWatcher()
+	if err != nil {
+		return nil, err
+	}
+
+	return cw, nil
+}
+
+// GetCertificate fetches the currently loaded certificate, which may be nil.
+func (cw *CertWatcher) GetCertificate(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	cw.Lock()
+	defer cw.Unlock()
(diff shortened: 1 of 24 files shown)


xing-yang:
Thanks @ggriffiths!  That was quick:)

pkg/validation-webhook/certwatcher.go
```diff
@@ -0,0 +1,159 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
xing-yang, line 2: 2020

pkg/validation-webhook/certwatcher.go
```diff
@@ -0,0 +1,159 @@
...
+
+import (
+	"context"
+	"crypto/tls"
+	"sync"
+
+	"github.com/fsnotify/fsnotify"
+	"k8s.io/klog"
+)
+
```
xing-yang, line 27: Is it because our webhook is not generated by controller-gen, so that we can't directly use "github.com/kubernetes-sigs/controller-runtime/pkg/webhook/internal/certwatcher"?

If so, please add comments in the beginning of this file.

ggriffiths: We cannot import github.com/kubernetes-sigs/controller-runtime/pkg/webhook/internal/certwatcher because it is an internal package:
https://golang.org/doc/go1.4#internalpackages

Added a comment about the origin of this file in certwatcher.go

xing-yang:
@ggriffiths Can you do some manual testing and simulate a cert rotation?

ggriffiths:
Yes I'll give that a shot.
/hold

ggriffiths:
Tested this with valid and invalid certs @xing-yang. The valid cert was rotated and refreshed in memory. The invalid cert caused problems without a pod restart, as expected.

Valid cert rotation:
1. attempted to create invalid snapshot, correctly denied
```
[grant@PDC4-SM26-N8 external-snapshotter]$ date
Tue Dec  1 15:32:50 PST 2020
[grant@PDC4-SM26-N8 external-snapshotter]$ kubectl create -f ./examples/kubernetes/invalid-snapshot.yaml
Error from server: error when creating "./examples/kubernetes/invalid-snapshot.yaml": admission webhook "validation-webhook.snapshot.storage.k8s.io" denied the request: only one of Spec.Source.PersistentVolumeClaimName = pvc and Spec.Source.VolumeSnapshotContentName = vsc should be set
```
2. Re-generated the cert
```
[grant@PDC4-SM26-N8 external-snapshotter]$ ./deploy/kubernetes/webhook-example/create-cert.sh --service snapshot-validation-service --secret snapshot-validation-secret --namespace kube-system
creating certs in tmpdir /tmp/tmp.9eOHCqeCsm 
Generating RSA private key, 2048 bit long modulus
...........................................................+++
..........................+++
e is 65537 (0x10001)
certificatesigningrequest.certificates.k8s.io "snapshot-validation-service.kube-system" deleted
certificatesigningrequest.certificates.k8s.io/snapshot-validation-service.kube-system created
NAME                                      AGE   REQUESTOR          CONDITION
snapshot-validation-service.kube-system   0s    kubernetes-admin   Pending
certificatesigningrequest.certificates.k8s.io/snapshot-validation-service.kube-system approved
secret/snapshot-validation-secret configured
[grant@PDC4-SM26-N8 external-snapshotter]$ date
Tue Dec  1 15:33:14 PST 2020
```
3. TLS certificate updated in the validation pod
 `I1201 23:33:20.251368       1 certwatcher.go:129] Updated current TLS certificate`

4. Validation webhook still works
```
[grant@PDC4-SM26-N8 external-snapshotter]$ kubectl create -f ./examples/kubernetes/invalid-snapshot.yaml
Error from server: error when creating "./examples/kubernetes/invalid-snapshot.yaml": admission webhook "validation-webhook.snapshot.storage.k8s.io" denied the request: only one of Spec.Source.PersistentVolumeClaimName = pvc and Spec.Source.VolumeSnapshotContentName = vsc should be set
```

Performed a similar flow above with an invalid cert and saw the below error, as expected. 
```
[grant@PDC4-SM26-N8 external-snapshotter]$ kk apply -f examples/kubernetes/invalid-snapshot.yaml 
Error from server (InternalError): error when creating "examples/kubernetes/invalid-snapshot.yaml": Internal error occurred: failed calling webhook "validation-webhook.snapshot.storage.k8s.io": Post https://snapshot-validation-service.kube-system.svc:443/volumesnapshot?timeout=2s: x509: certificate has expired or is not yet vali
```

xing-yang:
Thanks @ggriffiths!

/lgtm
/approve

PR #433, opened 2020-12-02
Don't change ReadyToUse in Snapshot when logging an event

### LINKED ISSUE #432: deleting snapshotclass is marking volumesnapshot readyToUse as false
deleting the snapshot class which is used to create volume snapshots is leading to marking all volume snapshot readyToUse  as false

steps to reproduce:-
* create volumesnapshotclass
* create volumesnapshot
* wait till readyToUse is set to true
* Now delete the volumesnapshotclass
* check readyToUse on volumesnapshot

logs of snapshot controller

```text
I1201 04:44:46.953076       1 reflector.go:369] github.com/kubernetes-csi/external-snapshotter/client/v3/informers/externalversions/factory.go:117: forcing resync
I1201 04:44:46.953295       1 snapshot_controller_base.go:158] enqueued "default/rbd-pvc-snapshot" for sync
I1201 04:44:46.953349       1 snapshot_controller_base.go:202] syncSnapshotByKey[default/rbd-pvc-snapshot]
I1201 04:44:46.953366       1 snapshot_controller_base.go:205] snapshotWorker: snapshot namespace [default] name [rbd-pvc-snapshot]
I1201 04:44:46.953381       1 snapshot_controller_base.go:328] checkAndUpdateSnapshotClass [rbd-pvc-snapshot]: VolumeSnapshotClassName [csi-rbdplugin-snapclass]
I1201 04:44:46.953393       1 snapshot_controller.go:1178] getSnapshotClass: VolumeSnapshotClassName [csi-rbdplugin-snapclass]
E1201 04:44:46.953415       1 snapshot_controller.go:1182] failed to retrieve snapshot class csi-rbdplugin-snapclass from the informer: "volumesnapshotclass.snapshot.storage.k8s.io \"csi-rbdplugin-snapclass\" not found"
E1201 04:44:46.953477       1 snapshot_controller_base.go:331] checkAndUpdateSnapshotClass failed to getSnapshotClass volumesnapshotclass.snapshot.storage.k8s.io "csi-rbdplugin-snapclass" not found
I1201 04:44:46.953525       1 snapshot_controller.go:721] updateSnapshotStatusWithEvent[default/rbd-pvc-snapshot]
I1201 04:44:46.953539       1 snapshot_controller.go:724] updateSnapshotStatusWithEvent[rbd-pvc-snapshot]: the same error &{2020-12-01 04:37:46 +0000 UTC 0xc0004baf90} is already set
I1201 04:44:46.953602       1 snapshot_controller_base.go:192] Failed to sync snapshot "default/rbd-pvc-snapshot", will retry again: volumesnapshotclass.snapshot.storage.k8s.io "csi-rbdplugin-snapclass" not found
I1201 04:45:46.894293       1 reflector.go:369] github.com/kubernetes-csi/external-snapshotter/client/v3/informers/externalversions/factory.go:117: forcing resync
I1201 04:45:46.894521       1 snapshot_controller_base.go:175] enqueued "snapcontent-7a477074-ecd2-464c-a925-1892e8b2f419" for sync
I1201 04:45:46.894710       1 snapshot_controller_base.go:277] syncContentByKey[snapcontent-7a477074-ecd2-464c-a925-1892e8b2f419]
I1201 04:45:46.894864       1 util.go:264] storeObjectUpdate updating content "snapcontent-7a477074-ecd2-464c-a925-1892e8b2f419" with version 102029
I1201 04:45:46.895012       1 snapshot_controller.go:86] synchronizing VolumeSnapshotContent[snapcontent-7a477074-ecd2-464c-a925-1892e8b2f419]: content is bound to snapshot default/rbd-pvc-snapshot
I1201 04:45:46.895148       1 snapshot_controller.go:88] syncContent[snapcontent-7a477074-ecd2-464c-a925-1892e8b2f419]: check if we should add invalid label on content
I1201 04:45:46.895289       1 snapshot_controller.go:1370] getSnapshotFromStore: snapshot default/rbd-pvc-snapshot found
I1201 04:45:46.895453       1 snapshot_controller.go:991] needsUpdateSnapshotStatus[default/rbd-pvc-snapshot]
I1201 04:45:46.895740       1 snapshot_controller.go:143] synchronizing VolumeSnapshotContent for snapshot [default/rbd-pvc-snapshot]: update snapshot status to true if needed.
I1201 04:45:46.895980       1 snapshot_controller_base.go:202] syncSnapshotByKey[default/rbd-pvc-snapshot]
I1201 04:45:46.897389       1 snapshot_controller_base.go:205] snapshotWorker: snapshot namespace [default] name [rbd-pvc-snapshot]
I1201 04:45:46.897424       1 snapshot_controller_base.go:328] checkAndUpdateSnapshotClass [rbd-pvc-snapshot]: VolumeSnapshotClassName [csi-rbdplugin-snapclass]
I1201 04:45:46.897431       1 snapshot_controller.go:1178] getSnapshotClass: VolumeSnapshotClassName [csi-rbdplugin-snapclass]
E1201 04:45:46.897448       1 snapshot_controller.go:1182] failed to retrieve snapshot class csi-rbdplugin-snapclass from the informer: "volumesnapshotclass.snapshot.storage.k8s.io \"csi-rbdplugin-snapclass\" not found"
E1201 04:45:46.897523       1 snapshot_controller_base.go:331] checkAndUpdateSnapshotClass failed to getSnapshotClass volumesnapshotclass.snapshot.storage.k8s.io "csi-rbdplugin-snapclass" not found
I1201 04:45:46.897545       1 snapshot_controller.go:721] updateSnapshotStatusWithEvent[default/rbd-pvc-snapshot]
I1201 04:45:46.897554       1 snapshot_controller.go:724] updateSnapshotStatusWithEvent[rbd-pvc-snapshot]: the same error &{2020-12-01 04:37:46 +0000 UTC 0xc0004baf90} is already set
I1201 04:45:46.897624       1 snapshot_controller_base.go:192] Failed to sync snapshot "default/rbd-pvc-snapshot", will retry again: volumesnapshotclass.snapshot.storage.k8s.io "csi-rbdplugin-snapclass" not found
I1201 04:45:46.955714       1 reflector.go:369] github.com/kubernetes-csi/external-snapshotter/client/v3/informers/externalversions/factory.go:117: forcing resync
I1201 04:45:46.955856       1 snapshot_controller_base.go:158] enqueued "default/rbd-pvc-snapshot" for sync
I1201 04:45:46.955909       1 snapshot_controller_base.go:202] syncSnapshotByKey[default/rbd-pvc-snapshot]
I1201 04:45:46.955925       1 snapshot_controller_base.go:205] snapshotWorker: snapshot namespace [default] name [rbd-pvc-snapshot]
I1201 04:45:46.955938       1 snapshot_controller_base.go:328] checkAndUpdateSnapshotClass [rbd-pvc-snapshot]: VolumeSnapshotClassName [csi-rbdplugin-snapclass]
I1201 04:45:46.955947       1 snapshot_controller.go:1178] getSnapshotClass: VolumeSnapshotClassName [csi-rbdplugin-snapclass]
E1201 04:45:46.955963       1 snapshot_controller.go:1182] failed to retrieve snapshot class csi-rbdplugin-snapclass from the informer: "volumesnapshotclass.snapshot.storage.k8s.io \"csi-rbdplugin-snapclass\" not found"
E1201 04:45:46.956023       1 snapshot_controller_base.go:331] checkAndUpdateSnapshotClass failed to getSnapshotClass volumesnapshotclass.snapshot.storage.k8s.io "csi-rbdplugin-snapclass" not found
I1201 04:45:46.956043       1 snapshot_controller.go:721] updateSnapshotStatusWithEvent[default/rbd-pvc-snapshot]
I1201 04:45:46.956068       1 snapshot_controller.go:724] updateSnapshotStatusWithEvent[rbd-pvc-snapshot]: the same error &{2020-12-01 04:37:46 +0000 UTC 0xc0004baf90} is already set
I1201 04:45:46.956135       1 snapshot_controller_base.go:192] Failed to sync snapshot "default/rbd-pvc-snapshot", will retry again: volumesnapshotclass.snapshot.storage.k8s.io "csi-rbdplugin-snapclass" not found

```

xing-yang:
**What type of PR is this?**
/kind bug

ReadyToUse in VolumeSnapshot's Status should only be updated based on ReadyToUse in VolumeSnapshotContent's status.

Fixes #432

In the sidecar, `readyToUse` won't be changed after it becomes `true`:
https://github.com/kubernetes-csi/external-snapshotter/blob/v3.0.2/pkg/sidecar-controller/snapshot_controller.go#L79

In the common controller, readyToUse in snapshot is updated based on readyToUse in the content.

However, currently `readyToUse` is also always set to `false` by a call to [updateSnapshotErrorStatusWithEvent](https://github.com/kubernetes-csi/external-snapshotter/blob/v3.0.2/pkg/common-controller/snapshot_controller.go#L720). This can happen outside of [updateSnapshotStatus](https://github.com/kubernetes-csi/external-snapshotter/blob/v3.0.2/pkg/common-controller/snapshot_controller.go#L1022).

The Fix:
Update updateSnapshotStatusWithEvent to include a "setReadyToFalse bool" input parameter. If false, ReadyToUse will not be changed.
If snapshot class is not found, ReadyToUse will not be changed.
If updateSnapshotStatus failed, ReadyToUse will not be changed.
For all other cases, the code stays the same as before, meaning ReadyToUse will be False.

```release-note
ReadyToUse in VolumeSnapshot Status should not be changed when VolumeSnapshotClass is not found.
```


pkg/common-controller/snapshot_controller.go
```diff
@@ -778,8 +778,9 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotErrorStatusWithEvent(snap
 		Message: &message,
 	}
 	snapshotClone.Status.Error = statusError
-	ready := false
-	snapshotClone.Status.ReadyToUse = &ready
+	// NOTE(xyang): ReadyToUse in VolumeSnapshot's Status should only be updated based on
+	// ReadyToUse in VolumeSnapshotContent's Status. When ReadyToUse is already True
+	// in VolumeSnapshot's Status, we should not set it to false in updateSnapshotStatusWithEvent.
 	newSnapshot, err := ctrl.clientset.SnapshotV1beta1().VolumeSnapshots(snapshotClone.Namespace).UpdateStatus(context.TODO(), snapshotClone, metav1.UpdateOptions{})
```
yuxiangqian, line 784: @xing-yang  and I have discussed, and initially I agreed on this approach. In general I still think this is the right way to do things except one case which I am not 100% clear after looking into the code. It does have make the bi-directional binding verification more critical though. With this, situations like previously ready VolumeSnapshot resources will maintain the ready state forever(this seems to go against the "status represent the current state of a resource" principle)! Even in cases like losing binding from VolumeSnapshotContent, or VolumeSnapshotContent has been modified to refer to another VolumeSnapshot. In those cases, there is an error on this VolumeSnapshot and a short-lived event. I am a little concerned it might break existing user's expectation where they could rely on checking the ready flag (though they should not). Other than that, the PR LGTM.
@msau42 @saad-ali do you have any concerns?

xing-yang: We could also add an input parameter "readyToUse" in updateSnapshotErrorStatusWithEvent() and let the caller decide whether to change it or not.

msau42: To summarize, I think the safer approach would be something like:

- Do not set any status fields on VolumeSnapshot (except error) before binding is complete. Only after binding is complete we sync the other fields.
- Error/event setting methods only set error and events, don't modify any other fields

yuxiangqian: > To summarize, I think the safer approach would be something like:
> 
> * Do not set any status fields on VolumeSnapshot (except error) before binding is complete. Only after binding is complete we sync the other fields.

>> This is already the logic, problem is once the ready == true, the controller does binding checks and based on the situation using this exact call to update status in the cases where binding is invalid (either content has been deleted or content has been modified to point to another snapshot) [code](https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/common-controller/snapshot_controller.go#L399). With this change, the ready flag will *not* be modified even though for those case.

> * Error/event setting methods only set error and events, don't modify any other fields

>> agreed

>

I am good with the change in general, at least we need to document the behavior change in the original issue.

xing-yang: > This is already the logic, problem is once the ready == true, the controller does binding checks and based on the situation using this exact call to update status in the cases where binding is invalid (either content has been deleted or content has been modified to point to another snapshot) code. With this change, the ready flag will not be modified even though for those case.

Given this, I'm leaning towards adding a parameter "ready" in updateSnapshotStatusWithEvent to limit the scope of this change. This way we only change the behavior when snapshot class is not found.

I see that in the PV controller, it updates the phase from bound to available if volume.Spec.ClaimRef is nil.  There's a updateVolumePhaseWithEvent function which has phase and input parameter.

xing-yang: Updated updateSnapshotStatusWithEvent to include a "ready *bool" input parameter.  If nil, ReadyToUse will not be changed.  
If snapshot class is not found, ReadyToUse will not be changed.
If updateSnapshotStatus failed, ReadyToUse will not be changed.
For all other cases, the code stays the same as before, meaning ReadyToUse will be False.

pkg/common-controller/snapshot_controller.go
```diff
@@ -754,17 +759,17 @@ func (ctrl *csiSnapshotCommonController) storeContentUpdate(content interface{})
...
-// updateSnapshotStatusWithEvent saves new snapshot.Status to API server and emits
+// updateSnapshotErrorStatusWithEvent saves new snapshot.Status to API server and emits
 // given event on the snapshot. It saves the status and emits the event only when
 // the status has actually changed from the version saved in API server.
 // Parameters:
 //   snapshot - snapshot to update
 //   eventtype, reason, message - event to send, see EventRecorder.Event()
-func (ctrl *csiSnapshotCommonController) updateSnapshotErrorStatusWithEvent(snapshot *crdv1.VolumeSnapshot, eventtype, reason, message string) error {
-	klog.V(5).Infof("updateSnapshotStatusWithEvent[%s]", utils.SnapshotKey(snapshot))
+func (ctrl *csiSnapshotCommonController) updateSnapshotErrorStatusWithEvent(snapshot *crdv1.VolumeSnapshot, ready *bool, eventtype, reason, message string) error {
```
yuxiangqian, line 768: logic lgtm
I have a small suggestion on implementation.
Passing in a pointer of bool here seems to be a bit misleading, i.e., caller can actually can pass in a "true" into this call.

How about taking in a bool parameter, say "unreadySnapshot", this way its a bit more clear?

Please also update the parameters in comments.

xing-yang: Sure.  I named the bool parameter "setReadyToFalse" to be more explicit.  Addressed your comments.  PTAL.

xing-yang:
Hi @yuxiangqian, addressed your comments.  PTAL.  Thanks.

pkg/common-controller/snapshot_controller.go
```diff
@@ -465,7 +465,7 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
 		if _, err = ctrl.updateSnapshotStatus(snapshot, newContent); err != nil {
 			// update snapshot status failed
 			klog.V(4).Infof("failed to update snapshot %s status: %v", utils.SnapshotKey(snapshot), err)
-			ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotStatusUpdateFailed", fmt.Sprintf("Snapshot status update failed, %v", err))
+			ctrl.updateSnapshotErrorStatusWithEvent(snapshot, false, v1.EventTypeWarning, "SnapshotStatusUpdateFailed", fmt.Sprintf("Snapshot status update failed, %v", err))
```
yuxiangqian, line 468: should pass in "true" but it should not matter here

pkg/common-controller/snapshot_controller.go
```diff
@@ -497,20 +497,20 @@ func (ctrl *csiSnapshotCommonController) syncUnreadySnapshot(snapshot *crdv1.Vol
...
+		ctrl.updateSnapshotErrorStatusWithEvent(snapshot, true, v1.EventTypeWarning, "SnapshotContentCreationFailed", fmt.Sprintf("Failed to create snapshot content with error %v", err))
 		return err
 	}
 
 	// Update snapshot status with BoundVolumeSnapshotContentName
 	klog.V(5).Infof("syncUnreadySnapshot [%s]: trying to update snapshot status", utils.SnapshotKey(snapshot))
 	if _, err = ctrl.updateSnapshotStatus(snapshot, content); err != nil {
 		// update snapshot status failed
-		ctrl.updateSnapshotErrorStatusWithEvent(snapshot, v1.EventTypeWarning, "SnapshotStatusUpdateFailed", fmt.Sprintf("Snapshot status update failed, %v", err))
+		ctrl.updateSnapshotErrorStatusWithEvent(snapshot, false, v1.EventTypeWarning, "SnapshotStatusUpdateFailed", fmt.Sprintf("Snapshot status update failed, %v", err))
```
yuxiangqian, line 513: ditto

yuxiangqian:
two nits and LGTM

pkg/common-controller/snapshot_controller.go
xing-yang: I think this should pass in "false".  I added this case in the issue and PR description as well.  When updateSnapshotStatus fails, we should leave it alone and let next update snapshot status to do its job.

pkg/common-controller/snapshot_controller.go
xing-yang: For the same reason, this should be false as well.

Issue #445, opened 2020-12-09
Missing snapshot class causes panic in release-2.1 branch


mattcary:
#380 is active in release-2.1 (and maybe others).

The fix #381 can be backported.

/assign

mattcary:
I've confirmed that it is not a problem for release-2.0 (the bug was introduced in #275).

I haven't attempted to test 1.x, which uses the alpha API.

I use the following to test: apply it, and then see if the snapshot controller errors, in particular looking for a nil pointer dereference error in the logs. (note it depends on installing a standard-rwo StorageClass)

```
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: busy-pvc
spec:
  accessModes: [ "ReadWriteOnce" ]
  storageClassName: standard-rwo
  resources:
    requests:
      storage: 10Gi

---

apiVersion: v1
kind: Pod
metadata:
  name: busy
spec:
  containers:
   - name: busy
     image: busybox 
     args: ["sleep", "3600"]
     volumeMounts:
       - mountPath: /data/.test
         name: mypvc
  volumes:
   - name: mypvc
     persistentVolumeClaim:
       claimName: busy-pvc
       readOnly: false

---

apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshot
metadata:
  name: new-snapshot
spec:
  source:
    persistentVolumeClaimName: busy-pvc
```

mattcary:
/close

Fixed with v2.1.3.

PR #446, opened 2020-12-10
Cherrypick #381 to 2.1.

mattcary:
/kind bug

I've tested this manually on a cluster to confirm it fixes the crashlooping, as described in #445.

#445

```release-note
Backports fix #381 for crashloop when there are errors in the VolumeSnapshot, like a missing VolumeSnapshotClass.
```

### DIFF
--- a/pkg/common-controller/framework_test.go
+++ b/pkg/common-controller/framework_test.go
@@ -164,6 +164,21 @@ type reactorError struct {
 	error    error
 }
 
+// testError is an error returned from a test that marks a test as failed even
+// though the test case itself expected a common error (such as API error)
+type testError string
+
+func (t testError) Error() string {
+	return string(t)
+}
+
+var _ error = testError("foo")
+
+func isTestError(err error) bool {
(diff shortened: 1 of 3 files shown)


mattcary:
Hmm, I suspect these test failures are due to the v1 CRDs used on the test cluster. The k8s version would have to be pinned to 1.19 or other heroics done to get the v1beta1 CRDs required by this controller to be used.

@xing-yang do you have any ideas?

xing-yang:
Can you reference PR https://github.com/kubernetes-csi/external-snapshotter/pull/381 in the release note?

xing-yang:
We probably need to fix release-tools in release-2.1 to pass CI.  See this PR: https://github.com/kubernetes-csi/external-snapshotter/pull/437

mattcary:
> We probably need to fix release-tools in release-2.1 to pass CI. See this PR: #437

Ok, I'll look into backporting that too.

mattcary:
> Can you reference PR #381 in the release note?

done

PR #448, opened 2020-12-10
Update release-tools and prow script for release-2.1

mattcary:
/kind cleanup

This forks release-tools for the release-2.1 branch. The current version of release-tools updates the snapshot CRD directory to what's in the 3.0 release, but also updates building to use this repo when testing snapshots it will actually test the snapshot version found here.

Hence we can use neither the head of release-tools, nor an old version. This means we need a fork of some sort, and so have to disable the release-tools subtree test.

```release-note
NONE
```

### DIFF
--- a/release-tools/build.make
+++ b/release-tools/build.make
@@ -20,7 +20,7 @@
 
 # This is the default. It can be overridden in the main Makefile after
 # including build.make.
-REGISTRY_NAME=quay.io/k8scsi
+REGISTRY_NAME?=quay.io/k8scsi
 
 # Can be set to -mod=vendor to ensure that the "vendor" directory is used.
 GOFLAGS_VENDOR=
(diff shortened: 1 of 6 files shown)


mattcary:
/assign @xing-yang 

This is ready now and seems to work.

release-tools/build.make
```diff
@@ -241,8 +246,7 @@ test-vendor:
 .PHONY: test-subtree
 test: test-subtree
 test-subtree:
-	@ echo; echo "### $@:"
-	./release-tools/verify-subtree.sh release-tools
+	@ echo; echo "### $@: skipping test-subtree"
```
msau42, line 249: It's unfortunate we have to do this, but being able to test the PR version of the snapshot controller is important. Eventually, this branch will become unsupported so I don't anticipate future issues.

release-tools/prow.sh
```diff
@@ -85,7 +65,7 @@ get_versioned_variable () {
     echo "$value"
 }
 
-configvar CSI_PROW_BUILD_PLATFORMS "linux amd64; windows amd64 .exe; linux ppc64le -ppc64le; linux s390x -s390x; linux arm64 -arm64" "Go target platforms (= GOOS + GOARCH) and file suffix of the resulting binaries"
+configvar CSI_PROW_BUILD_PLATFORMS "linux amd64; windows amd64 .exe; linux ppc64le -ppc64le; linux arm64 -arm64" "Go target platforms (= GOOS + GOARCH) and file suffix of the resulting binaries"
```
msau42, line 68: I don't think we officially supported s390 in v2.x anyway, so this is ok

msau42: I think you can remove the release note about this because it's not a regression

mattcary: sg, done.

release-tools/prow.sh
```diff
@@ -335,7 +335,7 @@ default_csi_snapshotter_version () {
 	if [ "${CSI_PROW_KUBERNETES_VERSION}" = "latest" ] || [ "${CSI_PROW_DRIVER_CANARY}" = "canary" ]; then
 		echo "master"
 	else
-		echo "v3.0.2"
+		echo "v2.1.2"
```
xing-yang, line 338: So this would be a change only in 2.1 branch, right?  Can you add some comments to places with such special logic?

mattcary: Done. I cleaned up the commits as well so that the release-2.1 specific changes are all together in a single commit immediately following the 3.0 merge commit.

xing-yang:
/lgtm
/approve

Thanks @mattcary!

PR #439, opened 2020-12-04
Bump csi-lib-utils to v0.9.0

xing-yang:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Bump csi-lib-utils to v0.9.0 to get rid of klogv1 and get health check on leader election.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
* Added leader election health check at the metrics port + path "/healthz/leader-election".
* klog/v2 is used for logging
* `metrics-address` flag is deprecated and replaced by `http-endpoint`, which enables handlers from both metrics manager and leader election health check.
```

PR #453, opened 2020-12-16
Add changelog for release 4.0

xing-yang:
**What type of PR is this?**
/kind documentation

Fixes #

```release-note
NONE
```

### DIFF
--- a/CHANGELOG/CHANGELOG-4.0.md
+++ b/CHANGELOG/CHANGELOG-4.0.md
@@ -1,0 +1,276 @@
+# Release notes for v4.0.0
+
+[Documentation](https://kubernetes-csi.github.io)
+
+# Changelog since v3.0.0
+
+## Breaking Changes
+
+- Update to client/v4 ([#451](https://github.com/kubernetes-csi/external-snapshotter/pull/451), [@xing-yang](https://github.com/xing-yang))
+
+- Update snapshot-controller, CSI snapshotter sidecar, and validation webhook based on V1 APIs. This requires V1 snapshot APIs to be installed. ([#425](https://github.com/kubernetes-csi/external-snapshotter/pull/425), [@xing-yang](https://github.com/xing-yang))
+
+### Deprecations
+
+#### CSI Snapshotter Sidecar
+
+- `metrics-address` flag is deprecated and replaced by `http-endpoint`, which enables handlers from both metrics manager and leader election health check. ([#439](https://github.com/kubernetes-csi/external-snapshotter/pull/439), [@xing-yang](https://github.com/xing-yang))
+
+### Feature
+
+#### API Changes
+
+- Move Snapshot APIs to V1. ([#419](https://github.com/kubernetes-csi/external-snapshotter/pull/419), [@xing-yang](https://github.com/xing-yang))
+
+#### CSI Snapshotter Sidecar
+
+- Add `process_start_time_seconds` metric. ([#387](https://github.com/kubernetes-csi/external-snapshotter/pull/387), [@Jiawei0227](https://github.com/Jiawei0227))
+- The CSI snapshotter has a new `--extra-create-metadata` flag.
+  When set to true, the name and namespace of the source VolumeSnapshot, and the name of the source  VolumeSnapshotContent will be passed as parameters to the CSI driver in CreateSnapshotRequest. ([#375](https://github.com/kubernetes-csi/external-snapshotter/pull/375), [@chrishenzie](https://github.com/chrishenzie))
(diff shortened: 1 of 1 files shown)


CHANGELOG/CHANGELOG-4.0.md
```diff
@@ -0,0 +1,261 @@
...
+
+# Changelog since v3.0.0
+
+## Breaking Changes
+
+- Update to client/v4 ([#451](https://github.com/kubernetes-csi/external-snapshotter/pull/451), [@xing-yang](https://github.com/xing-yang))
+
+- Move Snapshot APIs to V1. ([#419](https://github.com/kubernetes-csi/external-snapshotter/pull/419), [@xing-yang](https://github.com/xing-yang))
+
+- Update snapshot-controller, CSI snapshotter sidecar, and validation webhook based on v1 apis. ([#425](https://github.com/kubernetes-csi/external-snapshotter/pull/425), [@xing-yang](https://github.com/xing-yang))
```
msau42, line 13: Note that this requires the v1 CRD to be installed in the cluster.

CHANGELOG/CHANGELOG-4.0.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Feature
+
+- Add `process_start_time_seconds` metric. ([#387](https://github.com/kubernetes-csi/external-snapshotter/pull/387), [@Jiawei0227](https://github.com/Jiawei0227))
+- The CSI snapshotter has a new `--extra-create-metadata` flag.
+  When set to true, the name and namespace of the source VolumeSnapshot, and the name of the source  VolumeSnapshotContent will be passed as parameters to the CSI driver in CreateSnapshotRequest. ([#375](https://github.com/kubernetes-csi/external-snapshotter/pull/375), [@chrishenzie](https://github.com/chrishenzie))
+- The `--http-endpoint` and `--metrics-path` flags have been added to support metrics for the snapshot-controller. To enable metrics, the user must configure a `--metrics-path` such as `:9102`. The default value for `--metrics-path` is the empty string, meaning this feature is disabled by default. The default value for `--metrics-path` is `/metrics`. ([#409](https://github.com/kubernetes-csi/external-snapshotter/pull/409), [@ggriffiths](https://github.com/ggriffiths))
```
msau42, line 20: We may need to merge this with the note for #439 below that changed the flag names and added the deprecation notice.

CHANGELOG/CHANGELOG-4.0.md
```diff
@@ -0,0 +1,261 @@
...
+
+[Documentation](https://kubernetes-csi.github.io)
+
+# Changelog since v3.0.0
+
+## Breaking Changes
+
+- Update to client/v4 ([#451](https://github.com/kubernetes-csi/external-snapshotter/pull/451), [@xing-yang](https://github.com/xing-yang))
+
+- Move Snapshot APIs to V1. ([#419](https://github.com/kubernetes-csi/external-snapshotter/pull/419), [@xing-yang](https://github.com/xing-yang))
```
msau42, line 11: Adding the v1 API is not breaking because we're still keeping v1beta right?

xing-yang: Right. It's only the controller changes which requires this API change.  Let me try to clarify that.

CHANGELOG/CHANGELOG-4.0.md
xing-yang: Added a note there.

CHANGELOG/CHANGELOG-4.0.md
xing-yang: 409 is for snapshot-controller while 439 is for the sidecar.  Let me separate them into different categories.

msau42: Hmmm I think we want snapshot-controller and the sidecar to also use consistent flag names. Can we quickly put together to make them use similar flag names?

xing-yang: They do use similar flag names.

Snapshot-controller just introduced metrics support.  So for snapshot-controller, there's no deprecation as it never used metrics-address.  It just directly uses http-endpoint and metrics-path.

For the sidecar,  metrics-address is deprecated and replaced by http-endpoint.  metrics-path is already there and we are not changing it.

xing-yang: I modified the changelog.  Hope it is more clear.  PTAL.

msau42: ah sorry i'm confusing myself with all the flags :-) Thanks for clarifying

PR #454, opened 2020-12-17
Update README for v4.0.0

xing-yang:
**What type of PR is this?**
/kind documentation

Update readme for v4.0.0.

Fixes #

```release-note
NONE
```


README.md
```diff
@@ -30,13 +22,13 @@ This information reflects the head of this branch.
 
 | Compatible with CSI Version                                                                | Container Image             | [Min K8s Version](https://kubernetes-csi.github.io/docs/kubernetes-compatibility.html#minimum-version) | Snapshot CRD version |
 | ------------------------------------------------------------------------------------------ | ----------------------------| --------------- | -------------------- |
-| [CSI Spec v1.2.0](https://github.com/container-storage-interface/spec/releases/tag/v1.2.0) | k8s.gcr.io/sig-storage/csi-snapshotter | 1.17         | v1beta1              |
-| [CSI Spec v1.2.0](https://github.com/container-storage-interface/spec/releases/tag/v1.2.0) | k8s.gcr.io/sig-storage/snapshot-controller  | 1.17     | v1beta1              |
-| [CSI Spec v1.2.0](https://github.com/container-storage-interface/spec/releases/tag/v1.2.0) | k8s.gcr.io/sig-storage/snapshot-validation-webhook  | 1.17     | v1beta1              |
+| [CSI Spec v1.2.0](https://github.com/container-storage-interface/spec/releases/tag/v1.2.0) | k8s.gcr.io/sig-storage/csi-snapshotter | 1.17         | v1beta1,v1             |
```
msau42, line 25: How do we want to define compatibility? The new controllers need the v1 API to be installed, but once it is, it will handle objects stored as v1beta

README.md
```diff
@@ -87,6 +82,14 @@ Install CSI Driver:
 
 The snapshot validating webhook is an HTTP callback which responds to [admission requests](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/). It is part of a larger [plan](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1900-volume-snapshot-validation-webhook) to tighten validation for volume snapshot objects. This webhook introduces the [ratcheting validation](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1900-volume-snapshot-validation-webhook#backwards-compatibility) mechanism targeting the tighter validation. The cluster admin or Kubernetes distribution admin should install the webhook alongside the snapshot controllers and CRDs.
 
+Along with the validation webhook, the volume snapshot controller will start [labeling](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1900-volume-snapshot-validation-webhook#automatic-labelling-of-invalid-objects) invalid snapshot objects which already existed. This is to enable quick identification of invalid snapshot objects in the system by running:
```
msau42, line 85: clarify this is for v1beta1 objects

xing-yang: For v1beta1, there's a check to make sure there is only one source in the snapshot and also snapshotClassName is not empty string.  Otherwise, invalid label will be applied.

For v1, there's a still a check to make sure snapshotClassName is not empty string and apply label accordingly.

I'll add these.

msau42: > For v1, there's a still a check to make sure snapshotClassName is not empty string and apply label accordingly.

This cannot be validated by the CRD schema?

README.md
xing-yang: maybe we'll just say "v1 controller requires v1 CRDs to be installed, but it supports both v1 and v1beta1 objects"?

msau42: sounds good!

README.md
```diff
@@ -87,6 +82,14 @@ Install CSI Driver:
...
+```
+kubectl get volumesnapshots --selector=snapshot.storage.kubernetes.io/invalid-snapshot-resource: ""
+kubectl get volumesnapshotcontents --selector=snapshot.storage.kubernetes.io/invalid-snapshot-content-resource: ""
+```
+
+Users should run this to identify, remove any invalid objects, and correct their workflows. Once the API has been switched to the v1 type, those invalid objects will not be deletable from the system.
```
xing-yang, line 91: Here I'll add that users should remove invalid objects before upgrading to v1.

README.md
xing-yang: No, this cannot.  Only the oneOf is validated by the CRD schema.  Xiangqian has verified that.

msau42: does the validation webhook validate it for v1 objects?  theoretically v1 objects should all be legal right?

Actually how does this work if the controller is reading only v1 objects now?

xing-yang: Yes, the webhook will still validate v1 objects.

If there are no existing invalid v1beta1 objects, after upgrading to v1, the webhook and schema validation will prevent the user from creating new invalid v1 and v1beta1 objects.

If there are existing invalid v1beta1 objects, the user should install the validation webhook before upgrading to v1 so that those invalid objects will be labeled and can be identified easily and removed before upgrading to v1.

If there are invalid v1beta1 objects and the user didn't install the validation webhook before upgrading to v1, those existing invalid v1beta1 objects will not be labeled (note: labeling is done by the controller, not the validation webhook).  The user can still manually identify them and delete them.

So the recommendation is that before upgrading to v1, the user should install the validation webhook so that all existing invalid objects will be labeled and can be easily identified and deleted.

PR #459, opened 2020-12-24
Provide better defaults for external-snapshotter

kvaps:
**What type of PR is this?**

/kind feature

Provided manifests for external-snapshotter are not optimal in most use-cases.

Since external-snapshotter is most common component among various Kubernetes distributions and csi-plugins, I think we need to correct the manifests and include the following changes. They are needed to simplify the setup and provide optimal defaults for every user.

- Deploy external-snapshotter into kube-system namespace by default
- Enable leader-election
- Use Deployment instead of Statefulset and deploy minimum 2 replicas for redundancy

The following changes would allow to simple install working external-snapshotter using just few commands, as described here:
https://kubernetes-csi.github.io/docs/snapshot-controller.html#deployment

but without having to manually edit manifests

Fixes #

```release-note
external-snapshotter manifests adjusted to reflect more common example
```

### DIFF
--- a/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
+++ b/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
@@ -9,13 +9,12 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: snapshot-controller
-  namespace: default # TODO: replace with the namespace you want for your controller, e.g. kube-system
+  namespace: kube-system
 
 ---
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  # rename if there are conflicts
   name: snapshot-controller-runner
 rules:
   - apiGroups: [""]
@@ -51,10 +50,9 @@ metadata:
 subjects:
   - kind: ServiceAccount
     name: snapshot-controller
-    namespace: default # TODO: replace with the namespace you want for your controller, e.g. kube-system
+    namespace: kube-system
 roleRef:
   kind: ClusterRole
-  # change the name also here if the ClusterRole gets renamed
   name: snapshot-controller-runner
   apiGroup: rbac.authorization.k8s.io
 
@@ -62,8 +60,8 @@ roleRef:
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  namespace: default # TODO: replace with the namespace you want for your controller, e.g. kube-system
   name: snapshot-controller-leaderelection
+  namespace: kube-system
 rules:
 - apiGroups: ["coordination.k8s.io"]
   resources: ["leases"]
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
```diff
@@ -9,13 +9,12 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: snapshot-controller
-  namespace: default # TODO: replace with the namespace you want for your controller, e.g. kube-system
+  namespace: kube-system
```
xing-yang, line 12: This is used in CI so I'd like to ask Patrick to take a look.
@pohly Do you have any concerns for this namespace to be changed from "default" to "kube-system"?

release-tools/prow.sh
```diff
@@ -801,11 +801,12 @@ install_snapshot_controller() {
...
   cnt=0
-  expected_running_pods=$(curl https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/"${CSI_SNAPSHOTTER_VERSION}"/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml | grep replicas | cut -d ':' -f 2-)
-  while [ "$(kubectl get pods -l app=snapshot-controller | grep 'Running' -c)" -lt "$expected_running_pods" ]; do
+  expected_running_pods=$(awk -F '[: ]+' '$1 == "" && $2 == "replicas" {print $3; exit}' "${CONTROLLER_DIR}/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml")
+  expected_namespace=$(awk -F '[: ]+' '$1 == "" && $2 == "namespace" {print $3; exit}' "${CONTROLLER_DIR}/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml")
+  while [ "$(kubectl get pods -n "$expected_namespace" -l app=snapshot-controller | grep 'Running' -c)" -lt "$expected_running_pods" ]; do
     if [ $cnt -gt 30 ]; then
         echo "snapshot-controller pod status:"
-        kubectl describe pods -l app=snapshot-controller
+        kubectl describe pods -n "$expected_namespace" -l app=snapshot-controller
```
xing-yang, line 809: prow.sh should be changed in csi-release-tools first.

deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
pohly: No objections.

Just beware that it might be necessary to change prow.sh so that it works with a rbac-snapshot-controller.yaml that has `namespace: default` (current stable release) and a newer rbac-snapshot-controller.yaml that has `namespace: kube-system` (canary) because the same prow.sh file is used with both, depending on the test job.

release-tools/prow.sh
kvaps: Sure, It has been changed already, PR https://github.com/kubernetes-csi/csi-release-tools/pull/121

xing-yang:
The release-tools change is already included in this PR: https://github.com/kubernetes-csi/external-snapshotter/pull/464.  Can you remove those changes from yours and rebase?

kvaps:
Sure, rebased

kvaps:
Hi. Is it possible to create release with this change?

xing-yang:
This affects multiple sidecars because we have to make changes in release-tools.  I don't think this is a bug fix that can be backported.  The yaml files are provided as examples.  User can modify their own yamls if needed.  Our usual release timeline aligns with the next k8s release.

kvaps:
Got it! Thank you

PR #471, opened 2021-02-08
[release-2.1] Fix release-tools/prow.sh for K8s 1.20

ialidzhikov:
/kind bug

This PR backports https://github.com/kubernetes-csi/csi-release-tools/commit/1d60e779 into `release-2.1` branch to fix the `git clone` issue in for `pull-kubernetes-csi-external-snapshotter-1-20-on-kubernetes-1-20`.

```release-note
NONE
```

### DIFF
--- a/.prow.sh
+++ b/.prow.sh
@@ -3,6 +3,7 @@
 # This is specific to the release-2.1 branch and overrides the
 # version set in the prow config.
 export CSI_SNAPSHOTTER_VERSION=v2.1.2
+export CSI_PROW_DRIVER_VERSION=v1.4.0
 
 . release-tools/prow.sh
 
--- a/release-tools/prow.sh
+++ b/release-tools/prow.sh
@@ -65,6 +65,18 @@ get_versioned_variable () {
     echo "$value"
 }
 
+# This takes a version string like CSI_PROW_KUBERNETES_VERSION and
+# maps it to the corresponding git tag, branch or commit.
+version_to_git () {
+    version="$1"
+    shift
+    case "$version" in
+        latest|master) echo "master";;
+        release-*) echo "$version";;
+        *) echo "v$version";;
+    esac
+}
+
 # Note that in the release-2.1 branch the s390x architecture is not supported.
 configvar CSI_PROW_BUILD_PLATFORMS "linux amd64; windows amd64 .exe; linux ppc64le -ppc64le; linux arm64 -arm64" "Go target platforms (= GOOS + GOARCH) and file suffix of the resulting binaries"
 
@@ -210,16 +222,7 @@ configvar CSI_PROW_DRIVER_CANARY_REGISTRY "gcr.io/k8s-staging-sig-storage" "regi
 # all generated files are present.
 #
 # CSI_PROW_E2E_REPO=none disables E2E testing.
-tag_from_version () {
-    version="$1"
-    shift
-    case "$version" in
-        latest) echo "master";;
-        release-*) echo "$version";;
-        *) echo "v$version";;
-    esac
-}
-configvar CSI_PROW_E2E_VERSION "$(tag_from_version "${CSI_PROW_KUBERNETES_VERSION}")"  "E2E version"
+configvar CSI_PROW_E2E_VERSION "$(version_to_git "${CSI_PROW_KUBERNETES_VERSION}")"  "E2E version"
 configvar CSI_PROW_E2E_REPO "https://github.com/kubernetes/kubernetes" "E2E repo"
 configvar CSI_PROW_E2E_IMPORT_PATH "k8s.io/kubernetes" "E2E package"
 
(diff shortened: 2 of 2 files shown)


pohly:
The problem in https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-csi-external-snapshotter-1-20-on-kubernetes-1-20/1358823154201399296 is a mismatch between the CRD that gets installed (https://github.com/kubernetes-csi/external-snapshotter/tree/release-2.1/config/crd = the one which supports only v1beta1 API) and the API that is expected by the csi-driver-host-path v1.5.0 deployment for 1.20 (https://github.com/kubernetes-csi/csi-driver-host-path/blob/v1.5.0/deploy/kubernetes-1.20/snapshotter/csi-hostpath-snapshotclass.yaml = v1).

Several options:
- keep testing release-2.1 with the older deployment files from csi-driver-host-path v1.4.0 (i.e. the ones for Kubernetes 1.18 = https://github.com/kubernetes-csi/csi-driver-host-path/tree/v1.4.0/deploy/kubernetes-1.18)
- update the CRDs in the release-2.1 branch

To do the former, we can either:
- remove the CSI_PROW_DRIVER_VERSION override from the Prow jobs
- override the override in .prow.sh of the release-2.1 branch

Conceptually, setting CSI_PROW_DRIVER_VERSION in the Prow job isn't necessary because prow.sh also has a default. I think @msau42 introduced it in the Prow config to avoid having to update individual repos.

mattcary:
/lgtm

This makes sense to me, it's essentially what we've done internally in GKE for testing deployments that are using the beta snapshot crds.

mattcary:
Hmm, the presubmit is still using the 1.5 prow csi driver version. I'm not sure how to fix that.

I guess it's a little weird we're testing the 2.1 release branch on 1.20 at all. We should only be testing >= v4.0?

ialidzhikov:
> Hmm, the presubmit is still using the 1.5 prow csi driver version. I'm not sure how to fix that.

Hm, from the raw output of the job I see `CSI driver version: CSI_PROW_DRIVER_VERSION=v1.4.0` which makes me think that the override thing works. But I am not sure what is presubmit and why it would still use v1.5.0.

pohly:
> remove the CSI_PROW_DRIVER_VERSION override from the Prow jobs sounds the much better option but I think that it would require setting the appropriate CSI_PROW_DRIVER_VERSION for each branch (I see that the current master defaults to CSI_PROW_DRIVER_VERSION=v1.3.0 if there is no override).

It's part of the TODO list in https://github.com/kubernetes-csi/csi-release-tools/pull/120 - with low priority, because it wouldn't really be used in practice while the Prow job config overrides it.

It's now still failing in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-snapshotter/471/pull-kubernetes-csi-external-snapshotter-1-20-on-kubernetes-1-20/1358873015818915840 because the E2E test in Kubernetes 1.20 is used and (I think) that one needs the v1 API.

I think the right fix is to skip testing of Kubernetes 1.20 in the .prow.sh script. Just return immediately. The old release branch was simply not meant to be tested on Kubernetes 1.20. Technically it works when using the v1beta1 API everywhere, but that's not what we recommend for Kubernetes 1.20.

I suggested today in the CSI standup meeting to add such a version check for "unknown" Kubernetes versions in advance. We agreed to *not* add it because we want to know when testing on a new Kubernetes version breaks. Also, in other repos it works.

I guess this repo is the exception where we have learned that testing on Kubernetes 1.20 doesn't work...

mattcary:
> I think the right fix is to skip testing of Kubernetes 1.20 in the .prow.sh script. Just return immediately. The old release branch was simply not meant to be tested on Kubernetes 1.20. Technically it works when using the v1beta1 API everywhere, but that's not what we recommend for Kubernetes 1.20.
...
> I guess this repo is the exception where we have learned that testing on Kubernetes 1.20 doesn't work...

+1, I don't think testing in 1.20 is appropriate.

ialidzhikov:
> > I think the right fix is to skip testing of Kubernetes 1.20 in the .prow.sh script. Just return immediately. The old release branch was simply not meant to be tested on Kubernetes 1.20. Technically it works when using the v1beta1 API everywhere, but that's not what we recommend for Kubernetes 1.20.
> > ...
> > I guess this repo is the exception where we have learned that testing on Kubernetes 1.20 doesn't work...
> 
> +1, I don't think testing in 1.20 is appropriate.

Yep, I see that the e2e tests under k/k try to use the v1 api which won't work for `release-2.1`.
Then I would simply suggest adapting the `skip_branches` for the `pull-kubernetes-csi-external-snapshotter-1-20-on-kubernetes-1-20` job - ref https://github.com/kubernetes/test-infra/blob/5e490580b8a60705d24c62079a1f65ec1531cf5b/config/jobs/kubernetes-csi/external-snapshotter/external-snapshotter-config.yaml#L186.

mattcary:
Would it be easier to to just skip in .prow.sh as @pohly suggeted (presumably using CSI_PROW_KUBERNETES_VERSION)? It seems easier to manage per-branch test configuration in the branch rather than having to also track it in the prow config.

ialidzhikov:
> Would it be easier to to just skip in .prow.sh as @pohly suggeted (presumably using CSI_PROW_KUBERNETES_VERSION)? It seems easier to manage per-branch test configuration in the branch rather than having to also track it in the prow config.

I believe with the k/test-infra approach the job won't be displayed at all on the PR (at least this is my assumption). With the artificial skip in each branch, I believe the job will be still visible in the PR and it will be wrongly reported as passed/green when there are no test executions in it. Feel free to correct me if I am wrong. Meanwhile I created https://github.com/kubernetes/test-infra/pull/20792. I hope that it is enough and it will do the trick.

mattcary:
> I believe with the k/test-infra approach the job won't be displayed at all on the PR (at least this is my assumption). With the artificial skip in each branch, I believe the job will be still visible in the PR and it will be wrongly reported as passed/green when there are no test executions in it.

Ah, that's a good point.

Fingers crossed for your test-infra change :-)

ialidzhikov:
Closing this PR in favour of https://github.com/kubernetes/test-infra/pull/20792.

/close

PR #480, opened 2021-03-10
Switch from update to patch for VolumeSnapshot and Content

### LINKED ISSUE #368: Use patch instead of update
To reduce the number of update failures due to resources being out of date, it is recommended to use patch instead of update. For updates where it's important that no other field in the object got modified, then you can include the object version in the patch.

ggriffiths:
**What type of PR is this?**
/kind bug

Switches to use patch instead of update

Fixes #368 

n/a

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/csi-snapshotter/rbac-csi-snapshotter.yaml
+++ b/deploy/kubernetes/csi-snapshotter/rbac-csi-snapshotter.yaml
@@ -35,10 +35,10 @@ rules:
     verbs: ["get", "list", "watch"]
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshotcontents"]
-    verbs: ["create", "get", "list", "watch", "update", "delete"]
+    verbs: ["create", "get", "list", "watch", "delete", "patch"]
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshotcontents/status"]
-    verbs: ["update"]
+    verbs: ["patch"]
 
 ---
 kind: ClusterRoleBinding
--- a/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
+++ b/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
@@ -34,13 +34,16 @@ rules:
     verbs: ["get", "list", "watch"]
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshotcontents"]
-    verbs: ["create", "get", "list", "watch", "update", "delete"]
+    verbs: ["create", "get", "list", "watch", "delete", "patch"]
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshotcontents/status"]
+    verbs: ["patch"]
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshots"]
-    verbs: ["get", "list", "watch", "update"]
+    verbs: ["get", "list", "watch", "patch"]
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshots/status"]
-    verbs: ["update"]
+    verbs: ["patch"]
 
 ---
 kind: ClusterRoleBinding
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.15
 
 require (
 	github.com/container-storage-interface/spec v1.3.0
+	github.com/evanphx/json-patch v4.9.0+incompatible
 	github.com/fsnotify/fsnotify v1.4.9
 	github.com/go-logr/logr v0.3.0 // indirect
 	github.com/golang/mock v1.4.4
--- a/pkg/common-controller/snapshot_create_test.go
+++ b/pkg/common-controller/snapshot_create_test.go
@@ -26,7 +26,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var timeNow = time.Now()
+var timeNow = time.Now().Round(time.Second)
 var timeNowStamp = timeNow.UnixNano()
 var False = false
 var True = true
@@ -161,10 +161,10 @@ func TestCreateSnapshotSync(t *testing.T) {
 			initialClaims:     newClaimArray("claim7-9", "pvc-uid7-9", "1Gi", "volume7-9", v1.ClaimBound, &classGold),
 			initialVolumes:    newVolumeArray("volume7-9", "pv-uid7-9", "pv-handle7-9", "1Gi", "pvc-uid7-9", "claim7-9", v1.VolumeBound, v1.PersistentVolumeReclaimDelete, classGold),
 			errors: []reactorError{
-				{"update", "volumesnapshots", errors.New("mock update error")},
-				{"update", "volumesnapshots", errors.New("mock update error")},
-				{"update", "volumesnapshots", errors.New("mock update error")},
-				{"update", "volumesnapshots", errors.New("mock update error")},
+				{"patch", "volumesnapshots", errors.New("mock update error")},
+				{"patch", "volumesnapshots", errors.New("mock update error")},
+				{"patch", "volumesnapshots", errors.New("mock update error")},
+				{"patch", "volumesnapshots", errors.New("mock update error")},
 			},
 			expectSuccess: false,
 			test:          testSyncSnapshot,
--- a/pkg/sidecar-controller/snapshot_controller.go
+++ b/pkg/sidecar-controller/snapshot_controller.go
@@ -156,7 +156,7 @@ func (ctrl *csiSnapshotSideCarController) updateContentErrorStatusWithEvent(cont
 	}
 	ready := false
 	contentClone.Status.ReadyToUse = &ready
-	newContent, err := ctrl.clientset.SnapshotV1().VolumeSnapshotContents().UpdateStatus(context.TODO(), contentClone, metav1.UpdateOptions{})
+	newContent, err := utils.PatchVolumeSnapshotContent(content, contentClone, ctrl.clientset, utils.StatusSubResource)
 
 	// Emit the event even if the status update fails so that user can see the error
 	ctrl.eventRecorder.Event(newContent, eventtype, reason, message)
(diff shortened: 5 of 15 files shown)


ggriffiths:
Discovered that stategicmergepatch is not supported for CRs. Working now on an alternative

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.15
 
 require (
 	github.com/container-storage-interface/spec v1.3.0
+	github.com/evanphx/json-patch v4.9.0+incompatible
```
ggriffiths, line 7: Used in kubernetes/kubernetes:
https://github.com/kubernetes/kubernetes/blob/master/go.mod#L42

pkg/common-controller/snapshot_create_test.go
```diff
@@ -26,7 +26,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var timeNow = time.Now()
+var timeNow = time.Date(2021, time.January, 1, 0, 0, 0, 0, time.Local)
```
ggriffiths, line 29: Was running into unit test issues without this.
```
        -                       CreationTime:                   s"2021-04-01 17:45:55.275503069 -0700 PDT m=+0.021010896",
        +                       CreationTime:                   s"2021-04-01 17:45:55 -0700 PDT",
```

The patching behavior seems to treat timestamps with less precision.

xing-yang: hmm, interesting

ggriffiths: I think the precision is only for unit tests - I just tried on a cluster and the `creationTime` seems to not have the fractional seconds above.

```
  status:
    boundVolumeSnapshotContentName: snapcontent-6ac333b4-25c5-49d9-a55e-5a7cdc8db7f3
    creationTime: "2021-04-02T01:03:11Z"
    readyToUse: true
```

ggriffiths: Changed this to be `time.Now().Round(time.Second)` instead.

deploy/kubernetes/csi-snapshotter/rbac-csi-snapshotter.yaml
```diff
@@ -35,10 +35,10 @@ rules:
     verbs: ["get", "list", "watch"]
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshotcontents"]
-    verbs: ["create", "get", "list", "watch", "update", "delete"]
+    verbs: ["create", "get", "list", "watch", "update", "delete", "patch"]
```
xing-yang, line 38: Do we still need "update"?

deploy/kubernetes/csi-snapshotter/rbac-csi-snapshotter.yaml
```diff
@@ -35,10 +35,10 @@ rules:
...
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshotcontents/status"]
-    verbs: ["update"]
+    verbs: ["update", "patch"]
```
xing-yang, line 41: Same question about "update"?

deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
```diff
@@ -34,13 +34,16 @@ rules:
     verbs: ["get", "list", "watch"]
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshotcontents"]
-    verbs: ["create", "get", "list", "watch", "update", "delete"]
+    verbs: ["create", "get", "list", "watch", "update", "delete", "patch"]
```
xing-yang, line 37: Still need "update"?

deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
```diff
@@ -34,13 +34,16 @@ rules:
...
+  - apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshotcontents/status"]
+    verbs: ["patch"]
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshots"]
-    verbs: ["get", "list", "watch", "update"]
+    verbs: ["get", "list", "watch", "update", "patch"]
```
xing-yang, line 43: Still need "update"?

deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
```diff
@@ -34,13 +34,16 @@ rules:
...
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshots/status"]
-    verbs: ["update"]
+    verbs: ["update", "patch"]
```
xing-yang, line 46: Still need "update"?

pkg/utils/patch_test.go
```diff
@@ -0,0 +1,194 @@
...
+			updatedSnapshot: &crdv1.VolumeSnapshot{
+				Status: &crdv1.VolumeSnapshotStatus{
+					BoundVolumeSnapshotContentName: nil,
+					CreationTime:                   nil,
+					ReadyToUse:                     nil,
+					RestoreSize:                    nil,
+					Error:                          nil,
+				},
+			},
+			expectedPatch: `{"status":{"boundVolumeSnapshotContentName":null}}`,
```
xing-yang, line 56: Should it be "nil"?

pkg/utils/patch_test.go
```diff
@@ -0,0 +1,194 @@
...
+				},
+			},
+			updatedSnapshotContent: &crdv1.VolumeSnapshotContent{
+				Status: &crdv1.VolumeSnapshotContentStatus{
+					SnapshotHandle: nil,
+					CreationTime:   nil,
+					ReadyToUse:     nil,
+					RestoreSize:    nil,
+					Error:          nil,
+				},
```
xing-yang, line 146: Add some tests where the updated status has non-nil CreationTime, ReadyToUse, RestoreSize.

pkg/utils/patch_test.go
```diff
@@ -0,0 +1,194 @@
...
+					Error:                          nil,
+				},
+			},
+			updatedSnapshot: &crdv1.VolumeSnapshot{
+				Status: &crdv1.VolumeSnapshotStatus{
+					BoundVolumeSnapshotContentName: nil,
+					CreationTime:                   nil,
+					ReadyToUse:                     nil,
+					RestoreSize:                    nil,
+					Error:                          nil,
```
xing-yang, line 53: Add some tests where the updated status has non-nil CreationTime, ReadyToUse, RestoreSize.

deploy/kubernetes/csi-snapshotter/rbac-csi-snapshotter.yaml
ggriffiths: Removed these update rule.

pkg/utils/patch_test.go
ggriffiths: The way merge patch works is that it accepts a json object with all of the fields that are being changed. `null` json value marshals to `nil` in Go.

the createSnapshotPatch/createContentPatch funcs create json objects like this.

pkg/utils/patch_test.go
ggriffiths: Added these tests

pkg/utils/patch_test.go
ggriffiths: Added these tests

ggriffiths:
Addressed comments, ready for re-review @xing-yang

pkg/sidecar-controller/snapshot_controller.go
```diff
@@ -370,13 +370,14 @@ func (ctrl *csiSnapshotSideCarController) clearVolumeContentStatus(
...
-		content.Status.RestoreSize = nil
+	contentClone := content.DeepCopy()
+	if contentClone.Status != nil {
+		contentClone.Status.SnapshotHandle = nil
+		contentClone.Status.ReadyToUse = nil
+		contentClone.Status.CreationTime = nil
+		contentClone.Status.RestoreSize = nil
 	}
-	newContent, err := ctrl.clientset.SnapshotV1().VolumeSnapshotContents().UpdateStatus(context.TODO(), content, metav1.UpdateOptions{})
+	newContent, err := utils.PatchVolumeSnapshotContent(content, contentClone, ctrl.clientset, "status")
```
xing-yang, line 380: This string "status" is used in multiple places.  Can you make it a constant?

pkg/utils/patch_test.go
xing-yang: Sure.  Can you add this as comments here in the code?

pkg/utils/patch.go
```diff
@@ -42,12 +42,12 @@ func PatchVolumeSnapshot(
...
 	}
 
-	newSnapshot, err := client.SnapshotV1().VolumeSnapshots(updatedSnapshot.Namespace).Patch(context.TODO(), existingSnapshot.Name, types.MergePatchType, patch, metav1.PatchOptions{}, subresources...)
+	_, err = client.SnapshotV1().VolumeSnapshots(updatedSnapshot.Namespace).Patch(context.TODO(), existingSnapshot.Name, types.MergePatchType, patch, metav1.PatchOptions{}, subresources...)
 	if err != nil {
 		return updatedSnapshot, err
 	}
 
-	return newSnapshot, nil
+	return updatedSnapshot, nil
```
xing-yang, line 50: I thought "newSnapshot" is the latest one returned from API server?  Why returning "updatedSnapshot"?

ggriffiths: this was a test commit - I was trying to see if this reduced the 'modified object' error count.

ggriffiths: Reverting the above change as it did not lower the error count.

xing-yang: Can you take a look of the logs and see if you can tell why we got the "modified object" error?

ggriffiths: Asked SIG API machinery per your suggestion to try and find out. I'll also check the logs to see under what scenarios we face this. The error counts are fairly consistent.

ggriffiths:
Still investigating ways to lower the error count. merge patch alone with this implementation does not seem to lower the error count.
/hold

ggriffiths:
Findings so far:
1. Strategicmergepatch is not supported for CRDs. Moving to `mergepatch` via `json-patch` does not lower error count (~250 error count)
2. Removing `addResourceVersion` from implementation does not lower error count (error count: 240)
3. Performing a "Get" before every "Patch" lowered error count to 24 total, but is expensive.

ggriffiths:
As this PR did not actually decrease the "object has been modified" error count, I am closing this for now.

I still plan to investigate how this "object has been modified" problem can be solved.

Issue #489, opened 2021-04-02
Support for in-tree volumes migrated with CSIMigration feature-gate


mitchellhuang:
It appears that snapshot-controller does not support snapshotting volumes that have been migrated to CSI with the CSIMigration/CSIMigrationAWS feature-gate. When attempting to snapshot a migrated volume it will error:

```
cannot find CSI PersistentVolumeSource for volume pvc-b9e3ba60-6dd1-4f73-94e4-04db040598a7
```

You can see [here](https://github.com/kubernetes-csi/external-snapshotter/blob/533ed88/pkg/common-controller/snapshot_controller.go#L650) where the error occurs because the migrated PV spec does not contain `Spec.CSI.VolumeHandle`. It is instead located at `Spec.AwsElasticBlockStore.VolumeID`.

Is supporting migrated in-tree volumes within scope of this project? It seems like we can support migrated volumes using CSI translation methods like `TranslateInTreePVToCSI` found in the the main kubernetes [source code](https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/csi-translation-lib/plugins/aws_ebs.go#L124).

Migrated In-tree PV Spec:

```
  spec:
    accessModes:
    - ReadWriteOnce
    awsElasticBlockStore:
      fsType: ext4
      volumeID: vol-010b07de49ba5b821
    capacity:
      storage: 10Gi
    claimRef:
      apiVersion: v1
      kind: PersistentVolumeClaim
      name: nginx
      namespace: default
      resourceVersion: "11167"
      uid: 7ef0cf5c-425a-4a56-87de-39c84ac4df1a
    nodeAffinity:
      required:
        nodeSelectorTerms:
        - matchExpressions:
          - key: topology.ebs.csi.aws.com/zone
            operator: In
            values:
            - eu-west-1a
    persistentVolumeReclaimPolicy: Delete
    storageClassName: standard
    volumeMode: Filesystem
```

CSI PV Spec:

```
  spec:
    accessModes:
    - ReadWriteOnce
    capacity:
      storage: 50Gi
    claimRef:
      apiVersion: v1
      kind: PersistentVolumeClaim
      name: prometheus-storage-prometheus-prometheus-0
      namespace: kube-system
      resourceVersion: "3755"
      uid: b9e3ba60-6dd1-4f73-94e4-04db040598a7
    csi:
      driver: ebs.csi.aws.com
[...]

PR #490, opened 2021-04-05
Add support for snapshotting migrated in-tree volumes

mitchellhuang:
**What type of PR is this?**
/kind feature

* Add support for snapshotting in-tree volumes migrated by CSIMigration feature-gate in Kubernetes
* Maintain support for existing native CSI volumes
* Use csi-translation-lib for translation https://github.com/kubernetes/csi-translation-lib

Fixes https://github.com/kubernetes-csi/external-snapshotter/issues/489

n/a

```release-note
NONE
```


xing-yang:
Hi @mitchellhuang, CSI migration is intended to support features that are already supported by in-tree drivers. It is not intended to add new CSI features.

renan:
Any progress on this?

It would awesome to have this feature after having migrated a bunch of volumes to CSI and not really wanting to recreate those volumes, maybe even having downtime.

xing-yang:
There's no change here. CSI migration is intended to support features that are already supported by in-tree drivers. It is not intended to add new CSI features.

renan:
> When a Kubernetes cluster administrator updates a cluster to enable CSI migration, existing stateful deployments and workloads continue to function as they always have; however, behind the scenes Kubernetes hands control of all storage management operations (previously targeting in-tree drivers) to CSI drivers.
> https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-feature-csi-migration-beta/#what-is-csi-migration

As I understood, once a volume has been migrated to a CSI driver the volume should be treated as a CSI volume. Hence it should be possible to create snapshots for it since it is now backed up by a CSI driver. Am I missing something?

xing-yang:
See this new blog on CSI migration. It specifically calls out that new CSI features such as volume snapshotting is not supported for CSI migrated volumes.
https://kubernetes.io/blog/2021/12/10/storage-in-tree-to-csi-migration-status-update/#what-should-i-do-as-a-user

renan:
That clarifies things. Thank you!

humblec:
@mitchellhuang @xing-yang  yeah, I was also in an impression that we could support this functionality. The reason being many volumes are migrated and also the user get the choice to stick or use new CSI SC based on their requirement. so, appreciated if we could revalidate about this..  :+1:

msau42:
CSI migration is intended as a way to provide backwards compatibility for the intree  API when we remove the built-in cloud providers. It was not meant to provide forward compatibility to new features.

I understand how convenient it would be to not have to recreate volumes using the CSI API to get CSI functionality, but I also worry about the precedent this will set and additional burden this is going to place on all storage feature developers and maintainers of the intree plugins if every single feature we develop in the future needs to be designed, implemented, tested, and supported to handle csi migration.

PR #508, opened 2021-04-30
Automated cherry pick of #504: Make sure the VolumeSnapshots v1 CRDs exist before starting #505: Update sample implementation #507: fix typo in CRD check

mauriciopoppe:
Cherry pick of #504 #505 #507 on release-4.0.

#504: Make sure the VolumeSnapshots v1 CRDs exist before starting
#505: Update sample implementation
#507: fix typo in CRD check

For details on the cherry pick process, see the [cherry pick requests](https://git.k8s.io/community/contributors/devel/sig-release/cherry-picks.md) page.

```release-note
Add check for v1 CRDs to allow for rolling update of the snapshot-controller
```

### DIFF
--- a/cmd/snapshot-controller/main.go
+++ b/cmd/snapshot-controller/main.go
@@ -29,6 +29,10 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
 	klog "k8s.io/klog/v2"
 
 	"github.com/kubernetes-csi/csi-lib-utils/leaderelection"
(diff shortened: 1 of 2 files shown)


xing-yang:
I see the following errors.  This is caused by the changes in the snapshot controller deployment file.  Probably the new changes in StatefulSet API is not recognizable by an earlier version of K8S cluster.

```
Deploying snapshot-controller
   /home/prow/go/src/github.com/kubernetes-csi/external-snapshotter/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml
        using           image: snapshot-controller:csiprow
error: error validating "STDIN": error validating data: [ValidationError(StatefulSet.spec): unknown field "minReadySeconds" in io.k8s.api.apps.v1.StatefulSetSpec, ValidationError(StatefulSet.spec): unknown field "strategy" in io.k8s.api.apps.v1.StatefulSetSpec]; if you choose to ignore these errors, turn validation off with --validate=false
```

mauriciopoppe:
I think this is failing because #459 wasn't cherrypicked to the release branch, I read this comment too #459 (comment), what do you suggest?

xing-yang:
I know what the problem is now.  In release-4.0, we are using "statefulset" to deploy snapshot-controller.  In master, we are using "deployment".  So the new fields you added to "setup-snapshot-controller.yaml " are not valid fields on release-4.0 branch.

I'm not sure that we can backport a change that needs to change from "statefulset" to "deployment".

mauriciopoppe:
if we can't backport #459 then I can probably just cherrypick the changes on the startup code and not the manifests, what do you think?

xing-yang:
Hi @mauriciopoppe, v4.1.0 is just released.

Looks like StatefulSet also has an UpdateStrategy field.  Do you want to take a look and see if similar changes can be applied?
https://github.com/kubernetes/kubernetes/blob/v1.22.0-alpha.1/pkg/apis/apps/types.go#L153

mauriciopoppe:
Thanks for the suggestion @xing-yang, I've looked at this field and it has this shape:

```yaml
updateStrategy:
  type: "RollingUpdate" | "OnDelete"
  rollingUpdate:
    partition: int32
```

Unfortunately for the fix in #504 to work we rely on `minReadySeconds`, however because there's a new release I'm going to use that instead.

PR #497, opened 2021-04-15
Fix snapshotter metrics server to use httpEndpoint when provided

ggriffiths:
Signed-off-by: Grant Griffiths <ggriffiths@purestorage.com>

**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Cherry-pick of #496 

**Which issue(s) this PR fixes**:
n/a

**Special notes for your reviewer**:
n/a

**Does this PR introduce a user-facing change?**:
```release-note
--http-endpoint will now correctly be used for the metrics server address when --metrics-address is not provided.
```

PR #511, opened 2021-05-01
Cherry-picks of #409 and #439: Adding --http-endpoint to snapshot components

chrishenzie:
**What type of PR is this?**
> /kind feature

This PR contains cherry-picks of #409 and #439, which each provide an `--http-endpoint` flag for the snapshot-controller and csi-snapshotter.

This is already present in v4; backporting so v3 users can make use of leader election health checks and have a consistent `--http-endpoint` flag across the snapshot-controller and csi-snapshotter.

```release-note
The `--http-endpoint` and `--metrics-path` flags have been added to support metrics for the snapshot-controller. To enable metrics, the user must configure a `--metrics-path` such as `:9102`. The default value for `--metrics-path` is the empty string, meaning this feature is disabled by default. The default value for `--metrics-path` is `/metrics`.
Added leader election health check at the metrics port + path "/healthz/leader-election".
klog/v2 is used for logging
The `--metrics-address` flag is deprecated and replaced by `http-endpoint`, which enables handlers from both metrics manager and leader election health check.
```

/assign @xing-yang 

I copied over the release notes from the two PRs.

### DIFF
--- a/cmd/csi-snapshotter/main.go
+++ b/cmd/csi-snapshotter/main.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"net/http"
 	"os"
 	"os/signal"
 	"strings"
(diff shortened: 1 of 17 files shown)


cmd/csi-snapshotter/main.go
```diff
@@ -134,8 +145,20 @@ func main() {
...
-	metricsManager.StartMetricsEndpoint(*metricsAddress, *metricsPath)
+
+	// Prepare http endpoint for metrics + leader election healthz
+	mux := http.NewServeMux()
+	if addr != "" {
+		metricsManager.RegisterToServer(mux, *metricsPath)
+		metricsManager.SetDriverName(driverName)
+		go func() {
+			klog.Infof("ServeMux listening at %q", addr)
+			err := http.ListenAndServe(*metricsAddress, mux)
```
xing-yang, line 156: The following PR should be cherry-picked as well:
https://github.com/kubernetes-csi/external-snapshotter/pull/497

It's fine to submit a follow-up for that.

PR #519, opened 2021-05-21
Add gauge metric for snapshot controller operations in flight

### LINKED ISSUE #434: Add metric for detecting metrics operation leaks in snapshot controller
It is possible that operation leaks can occur if logic changes happen in the snapshot controller. We should add a metric to count how many snapshot operations are cached. If this count does not decrease after any operation is finished, we have an operation cache leak.

_Originally mentioned by @msau42 in https://github.com/kubernetes-csi/external-snapshotter/pull/409#discussion_r532979019_

ggriffiths:

**What type of PR is this?**
/kind feature

Adds an operations in flight metrics to find any leaky operations in our metrics implementation. If a snapshot is not in progress, and in flight metrics are >0, we have a leak. I've done some basic testing with this metric and haven't found any leaked operations yet.

Fixes #434 

I decided to change the sync.Map to be a normal map with a RWMutex. This is because sync.Map does not allow the user to find its length (golang/go issue 20680).

The original issue calls to check for any built-in way to determine if we have leaked operations. Except for some pprof memory profiling, I didn't see an easy way to do this via go tooling. This metric will allow us to easily tell if there's a leak.

```release-note
Adds an in-operations_in_flight metric for determining the number of snapshot operations in progress.
```

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -37,6 +37,8 @@ const (
 	subSystem                     = "snapshot_controller"
 	operationLatencyMetricName    = "operation_total_seconds"
 	operationLatencyMetricHelpMsg = "Total number of seconds spent by the controller on an operation"
+	operationInFlightName         = "operations_in_flight"
+	operationInFlightHelpMsg      = "Total number of operations in flight"
 	unknownDriverName             = "unknown"
 
 	// CreateSnapshotOperationName is the operation that tracks how long the controller takes to create a snapshot.
@@ -72,6 +74,10 @@ const (
 	// SnapshotStatusTypeCancel represents that a CreateSnapshot, CreateSnapshotAndReady,
 	// or DeleteSnapshot has been deleted before finishing.
 	SnapshotStatusTypeCancel snapshotStatusType = "cancel"
+)
+
+var (
+	inFlightCheckInterval = 30 * time.Second
 )
 
 // OperationStatus is the interface type for representing an operation's execution
@@ -152,19 +158,25 @@ type operationMetricsManager struct {
 	// ongoing operations.
 	// key is an Operation
 	// value is the timestamp of the start time of the operation
-	cache sync.Map
+	cache map[OperationKey]OperationValue
+
+	// mutex for protecting cache from concurrent access
+	mu sync.Mutex
 
 	// registry is a wrapper around Prometheus Registry
 	registry k8smetrics.KubeRegistry
 
 	// opLatencyMetrics is a Histogram metrics for operation time per request
 	opLatencyMetrics *k8smetrics.HistogramVec
+
+	// opInFlight is a Gauge metric for the number of operations in flight
+	opInFlight *k8smetrics.Gauge
 }
 
 // NewMetricsManager creates a new MetricsManager instance
 func NewMetricsManager() MetricsManager {
 	mgr := &operationMetricsManager{
-		cache: sync.Map{},
+		cache: make(map[OperationKey]OperationValue),
 	}
 	mgr.init()
 	return mgr
@@ -172,27 +184,31 @@ func NewMetricsManager() MetricsManager {
 
 // OperationStart starts a new operation
 func (opMgr *operationMetricsManager) OperationStart(key OperationKey, val OperationValue) {
-	val.startTime = time.Now()
-	opMgr.cache.LoadOrStore(key, val)
+	opMgr.mu.Lock()
+	defer opMgr.mu.Unlock()
+
+	if _, exists := opMgr.cache[key]; !exists {
+		val.startTime = time.Now()
+		opMgr.cache[key] = val
+	}
+	opMgr.opInFlight.Set(float64(len(opMgr.cache)))
 }
 
 // OperationStart drops an operation
 func (opMgr *operationMetricsManager) DropOperation(op OperationKey) {
-	opMgr.cache.Delete(op)
+	opMgr.mu.Lock()
+	defer opMgr.mu.Unlock()
+	delete(opMgr.cache, op)
+	opMgr.opInFlight.Set(float64(len(opMgr.cache)))
 }
 
 // RecordMetrics emits operation metrics
 func (opMgr *operationMetricsManager) RecordMetrics(opKey OperationKey, opStatus OperationStatus, driverName string) {
-	obj, exists := opMgr.cache.Load(opKey)
+	opMgr.mu.Lock()
+	defer opMgr.mu.Unlock()
+	opVal, exists := opMgr.cache[opKey]
 	if !exists {
 		// the operation has not been cached, return directly
-		return
-	}
-	opVal, ok := obj.(OperationValue)
-	if !ok {
-		// the cached item is not a OperationValue, should NEVER happen, clean and return
-		klog.Errorf("Invalid cache entry for key %v", opKey)
-		opMgr.cache.Delete(opKey)
 		return
 	}
 	status := string(SnapshotStatusTypeUnknown)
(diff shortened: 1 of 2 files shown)


pkg/metrics/metrics.go
```diff
@@ -261,6 +280,29 @@ func (opMgr *operationMetricsManager) init() {
 		[]string{labelDriverName, labelOperationName, labelSnapshotType, labelOperationStatus},
 	)
 	opMgr.registry.MustRegister(opMgr.opLatencyMetrics)
+	opMgr.opInFlight = k8smetrics.NewGauge(
+		&k8smetrics.GaugeOpts{
+			Subsystem: subSystem,
+			Name:      operationInFlightName,
+			Help:      operationInFlightHelpMsg,
+		},
```
yuxiangqian, line 288: it might be helpful to also add labels like DriverName

pkg/metrics/metrics_test.go
```diff
@@ -478,6 +478,87 @@ snapshot_controller_operation_total_seconds_count{driver_name="driver5",operatio
...
+
+	if err := verifyInFlightMetric(`snapshot_controller_operations_in_flight 2`, srvAddr); err != nil {
+		t.Errorf("failed testing [%v]", err)
+	}
+
+	//  Record, should be down to 1
+	mgr.RecordMetrics(opKey, nil, "driver")
+	time.Sleep(250 * time.Millisecond)
+
+	if err := verifyInFlightMetric(`# HELP snapshot_controller_operations_in_flight [ALPHA] Total number of operations in flight
```
yuxiangqian, line 518: why this string is different from line 510 & 536?

pkg/metrics/metrics_test.go
```diff
@@ -478,6 +478,87 @@ snapshot_controller_operation_total_seconds_count{driver_name="driver5",operatio
...
+	mgr.OperationStart(opKey, opVal)
+	time.Sleep(250 * time.Millisecond)
+
+	if err := verifyInFlightMetric(`snapshot_controller_operations_in_flight 2`, srvAddr); err != nil {
+		t.Errorf("failed testing [%v]", err)
+	}
+
+	//  Record, should be down to 1
+	mgr.RecordMetrics(opKey, nil, "driver")
+	time.Sleep(250 * time.Millisecond)
```
yuxiangqian, line 516: this can cause flaky tests easily, can we avoid it?

pkg/metrics/metrics_test.go
```diff
@@ -613,6 +695,14 @@ func containsMetrics(expectedMfs, gotMfs []*cmg.MetricFamily) bool {
 						fmt.Println("difference in counter values")
 						continue
 					}
+
+				case got.Metric[i].Gauge != nil && expected.Metric[j].Gauge != nil:
```
yuxiangqian, line 699: I do not see "expected" change?

pkg/metrics/metrics.go
```diff
@@ -152,49 +158,58 @@ type operationMetricsManager struct {
...
 	return mgr
 }
 
 // OperationStart starts a new operation
 func (opMgr *operationMetricsManager) OperationStart(key OperationKey, val OperationValue) {
+	opMgr.mu.Lock()
+	defer opMgr.mu.Unlock()
+
 	val.startTime = time.Now()
-	opMgr.cache.LoadOrStore(key, val)
```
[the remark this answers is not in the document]
ggriffiths: Yes, unfortunately it is not possible to get the length of a sync.Map. There are ways to do this with map.Range, but it seemed expensive for every operation.

I will change OperationStart to preserve the existing behavior of `LoadOrStore`.

pkg/metrics/metrics_test.go
ggriffiths: If you're referring to the time.Sleep, I think most of the existing tests depend on this already and I haven't noticed any flakes. I've increased this to 500 milliseconds for now.

pkg/metrics/metrics_test.go
ggriffiths: What do you mean by this? I just changed one of the test expectations from 51 to 52 and it failed as expected. Am I missing something here?

pkg/metrics/metrics.go
ggriffiths: The goal of this metric, per #434, is mainly for us to see if any leaks are happening. The metric counts all operations in transit regardless of driver name/labels.

There are some cases where we won't have driver name. i.e. in the scheduled go-routine and in DropOperation.

pkg/metrics/metrics_test.go
ggriffiths: Fixed this

pkg/metrics/metrics.go
```diff
@@ -246,7 +264,8 @@ func (opMgr *operationMetricsManager) recordCancelMetric(obj interface{}, key Op
 		val.SnapshotType,
 		string(SnapshotStatusTypeCancel),
 	).Observe(duration)
-	opMgr.cache.Delete(key)
+	delete(opMgr.cache, key)
+	opMgr.opInFlight.Set(float64(len(opMgr.cache)))
```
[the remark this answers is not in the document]
ggriffiths: Removed this

yuxiangqian (commented):
nits and LGTM, thanks @ggriffiths

pkg/metrics/metrics_test.go
ggriffiths: I spent some time trying to consolidate verifyInFlightMetric into verifyMetrics, but I found some unexpected test framework behavior that made it difficult to integrate the two. `verifyInFlightMetric` works well for this specific test.

ggriffiths: The main issues were the following:
1. `verifyMetrics` silently passes when an expected metric is not formatted correctly
2. `sortMfs` returns an empty slice always, which causes test validation to pass incorrectly
3. ran into some nil pointers when getting `sortMfs` to work with two different metric types.

I can open an issue to update the existing `verifyMetrics`. Was hoping to get this PR in with `verifyInFlightMetrics` for now.

ggriffiths:
@yuxiangqian ready for review

ggriffiths:
@yuxiangqian @xing-yang any additional work needed for this PR?

xing-yang:
@yuxiangqian Do you have more comments on this PR?

yuxiangqian (commented):
nit. and LGTM

pkg/metrics/metrics.go
```diff
@@ -213,31 +229,34 @@ func (opMgr *operationMetricsManager) RecordMetrics(opKey OperationKey, opStatus
...
 		}
 	}
 
-	opMgr.cache.Delete(opKey)
+	delete(opMgr.cache, opKey)
+	opMgr.opInFlight.Set(float64(len(opMgr.cache)))
 }
 
 // recordCancelMetric records a metric for a create operation that hasn't finished
 func (opMgr *operationMetricsManager) recordCancelMetric(obj interface{}, key OperationKey, duration float64) {
```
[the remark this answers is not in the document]
ggriffiths: Addressed this

pkg/metrics/metrics.go
```diff
@@ -152,49 +158,59 @@ type operationMetricsManager struct {
 	// ongoing operations.
 	// key is an Operation
 	// value is the timestamp of the start time of the operation
-	cache sync.Map
+	cache map[OperationKey]OperationValue
+
+	// mutex for protecting cache from concurrent access
+	mu sync.RWMutex
```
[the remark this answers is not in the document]
ggriffiths: Moved this to a sync.Mutex.

yuxiangqian (commented):
/lgtm

PR #524, opened 2021-05-25
Retrieve the VolumeSnapshotContent before attempting an update

huffmanca:
**What type of PR is this?**
/kind bug

It's been noticed that when we attempt to update the VSC's error the following error can be thrown:

```
I0525 13:07:49.293034       1 snapshot_controller.go:165] updating VolumeSnapshotContent[snapcontent-f24fd046-1664-4bb2-bb43-f91ae0909b02] error status failed Operation cannot be fulfilled on volumesnapshotcontents.snapshot.storage.k8s.io "snapcontent-f24fd046-1664-4bb2-bb43-f91ae0909b02": the object has been modified; please apply your changes to the latest version and try again
```

This PR forces a Get before we attempt the update so that we can avoid this error.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/sidecar-controller/snapshot_controller.go
+++ b/pkg/sidecar-controller/snapshot_controller.go
@@ -140,11 +140,17 @@ func (ctrl *csiSnapshotSideCarController) checkandUpdateContentStatus(content *c
 func (ctrl *csiSnapshotSideCarController) updateContentErrorStatusWithEvent(content *crdv1.VolumeSnapshotContent, eventtype, reason, message string) error {
 	klog.V(5).Infof("updateContentStatusWithEvent[%s]", content.Name)
 
+	latestContent, err := ctrl.clientset.SnapshotV1().VolumeSnapshotContents().Get(context.TODO(), content.Name, metav1.GetOptions{})
+	if err != nil {
+		klog.V(4).Infof("updateContentStatusWithEvent[%s]: error obtaining volumeSnapshotContent %v", content.Name, err)
+		return nil
+	}
+
 	if content.Status != nil && content.Status.Error != nil && *content.Status.Error.Message == message {
 		klog.V(4).Infof("updateContentStatusWithEvent[%s]: the same error %v is already set", content.Name, content.Status.Error)
 		return nil
 	}
-	contentClone := content.DeepCopy()
+	contentClone := latestContent.DeepCopy()
 	if contentClone.Status == nil {
 		contentClone.Status = &crdv1.VolumeSnapshotContentStatus{}
 	}


pkg/sidecar-controller/snapshot_controller.go
```diff
@@ -140,11 +140,17 @@ func (ctrl *csiSnapshotSideCarController) checkandUpdateContentStatus(content *c
 func (ctrl *csiSnapshotSideCarController) updateContentErrorStatusWithEvent(content *crdv1.VolumeSnapshotContent, eventtype, reason, message string) error {
 	klog.V(5).Infof("updateContentStatusWithEvent[%s]", content.Name)
 
+	updatedContent, err := ctrl.clientset.SnapshotV1().VolumeSnapshotContents().Get(context.TODO(), content.Name, metav1.GetOptions{})
```
ggriffiths, line 143: nit: latestContent might be more clear? In the context of `updateContentErrorStatusWithEvent `, I think `updatedContent` would imply that the content status has been updated with an event.

ggriffiths:
These seems to cut the number of errors in half from 56 to 28:

with this change:
```
~ curl -s https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-snapshotter/524/pull-kubernetes-csi-external-snapshotter-1-21-on-kubernetes-1-21/1397246208530976768/artifacts/kube-system/snapshot-controller-58f54d8f6c-7t6tx/snapshot-controller.log | grep 'the object has been modified; please apply your' | wc -l
      28
```

without:
```
~ curl -s https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-snapshotter/519/pull-kubernetes-csi-external-snapshotter-1-21-on-kubernetes-1-21/1396873247164731392/artifacts/kube-system/snapshot-controller-58f54d8f6c-9gt8c/snapshot-controller.log |  grep 'the object has been modified; please apply your' | wc -l
      56
```

I believe there was a concern about excessive API calls if we get the latest object before every update, but this will only pull the latest object for `updateContentErrorStatusWithEvent`. I think it's okay to add this change as a temporary improvement until a permanent fix is added. Thoughts on this @xing-yang ?

pkg/sidecar-controller/snapshot_controller.go
huffmanca: Valid. I've updated the name.

xing-yang:
> I believe there was a concern about excessive API calls if we get the latest object before every update, but this will only pull the latest object for updateContentErrorStatusWithEvent. I think it's okay to add this change as a temporary improvement until a permanent fix is added. Thoughts on this @xing-yang ?

I still have this concern.  Let's discuss about this in the CSI meeting.

pkg/sidecar-controller/snapshot_controller.go
```diff
@@ -140,11 +140,17 @@ func (ctrl *csiSnapshotSideCarController) checkandUpdateContentStatus(content *c
 func (ctrl *csiSnapshotSideCarController) updateContentErrorStatusWithEvent(content *crdv1.VolumeSnapshotContent, eventtype, reason, message string) error {
 	klog.V(5).Infof("updateContentStatusWithEvent[%s]", content.Name)
 
+	latestContent, err := ctrl.clientset.SnapshotV1().VolumeSnapshotContents().Get(context.TODO(), content.Name, metav1.GetOptions{})
+	if err != nil {
+		klog.V(4).Infof("updateContentStatusWithEvent[%s]: error obtaining volumeSnapshotContent %v", err)
```
xing-yang, line 145: Content name is missing to fill in %s.

PR #526, opened 2021-05-26
Use JSON patch for many VolumeSnapshot and VolumeSnapshotContent updates

### LINKED ISSUE #525: VolumeSnapshotContent update fails with "object has been modified"
Seeing this consistently on snapshotting e2e tests in the snapshot controller pod:

```
~ curl -s https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-snapshotter/519/pull-kubernetes-csi-external-snapshotter-1-21-on-kubernetes-1-21/1396873247164731392/artifacts/kube-system/snapshot-controller-58f54d8f6c-9gt8c/snapshot-controller.log |  grep 'the object has been modified; please apply your' | wc -l
      56
```

There seems to be multiple root causes for this error. We'll be tracking them here:
- [x] [updateContentErrorStatusWithEvent](https://github.com/kubernetes-csi/external-snapshotter/blob/679e71e89d4644ae4b50fa71874b4d9a77bb9e5a/pkg/sidecar-controller/snapshot_controller.go#L140) - ~~if we perform a Get before Update here, the errors seem to drop by half.~~ Will not do this initially as it will lower performance.
- [x] [functions not using or returning updatedContent after update calls](https://github.com/kubernetes-csi/external-snapshotter/blob/8751eb0ca5b33a8a3c2589d506b48cc26684e0cd/pkg/sidecar-controller/snapshot_controller.go#L599) - various functions do not propagate their `updatedContent` back up the stack to be cached properly (see below for explanation). https://github.com/kubernetes-csi/external-snapshotter/pull/527
- [ ] JSONPatch can also be used to lower the error count - https://github.com/kubernetes-csi/external-snapshotter/pull/526

ggriffiths:
Signed-off-by: Grant Griffiths <ggriffiths@purestorage.com>

**What type of PR is this?**
/kind bug

This PR addresses the "object has been modified" issue we see a lot in the snapshot-controller/snapshotter.

Before this PR, we were seeing ~30-50 occurrences of this error in the snapshot-controller/snapshotter logs.

This PR only addresses the problematic areas found in the e2e PR tests. If we continue to see more of these errors, we can    address replace the remaining Update/UpdateStatus calls in a separate PR.

Fixes #525 

n/a

```release-note
Replaces many VolumeSnapshot/VolumeSnapshotContent Update/UpdateStatus operations with Patch. This lowers the probability of the "object has been modified" update API errors occurring. This change introduces a dependency on two new RBAC rules for the snapshotter: volumesnapshotcontents:patch, volumesnapshotcontents/status:patch and four new RBAC rules for the snapshot-controller: volumesnapshotcontents:patch, volumesnapshotcontents/status:patch, volumesnapshots:patch, and volumesnapshots/status: patch.
```

### DIFF
--- a/pkg/common-controller/framework_test.go
+++ b/pkg/common-controller/framework_test.go
@@ -354,12 +354,6 @@ func (r *snapshotReactor) React(action core.Action) (handled bool, ret runtime.O
 
 			storedVer, _ := strconv.Atoi(storedSnapshot.ResourceVersion)
 			storedSnapshot.ResourceVersion = strconv.Itoa(storedVer + 1)
-
-			// // If we were updating annotations and the new annotations are nil, leave as empty.
-			// // This seems to be the behavior for merge-patching nil & empty annotations
-			// if !reflect.DeepEqual(storedSnapshotContent.Annotations, content.Annotations) && content.Annotations == nil {
-			// 	content.Annotations = make(map[string]string)
-			// }
 		} else {
 			return true, nil, fmt.Errorf("cannot update snapshot %s: snapshot not found", action.GetName())
 		}
--- a/pkg/common-controller/framework_test.go
+++ b/pkg/common-controller/framework_test.go
@@ -326,9 +326,7 @@ func (r *snapshotReactor) React(action core.Action) (handled bool, ret runtime.O
 		return true, snapshot, nil
 
 	case action.Matches("patch", "volumesnapshots"):
-		snapshot := &crdv1.VolumeSnapshot{}
 		action := action.(core.PatchAction)
-
 		// Check and bump object version
 		storedSnapshot, found := r.snapshots[action.GetName()]
 		if found {
@@ -359,9 +357,10 @@ func (r *snapshotReactor) React(action core.Action) (handled bool, ret runtime.O
 		}
 
 		// Store the updated object to appropriate places.
-		r.snapshots[snapshot.Name] = storedSnapshot
+		r.snapshots[storedSnapshot.Name] = storedSnapshot
 		r.changedObjects = append(r.changedObjects, storedSnapshot)
 		r.changedSinceLastSync++
+
 		klog.V(4).Infof("saved updated snapshot %s", storedSnapshot.Name)
 		return true, storedSnapshot, nil
 
@@ -1253,6 +1252,10 @@ func testAddSnapshotFinalizer(ctrl *csiSnapshotCommonController, reactor *snapsh
 	return ctrl.addSnapshotFinalizer(test.initialSnapshots[0], true, true)
 }
 
+func testAddSingleSnapshotFinalizer(ctrl *csiSnapshotCommonController, reactor *snapshotReactor, test controllerTest) error {
+	return ctrl.addSnapshotFinalizer(test.initialSnapshots[0], false, true)
+}
+
 func testRemoveSnapshotFinalizer(ctrl *csiSnapshotCommonController, reactor *snapshotReactor, test controllerTest) error {
 	return ctrl.removeSnapshotFinalizer(test.initialSnapshots[0], true, true)
 }
@@ -1510,8 +1513,10 @@ func evaluateFinalizerTests(ctrl *csiSnapshotCommonController, reactor *snapshot
 		if funcName == "testAddSnapshotFinalizer" {
 			for _, snapshot := range reactor.snapshots {
 				if test.initialSnapshots[0].Name == snapshot.Name {
-					if !utils.ContainsString(test.initialSnapshots[0].ObjectMeta.Finalizers, utils.VolumeSnapshotBoundFinalizer) && utils.ContainsString(snapshot.ObjectMeta.Finalizers, utils.VolumeSnapshotBoundFinalizer) &&
-						!utils.ContainsString(test.initialSnapshots[0].ObjectMeta.Finalizers, utils.VolumeSnapshotAsSourceFinalizer) && utils.ContainsString(snapshot.ObjectMeta.Finalizers, utils.VolumeSnapshotAsSourceFinalizer) {
+					if !utils.ContainsString(test.initialSnapshots[0].ObjectMeta.Finalizers, utils.VolumeSnapshotBoundFinalizer) &&
+						utils.ContainsString(snapshot.ObjectMeta.Finalizers, utils.VolumeSnapshotBoundFinalizer) &&
+						!utils.ContainsString(test.initialSnapshots[0].ObjectMeta.Finalizers, utils.VolumeSnapshotAsSourceFinalizer) &&
+						utils.ContainsString(snapshot.ObjectMeta.Finalizers, utils.VolumeSnapshotAsSourceFinalizer) {
 						klog.V(4).Infof("test %q succeeded. Finalizers are added to snapshot %s", test.name, snapshot.Name)
 						bHasSnapshotFinalizer = true
 					}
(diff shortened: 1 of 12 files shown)


ggriffiths:
50 error occurrences with MergePatchType as expected with ccfe3c7
```
$ curl -s https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-snapshotter/526/pull-kubernetes-csi-external-snapshotter-1-21-on-kubernetes-1-21/1397628215157067776/artifacts/kube-system/snapshot-controller-58f54d8f6c-btxgf/snapshot-controller.log | grep "object has been modified" | wc -l
50
```

Trying with JSONPatch as suggested

ggriffiths:
JSONPatch seems to bring this count down for the one area I added it (updateContentErrorStatusWithEvent)
```
$ curl -s https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-snapshotter/526/pull-kubernetes-csi-external-snapshotter-1-21-on-kubernetes-1-21/1397649661208039424/artifacts/kube-system/snapshot-controller-58f54d8f6c-s46br/snapshot-controller.log | grep "object has been modified" | wc -l
34
```

ggriffiths:
Made some good progress on this PR. Down to the following numbers:
**snapshot-controller**: 4-8 occurrences of "object has been modified" per test (down from ~50)
**snapshotter**: 16 occurrences of "object has been modified" per test (down from ~50)

I have an idea of where the remaining errors are coming from, so I'll focus on those first. We'll likely have a mix of "Patch" and "Update" calls until all Update calls have been replaced on a per-need basis. Each update call needs to be hand-replaced and validated with a corresponding `kubectl patch` command. The logic is different for every operation. See controller.go for some examples. cc @xing-yang

ggriffiths:
@xing-yang the latest PR tests have zero occurrences of the `object has been modified` API errors! Ready for review.

Thanks @bswartz for the JSON patch suggestion!

xing-yang:
Thanks @ggriffiths!  This is great progress!

pkg/sidecar-controller/snapshot_controller.go
```diff
@@ -580,12 +587,19 @@ func (ctrl *csiSnapshotSideCarController) setAnnVolumeSnapshotBeingCreated(conte
 	contentClone := content.DeepCopy()
 	metav1.SetMetaDataAnnotation(&contentClone.ObjectMeta, utils.AnnVolumeSnapshotBeingCreated, "yes")
 
-	updatedContent, err := ctrl.clientset.SnapshotV1().VolumeSnapshotContents().Update(context.TODO(), contentClone, metav1.UpdateOptions{})
+	var patches []utils.PatchOp
+	patches = append(patches, utils.PatchOp{
+		Op:    "replace",
+		Path:  "/metadata/annotations",
+		Value: contentClone.ObjectMeta.GetAnnotations(),
```
xing-yang, line 594: Since we only need the annotations part, can we call content.ObjectMeta.GetAnnotations() first, save it, and then append the new utils.AnnVolumeSnapshotBeingCreated="yes" annotation to it.  So we don't have to do a DeepCopy on line 587.

pkg/common-controller/framework_test.go
```diff
@@ -284,6 +325,52 @@ func (r *snapshotReactor) React(action core.Action) (handled bool, ret runtime.O
...
+			}
+
+			storedVer, _ := strconv.Atoi(storedSnapshot.ResourceVersion)
+			storedSnapshot.ResourceVersion = strconv.Itoa(storedVer + 1)
+
+			// // If we were updating annotations and the new annotations are nil, leave as empty.
+			// // This seems to be the behavior for merge-patching nil & empty annotations
+			// if !reflect.DeepEqual(storedSnapshotContent.Annotations, content.Annotations) && content.Annotations == nil {
+			// 	content.Annotations = make(map[string]string)
+			// }
```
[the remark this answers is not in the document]
ggriffiths: This was a solution for merge-patch unit tests. This isn't relevant for JSON patch. Will remove this.

pkg/sidecar-controller/snapshot_controller.go
xing-yang: Or may be do something similar to what you did with finalizers?  https://github.com/kubernetes-csi/external-snapshotter/pull/526/files#diff-1a3473f9016f7abcf5007414ff7141e39032aaf7aba079eeef262d04167208a2R813-R828

pkg/common-controller/snapshot_controller.go
```diff
@@ -1392,24 +1418,53 @@ func isControllerUpdateFailError(err *crdv1.VolumeSnapshotError) bool {
...
-	if addBoundFinalizer {
-		snapshotClone.ObjectMeta.Finalizers = append(snapshotClone.ObjectMeta.Finalizers, utils.VolumeSnapshotBoundFinalizer)
-	}
-	newSnapshot, err := ctrl.clientset.SnapshotV1().VolumeSnapshots(snapshotClone.Namespace).Update(context.TODO(), snapshotClone, metav1.UpdateOptions{})
-	if err != nil {
-		return newControllerUpdateError(utils.SnapshotKey(snapshot), err.Error())
+	var updatedSnapshot *crdv1.VolumeSnapshot
+	var err error
+
+	// Must perform an update if no finalizers exist
```
xing-yang, line 1424: Do you know why we must perform an update if no finalizers exist?

pkg/common-controller/snapshot_controller.go
```diff
@@ -1489,14 +1544,21 @@ func (ctrl *csiSnapshotCommonController) setAnnVolumeSnapshotBeingDeleted(conten
 	// Set AnnVolumeSnapshotBeingDeleted if it is not set yet
 	if !metav1.HasAnnotation(content.ObjectMeta, utils.AnnVolumeSnapshotBeingDeleted) {
 		klog.V(5).Infof("setAnnVolumeSnapshotBeingDeleted: set annotation [%s] on content [%s].", utils.AnnVolumeSnapshotBeingDeleted, content.Name)
+		var patches []utils.PatchOp
 		metav1.SetMetaDataAnnotation(&content.ObjectMeta, utils.AnnVolumeSnapshotBeingDeleted, "yes")
+		patches = append(patches, utils.PatchOp{
+			Op:    "replace",
+			Path:  "/metadata/annotations",
```
xing-yang, line 1551: Does this work if there were no annotations initially?

pkg/common-controller/snapshot_controller.go
ggriffiths: good catch, this was from before I found out about the "replace" patch operation with a new array. We can probably do that here instead. I did this with the content finalizers: https://github.com/kubernetes-csi/external-snapshotter/pull/526/files/0ccf8017800f0f6160f5003f2cf333954dfa4a91#diff-1a3473f9016f7abcf5007414ff7141e39032aaf7aba079eeef262d04167208a2R813-R830

pkg/common-controller/snapshot_controller.go
ggriffiths: Yes, tested this here:
```
grant@colo-70-0-0-129:~/workspace/kubeup$ k get volumesnapshotcontent snapcontent-52f8092e-148f-4747-8f57-f1f21eaed3d4 -o yaml | grep annotations
grant@colo-70-0-0-129:~/workspace/kubeup$ k patch volumesnapshotcontent snapcontent-52f8092e-148f-4747-8f57-f1f21eaed3d4 --type='json' -p='[{"op": "replace", "path": "/metadata/annotations", "value": {"snapshot.storage.kubernetes.io/volumesnapshot-being-deleted": "yes"}}]'
volumesnapshotcontent.snapshot.storage.k8s.io/snapcontent-52f8092e-148f-4747-8f57-f1f21eaed3d4 patched
grant@colo-70-0-0-129:~/workspace/kubeup$ k get volumesnapshotcontent snapcontent-52f8092e-148f-4747-8f57-f1f21eaed3d4 -o yaml | grep -A 2 -B 2 annotations
kind: VolumeSnapshotContent
metadata:
  annotations:
    snapshot.storage.kubernetes.io/volumesnapshot-being-deleted: "yes"
  creationTimestamp: "2021-09-24T02:15:45Z"
--
    fieldsV1:
      f:metadata:
        f:annotations:
          .: {}
          f:snapshot.storage.kubernetes.io/volumesnapshot-being-deleted: {}

```

pkg/sidecar-controller/snapshot_controller.go
ggriffiths: Good idea, addressed this by creating a new annotations map. I wanted to avoid editing the existing content object, as it's passed up the stack and used when we update the status error.

ggriffiths:
Will squash these commits once an initial LGTM is given. Smallers commits have been useful while testing whether or not individual patch additions break the build.

pkg/common-controller/snapshot_controller.go
ggriffiths: I just tried this, but patch seems to keep failing for this case. It's pretty hard to debug/find out why, so I would prefer to keep this as Update unless we start seeing errors related to this section. I just updated the PR to reflect this.

I've added a unit test for the patch case where we add a single finalizer, and that seems to work.

ggriffiths:
Ready for re-review @xing-yang

pkg/common-controller/snapshot_controller.go
ggriffiths: added this

xing-yang:
Good job @ggriffiths!

/lgtm
/approve

PR #527, opened 2021-05-27
Return VolumeSnapshotContent from various functions instead of nil

ggriffiths:
Signed-off-by: Grant Griffiths <ggriffiths@purestorage.com>

**What type of PR is this?**
/kind bug

When removeAnnVolumeSnapshotBeingCreated fails, updatedContent is not returned.

Part of #525 

This does not completely fix the problem, but it does seem to lower the count.

```release-note
NONE
```

### DIFF
--- a/pkg/sidecar-controller/snapshot_controller.go
+++ b/pkg/sidecar-controller/snapshot_controller.go
@@ -79,9 +79,11 @@ func (ctrl *csiSnapshotSideCarController) syncContent(content *crdv1.VolumeSnaps
 	// already true. We don't want to keep calling CreateSnapshot
 	// or ListSnapshots CSI methods over and over again for
 	// performance reasons.
+	var err error
 	if content.Status != nil && content.Status.ReadyToUse != nil && *content.Status.ReadyToUse == true {
 		// Try to remove AnnVolumeSnapshotBeingCreated if it is not removed yet for some reason
-		return ctrl.removeAnnVolumeSnapshotBeingCreated(content)
+		_, err = ctrl.removeAnnVolumeSnapshotBeingCreated(content)
+		return err
 	}
 	return ctrl.checkandUpdateContentStatus(content)
 }
@@ -101,7 +103,7 @@ func (ctrl *csiSnapshotSideCarController) createSnapshot(content *crdv1.VolumeSn
 	klog.V(5).Infof("createSnapshot for content [%s]: started", content.Name)
 	contentObj, err := ctrl.createSnapshotWrapper(content)
 	if err != nil {
-		ctrl.updateContentErrorStatusWithEvent(content, v1.EventTypeWarning, "SnapshotCreationFailed", fmt.Sprintf("Failed to create snapshot: %v", err))
+		ctrl.updateContentErrorStatusWithEvent(contentObj, v1.EventTypeWarning, "SnapshotCreationFailed", fmt.Sprintf("Failed to create snapshot: %v", err))
 		klog.Errorf("createSnapshot for content [%s]: error occurred in createSnapshotWrapper: %v", content.Name, err)
 		return err
 	}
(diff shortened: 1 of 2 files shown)


ggriffiths:
@jsafrane I have fixed the issue you mentioned with `removeAnnVolumeSnapshotBeingCreated`. Let me know what you think.

This seems to lower the "object has been modified" count a bit, but it's still not consistently low yet. Either way, I think this is probably heading the right direction.

jsafrane:
It's a step in the right direction. Every time the sidecar issues `Update` / `UpdateStatus` against the API server, it must use the returned objects (with updated ResourceVersion) in all subsequent `Update` / `UpdateStatus` calls.

* `setAnnVolumeSnapshotBeingCreated` must return `content` it got from the API server
* Whole `createSnapshotWrapper` should not return `nil`, but some `content`.
* When it calls `createSnapshotWrapper`, it should use the returned object in subsequent `updateContentErrorStatusWithEvent` and not the original one, i.e. use `contentObj` instead of `content` in both these places:
  * https://github.com/kubernetes-csi/external-snapshotter/blob/8751eb0ca5b33a8a3c2589d506b48cc26684e0cd/pkg/sidecar-controller/snapshot_controller.go#L104
  * https://github.com/kubernetes-csi/external-snapshotter/blob/8751eb0ca5b33a8a3c2589d506b48cc26684e0cd/pkg/sidecar-controller/snapshot_controller.go#L121
 
And please check all error paths for returning `nil` from `createSnapshotWrapper` / `checkandUpdateContentStatusOperation`.

ggriffiths:
Ok, I plan to update this PR soon with the above feedback.

ggriffiths:
@jsafrane ready for re-review, I've addressed your comments above.

I've also done the following:
* Searched for `return nil,` and added a content/snapshot object if possible
* Searched for all Update/UpdateStatus calls and made sure we were returning the updated objects.

I still see around the same number "object has been modified" error count in our PR tests, but I think these changes are good to do regardless.

xing-yang:
@ggriffiths are you saying we don't see the "object has been modified" count getting reduced at all with changes in this PR?

ggriffiths:
@xing-yang I wrote that comment before the final build ran. Up until then, the error count seemed to not be lowered after each commit.

However, with the most recent commit, it seems to be lowered:

1.19 - 28 occurrences of object has been modified - [link](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-snapshotter/527/pull-kubernetes-csi-external-snapshotter-1-19-on-kubernetes-1-19/1402771395213004800/artifacts/kube-system/snapshot-controller-58f54d8f6c-nr7s7/snapshot-controller.log)
1.20 - 42 occurrences of object has been modified - [link](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-snapshotter/527/pull-kubernetes-csi-external-snapshotter-1-20-on-kubernetes-1-20/1402771395229782016/artifacts/kube-system/snapshot-controller-58f54d8f6c-rl6zh/snapshot-controller.log)
1.21 - 26 occurrences of object has been modified - [link](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-snapshotter/527/pull-kubernetes-csi-external-snapshotter-1-21-on-kubernetes-1-21/1402771395821178880/artifacts/kube-system/snapshot-controller-58f54d8f6c-9dstp/snapshot-controller.log)

pkg/sidecar-controller/snapshot_controller.go
```diff
@@ -300,12 +302,12 @@ func (ctrl *csiSnapshotSideCarController) createSnapshotWrapper(content *crdv1.V
 		// storage system has responded with an error
 		klog.Infof("createSnapshotWrapper: CreateSnapshot for content %s returned error: %v", content.Name, err)
 		if isCSIFinalError(err) {
-			if removeAnnotationErr := ctrl.removeAnnVolumeSnapshotBeingCreated(content); removeAnnotationErr != nil {
-				return nil, fmt.Errorf("failed to remove VolumeSnapshotBeingCreated annotation from the content %s: %s", content.Name, removeAnnotationErr)
+			if content, removeAnnotationErr := ctrl.removeAnnVolumeSnapshotBeingCreated(content); removeAnnotationErr != nil {
```
jsafrane, line 305: This declares a new variable `content`, shadowing the existing one. `return` at line 310 below therefore returns the **old** `content`.

With following code, I don't see any "`the object has been modified`":

```
var removeAnnotationErr error
if content, removeAnnotationErr = ctrl.removeAnnVolumeSnapshotBeingCreated(...)
```

jsafrane:
I found one issue in the PR, but the rest looks good to me and it reduces nr. of conflicts, at least in my test case (taking snapshot of  a non-existing CSI hostpath volume - the driver constantly returns `NotFound`).

pkg/sidecar-controller/snapshot_controller.go
ggriffiths: good catch, let me address this

ggriffiths: Addressed this

jsafrane:
/lgtm
Thanks a lot for this PR!

Issue #349, opened 2020-08-11
VolumeSnapshot cannot be deleted


venkatsc:
Kubernetes version: v1.18.5
CSI snapshotter version: v2.1.1

ouptput of `kubectl get volumesnapshot`
```log
NAME                              READYTOUSE   SOURCEPVC          SOURCESNAPSHOTCONTENT   RESTORESIZE   SNAPSHOTCLASS               SNAPSHOTCONTENT                                    CREATIONTIME   AGE
test-snapshot-dynamic-provision   false        quobyte-csi-test                                         quobyte-csi-snapshotclass   snapcontent-e5af5c7e-89bd-42a4-bd5d-e00f205f243c                  14h
```
events from `kubectl describe volumesnapshot`
```log
Events:
  Type     Reason                         Age   From                 Message
  ----     ------                         ----  ----                 -------
  Warning  SnapshotContentCreationFailed  54m   snapshot-controller  Failed to create snapshot content with error snapshot controller failed to update quobyte-csi-test on API server: cannot add finalizer on claim because it is being deleted
```

logs from `snapshot-controller`
```log
I0811 09:58:46.942919       1 snapshot_controller.go:893] Keeping PVC default/quobyte-csi-test, it is used by snapshot default/test-snapshot-dynamic-provision
I0811 09:58:46.942935       1 snapshot_controller.go:229] PVC default/quobyte-csi-test is being used by snapshot default/test-snapshot-dynamic-provision as source
I0811 09:58:46.942945       1 snapshot_controller.go:609] createSnapshotContent: Creating content for snapshot default/test-snapshot-dynamic-provision through the plugin ...
E0811 09:58:46.942953       1 snapshot_controller.go:838] cannot add finalizer on claim [quobyte-csi-test] for snapshot [test-snapshot-dynamic-provision]: claim is being deleted
E0811 09:58:46.942957       1 snapshot_controller.go:615] createSnapshotContent failed to add finalizer for source of snapshot snapshot controller failed to update quobyte-csi-test on API server: cannot add finalizer on claim because it is being deleted
E0811 09:58:46.942973       1 snapshot_controller_base.go:378] could not sync volume "default/test-snapshot-dynamic-provision": snapshot controller failed to update quobyte-csi-test on API server: cannot add finalizer on claim because it is being deleted
[...]

PR #360, opened 2020-08-17
Fix the race between PVC finalizer and snapshot finalizer removal

xing-yang:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #349 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
Fix a problem deleting the PVC finalizer and snapshot finalizer.
```

PR #534, opened 2021-06-17
Fix the race between PVC finalizer and snapshot finalizer removal (2.1)

tareqhs:
**What type of PR is this?**
/kind bug

Fixes https://github.com/kubernetes-csi/external-snapshotter/issues/349
cherry-pick https://github.com/kubernetes-csi/external-snapshotter/pull/360


xing-yang:
@tareqhs release-2.1 branch is no longer supported.  See here:
https://kubernetes-csi.github.io/docs/external-snapshotter.html#unsupported-versions

PR #535, opened 2021-06-17
Add VS namespace to VSC printed columns

tsmetana:
/kind feature

When using `kubectl get volumesnapshotcontent` the result prints also referenced VolumeSnapshot object. That is not enough to identify the correct one though, since VolumeSnapshot is namespaced. It would be probably good to display also the namespace of the VolumeSnapshot.

Yes.

```release-note
The namespace of the referenced VolumeSnapshot is printed when printing a VolumeSnapshotContent.
```

### DIFF
--- a/client/config/crd/snapshot.storage.k8s.io_volumesnapshotcontents.yaml
+++ b/client/config/crd/snapshot.storage.k8s.io_volumesnapshotcontents.yaml
@@ -42,6 +42,10 @@ spec:
       jsonPath: .spec.volumeSnapshotRef.name
       name: VolumeSnapshot
       type: string
+    - description: Namespace of the VolumeSnapshot object to which this VolumeSnapshotContent object is bound.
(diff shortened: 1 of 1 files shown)


xing-yang:
@tsmetana, can you show the new results of `kubectl get volumesnapshotcontent`?

tsmetana:
The output with the new additional column looks like this:

```
NAME                                               READYTOUSE   RESTORESIZE   DELETIONPOLICY   DRIVER                  VOLUMESNAPSHOTCLASS   VOLUMESNAPSHOT   VOLUMESNAPSHOTNAMESPACE   AGE
snapcontent-7b12a296-c23e-4e8a-88ee-eacd191470ca   true         1073741824    Delete           pd.csi.storage.gke.io   csi-gce-pd-vsc        mysnap1          default                   5m47s
```
It would be definitely prettier if we could just display one column with `default/mysnap1` but I don't if that's even possible.

PR #540, opened 2021-07-04
Add "SnapshotCreated" event for snapshot-controller

### LINKED ISSUE #522: There's no Event for a sucessfully created snapshot
`Snapshot-controller` creates a new normal event with type `Normal` and reason `CreatingSnapshot` after `VolumeSnapshotContent` is created:

https://github.com/kubernetes-csi/external-snapshotter/blob/f255fe7b46bd5410370537319879d06f2bf1a156/pkg/common-controller/snapshot_controller.go#L704-L705

However, when the snapshot is created successfully, the snapshot controller/sidecar only changes `ReadyToUse` field to `true`, but it doesn't create a new event to inform the user the snapshot is created. In other words, it still says the following message when describing the snapshot until the event is deleted:

```
Status:
  Bound Volume Snapshot Content Name:  snapcontent-7e011c44-462e-4fc4-bb2e-47b28e881619
  Creation Time:                       2021-05-23T10:34:08Z
  Ready To Use:                        true
  Restore Size:                        2Gi
Events:
  Type    Reason            Age   From                 Message
  ----    ------            ----  ----                 -------
  Normal  CreatingSnapshot  10m   snapshot-controller  Waiting for a snapshot test-ns/test-volume-1621766047 to be created by the CSI driver.
```

Related PR: https://github.com/kubernetes-csi/external-snapshotter/pull/289

rexagod:
Fixes: #522

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

Adds `SnapshotCreated` and `SnapshotReady` events to snapshot-controller.

Fixes #522 

**Special notes for your reviewer**: This is my first PR here, and I hope this is the expected change, but if not, please let me know and I'll look into it right away. Thanks!

```release-note
Introduced "SnapshotCreated" and "SnapshotReady" events.
```

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -1159,6 +1159,8 @@ func (ctrl *csiSnapshotCommonController) updateSnapshotStatus(snapshot *crdv1.Vo
 		createOperationKey := metrics.NewOperationKey(metrics.CreateSnapshotOperationName, snapshot.UID)
 		if !utils.IsSnapshotCreated(snapshotObj) && utils.IsSnapshotCreated(snapshotClone) {
 			ctrl.metricsManager.RecordMetrics(createOperationKey, metrics.NewSnapshotOperationStatus(metrics.SnapshotStatusTypeSuccess), driverName)
+			msg := fmt.Sprintf("Snapshot %s was successfully created by the CSI driver.", utils.SnapshotKey(snapshot))
+			ctrl.eventRecorder.Event(snapshot, v1.EventTypeNormal, "SnapshotCreated", msg)
 		}
 
 		// Must meet the following criteria to emit a successful CreateSnapshotAndReady status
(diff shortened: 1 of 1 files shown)


pkg/common-controller/snapshot_controller.go
```diff
@@ -715,6 +715,9 @@ func (ctrl *csiSnapshotCommonController) createSnapshotContent(snapshot *crdv1.V
 		klog.Errorf("failed to update content store %v", err)
 	}
 
+	msg = fmt.Sprintf("Snapshot %s was successfully created by the CSI driver.", utils.SnapshotKey(snapshot))
+	ctrl.eventRecorder.Event(snapshot, v1.EventTypeNormal, "SnapshotCreated", msg)
+
```
xing-yang, line 720: At this point, the VolumeSnapshotContent is just created.  The CSI snapshotter sidecar will call the CSI driver to create a snapshot, but we don't know whether it is completed yet.

In updateSnapshotStatus (https://github.com/kubernetes-csi/external-snapshotter/blob/v4.1.1/pkg/common-controller/snapshot_controller.go#L1072), we know whether snapshot is created successfully.

Add a snapshot is created successfully event in this if block:
https://github.com/kubernetes-csi/external-snapshotter/blob/v4.1.1/pkg/common-controller/snapshot_controller.go#L1155

Add a snapshot is ready to use event in this if block:
https://github.com/kubernetes-csi/external-snapshotter/blob/v4.1.1/pkg/common-controller/snapshot_controller.go#L1162

xing-yang:
Can you change this from "/kind bug" to "/kind feature"?

pkg/common-controller/snapshot_controller.go
rexagod: Thank you for the detailed explaination, Xing!

PR #550, opened 2021-07-12
Use distroless/static image for webhook

WanzenBug:
**What type of PR is this?**
/kind cleanup

The webhook image needlessly used the distroless image containing glibc. The webhook server is a static binary, so it can be packaged using the distroless/static image like the other components.

This should make image security scanners happy. As minor benefits, it also slightly decreases image size and makes it so that the same base image is used for all components

```release-note
NONE
```


xinyanw409 (commented):
LGTM

PR #578, opened 2021-08-13
Add CSI secret templates as annotation in volumeSnapshotContent

### LINKED ISSUE #577: Add CSI secret templates as annotation in volumeSnapshotContent
There was a discussion in the meeting on PVC/Snapshot transfer about the model that __operation on volumeSnapshotContent should not rely on information in SnapshotClass after volumeSnapshotContent is created__. The reason for it is that SnapshotClass can be changed/deleted after the volumeSnapshotContent is created, as a result, the controllers might not rely on the current information in SnapshotClass.

In the case of  volumeSnapshotContent transfer, SecretRef in volumeSnapshotContent may be modified based on the templates after transferred. Therefore, these templates should be kept in volumeSnapshotContent as annotations.

Similar issue for PVC has already been opened [here](https://github.com/kubernetes-csi/external-provisioner/issues/654).

mkimuram:
**What type of PR is this?**
/kind feature

Add CSI secret templates as annotation in volumeSnapshotContent

This is required to handle volumeSnapshotContent with secretRef after volumeSnapshotClass is changed/deleted.

Fixes #577

```release-note
If "csi.storage.k8s.io/snapshotter-secret-namespace" and "snapshot.storage.kubernetes.io/deletion-secret-name" are defined as SnapshotClass parameters, they are copied to volumeSnapshotContent as annotations.
```

### DIFF
--- a/pkg/common-controller/snapshot_create_test.go
+++ b/pkg/common-controller/snapshot_create_test.go
@@ -63,6 +63,8 @@ func TestCreateSnapshotSync(t *testing.T) {
 				map[string]string{
 					"snapshot.storage.kubernetes.io/deletion-secret-name":      "secret",
 					"snapshot.storage.kubernetes.io/deletion-secret-namespace": "default",
+					"csi.storage.k8s.io/snapshotter-secret-name":               "secret",
+					"csi.storage.k8s.io/snapshotter-secret-namespace":          "default",
 				}),
 			initialSnapshots:  newSnapshotArray("snap6-2", "snapuid6-2", "claim6-2", "", validSecretClass, "", &False, nil, nil, nil, false, true, nil),
 			expectedSnapshots: newSnapshotArray("snap6-2", "snapuid6-2", "claim6-2", "", validSecretClass, "snapcontent-snapuid6-2", &False, nil, nil, nil, false, true, nil),
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -678,6 +678,17 @@ func (ctrl *csiSnapshotCommonController) createSnapshotContent(snapshot *crdv1.V
 
 		klog.V(5).Infof("createSnapshotContent: set annotation [%s] on content [%s].", utils.AnnDeletionSecretRefNamespace, snapshotContent.Name)
 		metav1.SetMetaDataAnnotation(&snapshotContent.ObjectMeta, utils.AnnDeletionSecretRefNamespace, snapshotterSecretRef.Namespace)
+
+		nameTemplate, namespaceTemplate, err := utils.VerifyAndGetSecretNameAndNamespaceTemplate(utils.SnapshotterSecretParams, class.Parameters)
+		if err != nil {
+			return nil, err
+		}
+
+		klog.V(6).Infof("createSnapshotContent: set annotation [%s] on content [%s].", utils.PrefixedSnapshotterSecretNameKey, nameTemplate)
+		metav1.SetMetaDataAnnotation(&snapshotContent.ObjectMeta, utils.PrefixedSnapshotterSecretNameKey, nameTemplate)
+
+		klog.V(5).Infof("createSnapshotContent: set annotation [%s] on content [%s].", utils.PrefixedSnapshotterSecretNamespaceKey, namespaceTemplate)
+		metav1.SetMetaDataAnnotation(&snapshotContent.ObjectMeta, utils.PrefixedSnapshotterSecretNamespaceKey, namespaceTemplate)
 	}
 
 	var updateContent *crdv1.VolumeSnapshotContent
--- a/pkg/common-controller/snapshot_update_test.go
+++ b/pkg/common-controller/snapshot_update_test.go
@@ -466,7 +466,7 @@ func TestSync(t *testing.T) {
 			// Snapshot status nil, no initial content, new content should be created.
 			name:              "8-1 - Snapshot status nil, no initial snapshot content, new content should be created",
 			initialContents:   nocontents,
-			expectedContents:  withContentAnnotations(newContentArrayNoStatus("snapcontent-snapuid8-1", "snapuid8-1", "snap8-1", "sid8-1", validSecretClass, "", "pv-handle8-1", deletionPolicy, nil, nil, false, false), map[string]string{utils.AnnDeletionSecretRefName: "secret", utils.AnnDeletionSecretRefNamespace: "default"}),
+			expectedContents:  withContentAnnotations(newContentArrayNoStatus("snapcontent-snapuid8-1", "snapuid8-1", "snap8-1", "sid8-1", validSecretClass, "", "pv-handle8-1", deletionPolicy, nil, nil, false, false), map[string]string{utils.AnnDeletionSecretRefName: "secret", utils.AnnDeletionSecretRefNamespace: "default", utils.PrefixedSnapshotterSecretNameKey: "secret", utils.PrefixedSnapshotterSecretNamespaceKey: "default"}),
 			initialSnapshots:  newSnapshotArray("snap8-1", "snapuid8-1", "claim8-1", "", validSecretClass, "", nil, nil, nil, nil, true, false, nil),
 			expectedSnapshots: newSnapshotArray("snap8-1", "snapuid8-1", "claim8-1", "", validSecretClass, "snapcontent-snapuid8-1", &False, nil, nil, nil, false, false, nil),
 			initialClaims:     newClaimArray("claim8-1", "pvc-uid8-1", "1Gi", "volume8-1", v1.ClaimBound, &classEmpty),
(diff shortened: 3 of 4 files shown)


mkimuram:
Tested as below:

- Setup
```
make container-snapshot-controller
make container-csi-snapshotter

cp -r deploy/ /tmp/

grep "image:" /tmp/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml
sed -i 's|k8s.gcr.io/sig-storage/snapshot-controller:v4.1.1|snapshot-controller:latest|' /tmp/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml

grep "image:" /tmp/deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
sed -i 's|k8s.gcr.io/sig-storage/csi-snapshotter:v4.1.1|csi-snapshotter:latest|' /tmp/deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml

kubectl create -f client/config/crd
kubectl create -f /tmp/deploy/kubernetes/snapshot-controller
kubectl create -f /tmp/deploy/kubernetes/csi-snapshotter
```

- Test
```
cp -r examples/ /tmp/
cat << 'EOF' >> /tmp/examples/kubernetes/snapshotclass-v1.yaml
parameters:
  csi.storage.k8s.io/snapshotter-secret-name: ${volumesnapshot.name}-${volumesnapshotcontent.name}
  csi.storage.k8s.io/snapshotter-secret-namespace: ${volumesnapshot.namespace}
EOF
kubectl create -f /tmp/examples/kubernetes/storageclass.yaml
kubectl create -f /tmp/examples/kubernetes/snapshotclass-v1.yaml
kubectl create -f /tmp/examples/kubernetes/pvc.yaml
kubectl create -f /tmp/examples/kubernetes/snapshot-v1.yaml

kubectl get volumesnapshotcontent -o jsonpath='{.items[*].metadata.annotations}'
map[csi.storage.k8s.io/snapshotter-secret-name:${volumesnapshot.name}-${volumesnapshotcontent.name} csi.storage.k8s.io/snapshotter-secret-namespace:${volumesnapshot.namespace} snapshot.storage.kubernetes.io/deletion-secret-name:new-snapshot-demo-v1-snapcontent-41132b6c-1644-490d-8d3a-23ecc070e4ee snapshot.storage.kubernetes.io/deletion-secret-namespace:default]
```

mkimuram:
It may be out of the scope of this PR itself. However, providing a script to add annotations for exisiting VolumeSnapshotContents based on the current VolumeSnapshotClass like below would be helpful (We will need to improve it to exclude certain VolumeSnapshotClasses and so on. And it is just a prototype which needs more tests).

```bash
#!/usr/bin/env bash

function update_secret_annotations() {
  local dry=$1
  
  kubectl get volumesnapshotcontent -o jsonpath='{.items[*].metadata.name}{"\n"}' \
  | while read vsc;do 
    vsc_json=$(kubectl get volumesnapshotcontent ${vsc} -o json)
    ann=$(echo ${vsc_json} | jq '.metadata.annotations')

    del_sec_name=$(echo ${ann} | jq 'to_entries | .[] | select(.key == "snapshot.storage.kubernetes.io/deletion-secret-name") | .value')
    del_sec_ns=$(echo ${ann} | jq 'to_entries | .[] | select(.key == "snapshot.storage.kubernetes.io/deletion-secret-namespace") | .value')
    name_tmpl=$(echo ${ann} | jq 'to_entries | .[] | select(.key == "csi.storage.k8s.io/snapshotter-secret-name") | .value')
    ns_templ=$(echo ${ann} | jq 'to_entries | .[] | select(.key == "csi.storage.k8s.io/snapshotter-secret-namespace") | .value')

    if [[ -z "${del_sec_name}" && -z "${del_sec_ns}" ]];then
      echo "No secret used for volumesnapshotcontent ${vsc}: skipping"
      continue
    fi

    if [[ -n "${name_tmpl}" && -n "${ns_templ}" ]];then
      echo "Secret templates already defined for volumesnapshotcontent ${vsc}: skipping"
      continue
    fi 

    class=$(echo ${vsc_json} | jq -r '.spec.volumeSnapshotClassName')

    if [[ -z "${class}" ]];then
      echo "Error: volumeSnapshotClassName not defined for volumesnapshotcontent ${vsc}: skipping"
      continue
    fi

    class_json=$(kubectl get volumesnapshotclass ${class} -o json)
    params=$(echo ${class_json} | jq '.parameters')
    class_name_tmpl=$(echo ${params} | jq -r 'to_entries | .[] | select(.key == "csi.storage.k8s.io/snapshotter-secret-name") | .value')
    class_ns_tmpl=$(echo ${params} | jq -r 'to_entries | .[] | select(.key == "csi.storage.k8s.io/snapshotter-secret-namespace") | .value')

    if [[ -z "${class_name_tmpl}" || -z "${class_ns_tmpl}" ]];then
      echo "Error: Failed to get templates from VolumeSnapshotClass ${class} for volumesnapshotcontent ${vsc}: skipping"
      continue
    fi

    echo -n "Updating annotation for volumesnapshotcontent ${vsc}: "


    if [ "true" == "$dry" ];then
      opt="--dry-run=client -o jsonpath='{.metadata.annotations}'"
	else
      opt="-o jsonpath='{.metadata.annotations}'"
    fi

    kubectl annotate ${opt} volumesnapshotcontent ${vsc} csi.storage.k8s.io/snapshotter-secret-name=${class_name_tmpl} csi.storage.k8s.io/snapshotter-secret-namespace=${class_ns_tmpl}
    echo

  done
}

[ "-d" == "$1" ] && dry=true

update_secret_annotations ${dry}
```

With `-d` option, it will dry-run and show which vsc will be updated with what annotations like below:
```bash
./vsc_ann_secret.sh -d
Updating annotation for volumesnapshotcontent snapcontent-41132b6c-1644-490d-8d3a-23ecc070e4ee: 'map[csi.storage.k8s.io/snapshotter-secret-name:${volumesnapshot.name}-${volumesnapshotcontent.name} csi.storage.k8s.io/snapshotter-secret-namespace:${volumesnapshot.namespace} snapshot.storage.kubernetes.io/deletion-secret-name:new-snapshot-demo-v1-snapcontent-41132b6c-1644-490d-8d3a-23ecc070e4ee snapshot.storage.kubernetes.io/deletion-secret-namespace:default]'
```

Without option, it actually updates the annotations.

pkg/common-controller/snapshot_create_test.go
```diff
@@ -63,6 +63,8 @@ func TestCreateSnapshotSync(t *testing.T) {
 				map[string]string{
 					"snapshot.storage.kubernetes.io/deletion-secret-name":      "secret",
 					"snapshot.storage.kubernetes.io/deletion-secret-namespace": "default",
+					"csi.storage.k8s.io/snapshotter-secret-name":               "secret",
+					"csi.storage.k8s.io/snapshotter-secret-namespace":          "default",
```
xing-yang, line 67: Why do you need to add this? Deletion secret is already added to content as annotation above.

jsafrane:
/retest
(sorry, mis-using this PR for CI flake tests)

pkg/common-controller/snapshot_create_test.go
mkimuram: Please see my comment [here](https://github.com/kubernetes/enhancements/pull/2849#issuecomment-897759728).
This is to save the secret template itself as annotation without being resolved, not the resolved value, to be latter used to resolve again with different namespace or different volumeSnapshotContent name.
(For above comment case, it is used for just checking if the secret is defined through the template to see if transferring the snapshot could cause issues by secret).

mkimuram: Also, check the discussion from 35:00. It was talking about PV, but the same thing is true to volumeSnapshotContent.
 https://youtu.be/hQZDcJmAB80?t=2105

xing-yang: I'm not sure if it is safe to adjust the secret based on template like that.  I'd prefer a more explicit way of transferring secret. Here's an example: https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/develop/docs/secret-export.md

mkimuram: > Here's an example: https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/develop/docs/secret-export.md

I agree with you that it would be good to provide a feature to move/copy the secret itself across the namespace by such a way (and it will be one of the another use cases for `StorageTransferRequest` and `StorageTransferApproval`).

However, what I'm discussing here is different from the use case. It is to update the reference to the secret that is defined in the VolumeSnapshotContent. After transfer, VolumeSnapshotContent may still referencing to the secret that may be only allowed to be accessed from the original namespace, as I explained [here]( https://github.com/kubernetes/enhancements/pull/2849#discussion_r692459041).
Therefore, the reference may need to be updated in some cases. And to do so, this feature will be of help (I'm not saying this is the only way and we can discuss if there will be a better way).

PR #596, opened 2021-10-04
Fix VolumeSnapshotContentStatus field type

### LINKED ISSUE #595: `VolumeSnapshotContentStatus` field type inconsistency
When get a ready `VolumeSnapshotContentStatus`: `kubectl get volumesnapshotcontent snapcontent-xxx -oyaml`
got follows output:
```
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshotContent
metadata:
  ...
spec:
  ...
status:
  creationTime: 1633315996000000000
  readyToUse: true
  restoreSize: 536870912000
  snapshotHandle: snap-xxx
```

To be consistent with `VolumeSnapshotStatus`:
the `creationTime: 1633315996000000000` should be like `creationTime: "2021-10-04T02:53:16Z"`
the `restoreSize: 536870912000` should be like `restoreSize: 500Gi`

astraw99:
**What type of PR is this?**
/kind bug

When get a ready `VolumeSnapshotContentStatus`: `kubectl get volumesnapshotcontent snapcontent-xxx -oyaml`
got follows output:
```
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshotContent
metadata:
  ...
spec:
  ...
status:
  creationTime: 1633315996000000000
  readyToUse: true
  restoreSize: 536870912000
  snapshotHandle: snap-xxx
```

To be consistent with `VolumeSnapshotStatus`:
the `creationTime: 1633315996000000000` should be like `creationTime: "2021-10-04T02:53:16Z"`
the `restoreSize: 536870912000` should be like `restoreSize: 500Gi`

Fixes #595

```release-note
NONE
```

### DIFF
--- a/client/apis/volumesnapshot/v1/types.go
+++ b/client/apis/volumesnapshot/v1/types.go
@@ -379,7 +379,7 @@ type VolumeSnapshotContentStatus struct {
 	// On Unix, the command `date +%s%N` returns the current time in nanoseconds
 	// since 1970-01-01 00:00:00 UTC.
 	// +optional
-	CreationTime *int64 `json:"creationTime,omitempty" protobuf:"varint,2,opt,name=creationTime"`
+	CreationTime *metav1.Time `json:"creationTime,omitempty" protobuf:"varint,2,opt,name=creationTime"`
 
 	// restoreSize represents the complete size of the snapshot in bytes.
 	// In dynamic snapshot creation case, this field will be filled in by the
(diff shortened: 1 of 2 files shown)


client/apis/volumesnapshot/v1/types.go
```diff
@@ -379,7 +379,7 @@ type VolumeSnapshotContentStatus struct {
 	// On Unix, the command `date +%s%N` returns the current time in nanoseconds
 	// since 1970-01-01 00:00:00 UTC.
 	// +optional
-	CreationTime *int64 `json:"creationTime,omitempty" protobuf:"varint,2,opt,name=creationTime"`
+	CreationTime *metav1.Time `json:"creationTime,omitempty" protobuf:"varint,2,opt,name=creationTime"`
```
RaunakShah, line 382: It seems from the comment that this is meant to return the time in nanoseconds, rather than in the time format. I'm also not sure whether you can change the type of a released API...

xing-yang:
This works as designed. The CreationTime in VolumeSnapshotContent is designed to be the same as whatever is returned by the CSI Driver without conversion. VolumeSnapshotContent is meant for admin so this is fine. For users, the CreationTime in VolumeSnapshot shows readable Data Time format.

xing-yang:
>> This works as designed. The CreationTime in VolumeSnapshotContent is designed to be the same as whatever is returned by the CSI Driver without conversion. VolumeSnapshotContent is meant for admin so this is fine. For users, the CreationTime in VolumeSnapshot shows readable Data Time format.

> Got it, but for admin if we can make the VolumeSnapshotContent more readable is preferred?

We've already discussed about this during the design and made the decision, so it will stay as is.

PR #606, opened 2021-11-08
Add kustomization manifests to CRDs and controller components

### LINKED ISSUE #605: Kustomize manifests for CRDs and snapshot controller?
This repository is not very GitOps-friendly. While building and maintaining a Helm chart as suggested in #551 might be something that's not very high on the priority list, adding `kustomization.yaml` files to a few directories is neither a huge task nor does it require maintenance, and it would (at least for me and likely other users of automation frameworks like FluxCD) pretty much achieve the same thing - being able to install CRDs and controller from a versioned git repository.

I'll open a PR implementing this shortly.

itspngu:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind design

The changes introduced with this PR allow for better integration of the controller components and, in particular, the CRDs (which other projects rely on) into a GitOps workflow by providing a means to reference versioned manifests via `kustomize`.

Fixes #605 

none

```release-note
Added kustomization manifests to CRDs and controller components
```

### DIFF
--- a/client/config/crd/kustomization.yaml
+++ b/client/config/crd/kustomization.yaml
@@ -1,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - snapshot.storage.k8s.io_volumesnapshotclasses.yaml
+  - snapshot.storage.k8s.io_volumesnapshotcontents.yaml
+  - snapshot.storage.k8s.io_volumesnapshots.yaml
--- a/deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
+++ b/deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
@@ -118,4 +118,4 @@ spec:
               mountPath: /csi
       volumes:
         - name: socket-dir
-          emptyDir:
+          emptyDir: {}
--- a/README.md
+++ b/README.md
@@ -62,20 +62,19 @@ If your Kubernetes distribution does not bundle the snapshot controller, you may
 There is a new validating webhook server which provides tightened validation on snapshot objects. The cluster admin or Kubernetes distribution admin should install the webhook alongside the snapshot controllers and CRDs. More details [below](#validating-webhook).
 
 Install Snapshot CRDs:
-* kubectl create -f client/config/crd
+* kubectl kustomize client/config/crd | kubectl create -f -
 * https://github.com/kubernetes-csi/external-snapshotter/tree/master/client/config/crd
 * Do this once per cluster
 
 Install Common Snapshot Controller:
 * Update the namespace to an appropriate value for your environment (e.g. kube-system)
-* kubectl create -f deploy/kubernetes/snapshot-controller
-* https://github.com/kubernetes-csi/external-snapshotter/tree/master/deploy/kubernetes/snapshot-controller
+* kubectl -n kube-system kustomize deploy/kubernetes/snapshot-controller | kubectl create -f -
 * Do this once per cluster
 
 Install CSI Driver:
 * Follow instructions provided by your CSI Driver vendor.
 * Here is an example to install the sample hostpath CSI driver
-  * kubectl create -f deploy/kubernetes/csi-snapshotter
+  * kubectl kustomize deploy/kubernetes/csi-snapshotter | kubectl create -f -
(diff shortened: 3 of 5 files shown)


client/config/crd/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - snapshot.storage.k8s.io_volumesnapshotclasses.yaml
+  - snapshot.storage.k8s.io_volumesnapshotcontents.yaml
+  - snapshot.storage.k8s.io_volumesnapshots.yaml
```
xing-yang, line 6: @pohly Do you see any issue with this change?  We install snapshot CRDs from external-snapshotter/pkg/client/config/crd folder:
https://github.com/kubernetes-csi/csi-release-tools/blob/master/prow.sh#L742

deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
```diff
@@ -118,4 +118,4 @@ spec:
               mountPath: /csi
       volumes:
         - name: socket-dir
-          emptyDir:
+          emptyDir: {}
```
pohly, line 121: Why is this needed?

client/config/crd/kustomization.yaml
pohly: Adding the kustomization.yaml makes sense to me, I can imagine that some users may want to use the directory as their base.

But what does that have to do with prow.sh? Our usage of the files will simply ignore the additional kustomization.yaml.

deploy/kubernetes/csi-snapshotter/setup-csi-snapshotter.yaml
itspngu: VSCode was complaining about it (`Incorrect type. Expected "object".`) when I went through the files and I must have impulsively... "fixed" it. 

To be on the safe side, I just ran kubeval on it. It doesn't mind either version (with or without empty object). I can remove that bit if you'd prefer to keep it as is to avoid complications in automation, sorry for not noticing myself when opening the PR.

pohly: Let's keep it. I was just wondering because it seemed unrelated to the stated purpose of the PR.

client/config/crd/kustomization.yaml
xing-yang: Just want to double check.  It should be fine.

xing-yang: @itspngu Can you update the "Install Snapshot CRDs:" section from "kubectl create -f client/config/crd" to install individual CRDs?  I want to avoid confusion for folks who are new to this.
https://github.com/kubernetes-csi/external-snapshotter/blob/master/README.md#usage

itspngu: > install individual CRDs

Good catch! Are you sure you want all the individual files though? That'd be rather verbose. Part of the beauty of kustomize is that you can easily do something like this:

```
$ kubectl kustomize client/config/crd | kubectl create -f - --dry-run=server
Error from server (AlreadyExists): error when creating "STDIN": customresourcedefinitions.apiextensions.k8s.io "volumesnapshotclasses.snapshot.storage.k8s.io" already exists
Error from server (AlreadyExists): error when creating "STDIN": customresourcedefinitions.apiextensions.k8s.io "volumesnapshotcontents.snapshot.storage.k8s.io" already exists
Error from server (AlreadyExists): error when creating "STDIN": customresourcedefinitions.apiextensions.k8s.io "volumesnapshots.snapshot.storage.k8s.io" already exists
```

As such, I'd update the readme with something similar to the above where needed, unless you prefer the explicit way using individual files.

xing-yang: That should be fine.  Thanks

itspngu: I pushed an update to the readme. Let me know if there's anything else to change.

xing-yang:
@itspngu Can you backport this to release-5.0 branch?

itspngu:
> @itspngu Can you backport this to release-5.0 branch?

Okay, will do so later today in a separate PR.

PR #621, opened 2021-11-29
Make the QPS and Burst of kube client config to be configurable

lintongj:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

>/kind flake

**What this PR does / why we need it**:

This change adds two new flags "kube-api-qps" and "kube-api-burst" to make the QPS and Burst of clients to the API server configurable. Their default values stay the same as defaults of kube client.

**Experiment**

To justify the necessity to make this change, we run experiments on external-snapshotters between defaullt qps and burst and a pair of hardcoded ones. Following are the results for 32 threads runs.

1) Default QPS and Burst: QPS=5.0 and Burst=10

| tput.ops.per.sec | littles.law |
| ---------------- | ----------- |
|            0.497 |      32.367 |
|            1.266 |      31.614 |

2) QPS and Burst set to 100

| tput.ops.per.sec | littles.law |
| ---------------- | ----------- |
|            9.760 |      32.050 |
|           11.361 |      32.097 |

We built the snapshot-controller and csi-snapshotter with the hardcoded value of 100 for both QPS and Burst and observed 9x improvement for CreateSnapshot and 11x improvement for DeleteSnapshot.

** Testing **

Run the following unit tests to ensure there is no regression.

```
$pwd
/Users/lintongj/go/src/github.com/kubernetes-csi/external-snapshotter
$go test -timeout 30s  ./pkg/... --count=1
ok  	github.com/kubernetes-csi/external-snapshotter/v4/pkg/common-controller	0.495s
ok  	github.com/kubernetes-csi/external-snapshotter/v4/pkg/metrics	4.454s
ok  	github.com/kubernetes-csi/external-snapshotter/v4/pkg/sidecar-controller	0.337s
ok  	github.com/kubernetes-csi/external-snapshotter/v4/pkg/snapshotter	0.245s
ok  	github.com/kubernetes-csi/external-snapshotter/v4/pkg/utils	0.319s
ok  	github.com/kubernetes-csi/external-snapshotter/v4/pkg/validation-webhook	9.735s
```

Also run through `make test` locally. And get the following message.

```
$make test
...
Congratulations! All shell files are passing lint.
```

**Which issue(s) this PR fixes**:

Fixes #

There is no issue opened for this issue. I can open one if required.

[...]

PR #626, opened 2021-12-10
Fixed the bug introduced by PR 621 and exposed the kube-api-qps cmd option properly

lintongj:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

There is an issue introduced by #621 that the pflag were not added to the go flag set, so that the pflag cmd option, `kube-api-qps`, was not exposed. This PR fixes this bug by getting rid of the dependency on pflag package.

Fixes #
N/A

**Testing**

```
$./bin/csi-snapshotter -h
...
      --kube-api-burst int                        Burst to use while communicating with the kubernetes apiserver. Defaults to 10. (default 10)
      --kube-api-qps float32                      QPS to use while communicating with the kubernetes apiserver. Defaults to 5.0. (default 5)
...
```

```
$./bin/snapshot-controller -h
...
      --kube-api-burst int                        Burst to use while communicating with the kubernetes apiserver. Defaults to 10. (default 10)
      --kube-api-qps float32                      QPS to use while communicating with the kubernetes apiserver. Defaults to 5.0. (default 5)
...
```

```
make test
...
Congratulations! All shell files are passing lint.
```

N/A

```release-note
Fixed a bug introduced by PR 621 and exposed the kube-api-qps cmd option properly
```

### DIFF
--- a/cmd/csi-snapshotter/main.go
+++ b/cmd/csi-snapshotter/main.go
@@ -28,7 +28,6 @@ import (
 
 	"google.golang.org/grpc"
 
-	"github.com/spf13/pflag"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
@@ -73,7 +72,7 @@ var (
 	leaderElectionRenewDeadline = flag.Duration("leader-election-renew-deadline", 10*time.Second, "Duration, in seconds, that the acting leader will retry refreshing leadership before giving up. Defaults to 10 seconds.")
 	leaderElectionRetryPeriod   = flag.Duration("leader-election-retry-period", 5*time.Second, "Duration, in seconds, the LeaderElector clients should wait between tries of actions. Defaults to 5 seconds.")
 
-	kubeAPIQPS   = pflag.Float32("kube-api-qps", 5, "QPS to use while communicating with the kubernetes apiserver. Defaults to 5.0.")
+	kubeAPIQPS   = flag.Float64("kube-api-qps", 5, "QPS to use while communicating with the kubernetes apiserver. Defaults to 5.0.")
 	kubeAPIBurst = flag.Int("kube-api-burst", 10, "Burst to use while communicating with the kubernetes apiserver. Defaults to 10.")
 
 	metricsAddress     = flag.String("metrics-address", "", "(deprecated) The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled. Only one of `--metrics-address` and `--http-endpoint` can be set.")
@@ -106,7 +105,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	config.QPS = *kubeAPIQPS
+	config.QPS = (float32)(*kubeAPIQPS)
 	config.Burst = *kubeAPIBurst
 
 	kubeClient, err := kubernetes.NewForConfig(config)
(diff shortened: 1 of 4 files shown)


xing-yang:
Why are kubeAPIQPS and kubeAPIBurst not using the same flag package?

	kubeAPIQPS   = pflag.Float32("kube-api-qps", 5, "QPS to use while communicating with the kubernetes apiserver. Defaults to 5.0.")
	kubeAPIBurst = flag.Int("kube-api-burst", 10, "Burst to use while communicating with the kubernetes apiserver. Defaults to 10.")

xing-yang:
We should do what is done in csi-provisioner and make all parameters use pflag:
https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go

lintongj:
> Why are kubeAPIQPS and kubeAPIBurst not using the same flag package?
> 
> ```
> kubeAPIQPS   = pflag.Float32("kube-api-qps", 5, "QPS to use while communicating with the kubernetes apiserver. Defaults to 5.0.")
> kubeAPIBurst = flag.Int("kube-api-burst", 10, "Burst to use while communicating with the kubernetes apiserver. Defaults to 10.")
> ```

What external-snapshotter use is the golang built-in flag package, which does not support Float32 type. However, k8s config.QPS requires Float32 type. That's why I use pflag.Float32 here.


> We should do what is done in csi-provisioner and make all parameters use pflag: https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go

Actually, csi-provisioner also use pflag. But, there is some little trick. In https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go, pflag was renamed to flag. 
```
import (
   ...
   goflag "flag"
   ...
   flag "github.com/spf13/pflag"
   ...
)
```
We can do the same thing in external-snapshotter too, but it doesn't make much difference functionally.

pohly:
Beware that switching from flag to pflag is an API break in a binary because single-dash parameters (= the current API) are no longer accepted when parsing with pflag.

cmd/csi-snapshotter/main.go
```diff
@@ -90,8 +90,9 @@ var (
 
 func main() {
 	klog.InitFlags(nil)
-	flag.Set("logtostderr", "true")
-	flag.Parse()
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
```
pohly, line 93: Unless you want to cause a major version bump, the fix has to be the other way around: copy the parameter defined with pflag into the flag.CommandLine.

xing-yang:
Actually it is not necessary to use `pflag` here. `pflag` was added to external-provisioner because it needs a feature gate.  Here's an example in external-attacher:
https://github.com/kubernetes-csi/external-attacher/pull/286

lintongj:
Thanks @pohly @xing-yang for the comments. I updated code and got rid of the dependency on pflag.

xing-yang:
Can you squash the commits?

xing-yang:
You can remove "Fixes #621" from the PR description.  That should be an issue number, not a PR number.  The bot will try to automatically close the issue mentioned after "Fixes" when this PR is merged.

xing-yang:
Release note needs an update.

xing-yang:
@lintongj Please cherry-pick this PR to release-5.0 branch.

PR #651, opened 2022-01-24
external-snapshotter constantly retrying CreateSnapshot calls on erro…

### LINKED ISSUE #533: external-snapshotter constantly retrying CreateSnapshot calls on error w/o backoff
Creating a snapshot from a non-existing volume will repro this. @jsafrane discovered this, opening this issue for tracking.

zhucan:
…r w/o backoff

Signed-off-by: zhucan <zhucan.k8s@gmail.com>

**What type of PR is this?**
/kind bug

external-snapshotter constantly retrying CreateSnapshot calls on error w/o backoff

Fixes #533 

@xing-yang 

```release-note
Fix a problem in CSI snapshotter sidecar that constantly retries CreateSnapshot call on error without exponential backoff.
```

### DIFF
--- a/pkg/sidecar-controller/snapshot_controller_base.go
+++ b/pkg/sidecar-controller/snapshot_controller_base.go
@@ -25,6 +25,7 @@ import (
 	storageinformers "github.com/kubernetes-csi/external-snapshotter/client/v4/informers/externalversions/volumesnapshot/v1"
 	storagelisters "github.com/kubernetes-csi/external-snapshotter/client/v4/listers/volumesnapshot/v1"
 	"github.com/kubernetes-csi/external-snapshotter/v4/pkg/snapshotter"
+	"github.com/kubernetes-csi/external-snapshotter/v4/pkg/utils"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -94,8 +95,22 @@ func NewCSISnapshotSideCarController(
 
 	volumeSnapshotContentInformer.Informer().AddEventHandlerWithResyncPeriod(
 		cache.ResourceEventHandlerFuncs{
-			AddFunc:    func(obj interface{}) { ctrl.enqueueContentWork(obj) },
-			UpdateFunc: func(oldObj, newObj interface{}) { ctrl.enqueueContentWork(newObj) },
+			AddFunc: func(obj interface{}) { ctrl.enqueueContentWork(obj) },
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				// If the CSI driver fails to create a snapshot and returns a failure, the CSI Snapshotter sidecar
+				// will remove the "AnnVolumeSnapshotBeingCreated" annotation from the VolumeSnapshotContent.
+				// This will trigger a VolumeSnapshotContent update and it will cause the obj to be re-queued immediately
+				// and CSI CreateSnapshot will be called again without exponential backoff.
+				// So we are skipping the re-queue here to avoid CreateSnapshot being called without exponential backoff.
+				newSnapContent := newObj.(*crdv1.VolumeSnapshotContent)
+				oldSnapContent := oldObj.(*crdv1.VolumeSnapshotContent)
+				_, newExists := newSnapContent.ObjectMeta.Annotations[utils.AnnVolumeSnapshotBeingCreated]
+				_, oldExists := oldSnapContent.ObjectMeta.Annotations[utils.AnnVolumeSnapshotBeingCreated]
+				if !newExists && oldExists {
+					return
+				}
+				ctrl.enqueueContentWork(newObj)
+			},
 			DeleteFunc: func(obj interface{}) { ctrl.enqueueContentWork(obj) },
 		},
 		ctrl.resyncPeriod,


pohly:
/skip

I don't know this code.

pkg/sidecar-controller/snapshot_controller_base.go
```diff
@@ -94,8 +95,15 @@ func NewCSISnapshotSideCarController(
 
 	volumeSnapshotContentInformer.Informer().AddEventHandlerWithResyncPeriod(
 		cache.ResourceEventHandlerFuncs{
-			AddFunc:    func(obj interface{}) { ctrl.enqueueContentWork(obj) },
-			UpdateFunc: func(oldObj, newObj interface{}) { ctrl.enqueueContentWork(newObj) },
+			AddFunc: func(obj interface{}) { ctrl.enqueueContentWork(obj) },
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				newSnapContent := newObj.(*crdv1.VolumeSnapshotContent)
+				oldSnapContent := oldObj.(*crdv1.VolumeSnapshotContent)
+				if newSnapContent.ObjectMeta.Annotations[utils.AnnVolumeSnapshotBeingCreated] != oldSnapContent.ObjectMeta.Annotations[utils.AnnVolumeSnapshotBeingCreated] {
```
xing-yang, line 102: We also need to make sure this is the only difference.

zhucan: will update different fields at the same time? I means update diferent fileds only do once update or patch.  Beause of the client-go is the fifo queue, I think we no need to keep other are same.

zhucan:
image
This is the test results. WIth the defailt retry interval times, The retry interval times will be 1s , 2s, 4s, 8s, 16s, 32s, 64s, 128s, 256s, 5m.

pkg/sidecar-controller/snapshot_controller_base.go
xing-yang: Can you add comments to the code to explain why this change is needed?

zhucan: yes

pkg/sidecar-controller/snapshot_controller_base.go
```diff
@@ -94,8 +95,18 @@ func NewCSISnapshotSideCarController(
 
 	volumeSnapshotContentInformer.Informer().AddEventHandlerWithResyncPeriod(
 		cache.ResourceEventHandlerFuncs{
-			AddFunc:    func(obj interface{}) { ctrl.enqueueContentWork(obj) },
-			UpdateFunc: func(oldObj, newObj interface{}) { ctrl.enqueueContentWork(newObj) },
+			AddFunc: func(obj interface{}) { ctrl.enqueueContentWork(obj) },
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				/* Update the snapshotcontent (remove the 'AnnVolumeSnapshotBeingCreated' from annotations) after the snapshotcontent was created failed,
+				it will cause the obj be re-queued immediately and retrying CreateSnapshot calls on error w/o backoff.
+				*/
```
xing-yang, line 102: Can you change to the following?

// If the CSI driver fails to create a snapshot and returns a failure, the CSI Snapshotter sidecar
// will remove the "AnnVolumeSnapshotBeingCreated" annotation from the VolumeSnapshotContent.
// This will trigger a VolumeSnapshotContent update and it will cause the obj to be re-queued immediately
// and CSI CreateSnapshot will be called again without exponential backoff.
// So we are skipping the re-queue here to avoid CreateSnapshot being called without exponential backoff.

pkg/sidecar-controller/snapshot_controller_base.go
xing-yang: We should check the annotation is on the newSnapContent, but not on the oldSnapContent.

pkg/sidecar-controller/snapshot_controller_base.go
zhucan: ok

pkg/sidecar-controller/snapshot_controller_base.go
zhucan: Only check the "Annotations[utils.AnnVolumeSnapshotBeingCreated]" whether it is empty, no need to compare?

zhucan: But I think whatever remove or add the "Annotations[utils.AnnVolumeSnapshotBeingCreated]"  to the annotations, no need to re-queue.

anshulahuja98:
@xing-yang  can we take a release of the external snapshotter with this fix included?

pwschuurman:
@xing-yang For PD CSI driver, volume snapshot typically takes <1 minute for a snapshot to be ready. However with this change, a call to syncContent won't happen until the next controller resync interval, which is 15 minutes by default.

I chatted with @mattcary, and we think a short term fix would be to reintroduce the call to `enqueueContentWork` in the success case, but keep the logic as is for the error case (by checking the status on the new `VolumeSnapshotContent`).

xing-yang:
@pwschuurman That makes sense.  Do you want to submit a fix?

pwschuurman:
@xing-yang Yes, I can send out a PR to fix.

PR #666, opened 2022-03-02
Allow resync on VolumeSnapshotContent when there is no error from CreateSnapshot

### LINKED ISSUE #533: external-snapshotter constantly retrying CreateSnapshot calls on error w/o backoff

pwschuurman:
**What type of PR is this?**
/kind bug

After creating a snapshot, external-snapshotter needs to wait for a VolumeSnapshotContent to be ready. This re-adds logic (removed in #651) to enqueue a VolumeSnapshotContent after it has been updated, as long as there wasn't an error encountered in creating (or updating) the snapshot. Effectively this allows the controller to continuously poll the VolumeSnapshotContent until it is ready.
Fixes #533 

```release-note
Ensure that the CSI snapshotter sidecar allows for CreateSnapshot retries while waiting for a snapshot to be ready.
```

### DIFF
--- a/pkg/sidecar-controller/snapshot_controller_base.go
+++ b/pkg/sidecar-controller/snapshot_controller_base.go
@@ -97,17 +97,20 @@ func NewCSISnapshotSideCarController(
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: func(obj interface{}) { ctrl.enqueueContentWork(obj) },
 			UpdateFunc: func(oldObj, newObj interface{}) {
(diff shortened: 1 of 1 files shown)


xing-yang:
@pwschuurman Can you add a release note?  Have you tested this fix with those PD CSI driver tests that failed previously?

pwschuurman:
Updated the release note. I have tested this fix with the PD CSI driver tests, and this update fixes them.

PR #674, opened 2022-03-25
Add default validation for VolumeSnapshotClass

### LINKED ISSUE #667: Webhook to ensure single default VolumeSnapshotClass

When users create two or more default VolumeSnapshotClasses, any subsequent VolumeSnapshot they create is then pending with `Failed to set default snapshot class with error 2 default snapshot classes were found` event.

It would be cool to have an admission webhook that would prevent creation / update of the second default VolumeSnapshotClass.

A simple webhook that makes sure that there is only one default VolumeSnapshotClass for a given CSI driver, returning a nice error to users when they try to make two default snapshot classes.

**Additional context**

Suggested workflow for changing the default VolumeSnapshotClass for a CSI driver would be:

1. Delete the defaulting annotation from the "old" default VolumeSnapshotClass.
2. Add the annotation to the "new" default VolumeSnapshotClass.

As you can see, there is a (brief?) moment when the snapshot-controller will not see any default class and it won't be able to process VolumeSnapshots that don't have any class set (yet). I *assume* the controller will re-sync these snapshots after step 2. is completed and file the new default VolumeSnapshotClass in them. Please test this scenario carefully.

shawn-hurley:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Adding validation so that a user can not create or update to a situation where they have more than default VolumeSnapshotClass

Fixes #667 

```release-note
Adding validation for VolumeSnapshotClass to only have a single default for a particular driver. 
```


shawn-hurley:
Still working on e2e tests but @jsafrane when you get a chance can you validate that this approach works?

pkg/validation-webhook/snapshot.go
```diff
@@ -36,10 +39,14 @@ var (
 	SnapshotContentV1Beta1GVR = metav1.GroupVersionResource{Group: volumesnapshotv1beta1.GroupName, Version: "v1beta1", Resource: "volumesnapshotcontents"}
 	// SnapshotContentV1GVR is GroupVersionResource for v1 VolumeSnapshotContents
 	SnapshotContentV1GVR = metav1.GroupVersionResource{Group: volumesnapshotv1.GroupName, Version: "v1", Resource: "volumesnapshotcontents"}
+	// SnapshotContentV1Beta1GVR is GroupVersionResource for v1beta1 VolumeSnapshotContents
+	SnapshotClassV1Beta1GVR = metav1.GroupVersionResource{Group: volumesnapshotv1beta1.GroupName, Version: "v1beta1", Resource: "volumesnapshotclasses"}
```
xing-yang, line 43: Beta APIs will be removed before the next major release.

shawn-hurley: Thanks will make the changes!

pkg/validation-webhook/snapshot.go
```diff
@@ -223,6 +242,51 @@ func decideSnapshotContentV1(snapcontent, oldSnapcontent *volumesnapshotv1.Volum
...
+			// but not let the problem get worse. having a single one will
+			// suffice.
+			klog.V(2).Infof("snapshot driver: %v has multiple default classes", snapshotClass.Driver)
+		}
+		driverToSnapshotDefautlMap[snapshotClass.Driver] = snapshotClass
+	}
+
+	if _, ok := driverToSnapshotDefautlMap[snapClass.Driver]; ok {
+		reviewResponse.Allowed = false
+		reviewResponse.Result.Message = fmt.Sprintf("default snapshot class already exits for driver: %v", snapClass.Driver)
```
jsafrane, line 284: Please add the existing VolumeSnapshotClass name to the message, so people don't need to list all storage classes to find the conflicting one.

pkg/validation-webhook/webhook.go
```diff
@@ -215,7 +230,30 @@ func main(cmd *cobra.Command, args []string) {
...
+		os.Exit(1)
+	}
+	snapClient, err := clientset.NewForConfig(config)
+	if err != nil {
+		klog.Errorf("Error building snapshot clientset: %s", err.Error())
+		os.Exit(1)
+	}
+
+	factory := informers.NewSharedInformerFactory(snapClient, 0)
+	lister := factory.Snapshot().V1().VolumeSnapshotClasses().Lister()
```
jsafrane, line 247: Did you test it in a real cluster? I *think* it was necessary to run the informer (`go factory.Run()`) and also to wait for the informer to get populated by initial storage class list (`cache.WaitForCacheSync()`) and only then start the server.

jsafrane:
Please update the unit tests with the new message
```
--- FAIL: TestAdmitVolumeSnapshotClassV1 (0.00s)
    --- FAIL: TestAdmitVolumeSnapshotClassV1/new_default_for_class_with_existing_default_class_same_driver (0.00s)
        snapshot_test.go:945: expected "default snapshot class:  already exits for driver: test.csi.io" to equal "default snapshot class: driver-a already exits for driver: test.csi.io"
    --- FAIL: TestAdmitVolumeSnapshotClassV1/new_snapshot_for_class_with_existing_default_classes (0.00s)
        snapshot_test.go:945: expected "default snapshot class:  already exits for driver: test.csi.io" to equal "default snapshot class already exits for driver: test.csi.io"
```

pkg/validation-webhook/webhook.go
shawn-hurley: Why it is still a WIP :) I am working on this part now, but thanks for the pointer!

shawn-hurley:
I also test the scenario where a default for a driver was removed and a new default added but in between a volume snapshot was created. it does get the default eventually just like you said once the default is created.

jsafrane:
I played with this in a real cluster and it works in the simple cases (create a new default VolumeSnapshotClass (VSC), edit existing VCS and make it default). I was able to fool it like this:

1. Create default VSC 1 for `driver: ebs.csi.aws.com`.
2. Create default VSC 2 for `driver: pd.csi.storage.gke.io`. It's a different driver, therefore it's OK.
3. Edit VSC 2 and change `driver: pd.csi.storage.gke.io` to `driver: ebs.csi.aws.com` and keep it default. From some reason, the webhook allows this and I end up with two default VSCs for `driver: ebs.csi.aws.com` :smiling_imp:

pkg/validation-webhook/snapshot.go
```diff
@@ -223,6 +254,43 @@ func decideSnapshotContentV1(snapcontent, oldSnapcontent *volumesnapshotv1.Volum
...
+
+	// Only Validate when a new snapClass is being set as a default.
+	if snapClass.Annotations[utils.IsDefaultSnapshotClassAnnotation] != "true" {
+		return reviewResponse
+	}
+
+	// If Old snapshot class has this, then we can assume that it was validated
+	if oldSnapClass.Annotations[utils.IsDefaultSnapshotClassAnnotation] == "true" {
+		return reviewResponse
+	}
```
jsafrane, line 271: This is tricky. As I pointed out above, if `driver:` is changed in the new VSC, it may be the second default VSC for the new driver and some validation is needed.

IMO it would make sense not to allow change of `driver:` in any VSC, we don't allow it in StorageClasses too. Then the `if` above will be correct.

Or re-check classes of the new driver if driver name is changed.

shawn-hurley: I will add this as a test case and fix it. I should have thought of this sorry.

shawn-hurley: I think we probably don't want to change existing behavior right?

jsafrane: > I think we probably don't want to change existing behavior right?

It's tempting, but no, let's allow changing VSC.driver.

jsafrane:
Did anyone had a chance to test it in a real cluster? The code looks good to me.

jsafrane:
/lgtm
/approve

I tested it shortly in a real cluster and I did not find any obvious bug. Please update README and deploy/kubernetes/webhook-example to validate also volumesnapshotclasses

xing-yang:
@shawn-hurley Can you backport this to release-6.0?

xing-yang:
@shawn-hurley Looks like RBAC rules are missing for listing VolumeSnapshotClass in the validation webhook.  Can you please take a look?

PR #707, opened 2022-05-16
adjust the RBAC of csi snapshotter to only have `PATCH` verb

humblec:
Considering Update() are replaced by Patch() call, we no longer
need the UPDATE RBAC for the sidecar.

> /kind cleanup

```release-note
NONE
```


xing-yang:
I think "Update" is still used in a few places.

cc @ggriffiths

ggriffiths:
Update is still used in certain cases. We handled and replaced the major cases where Update was causing API update conflicts.

humblec:
> Update is still used in certain cases. We handled and replaced the major cases where Update was causing API update conflicts.

Yep, noticed, we are half way through here.. I am closing this PR for now and can reopen once we are completely away from Update. thanks ..

PR #719, opened 2022-05-25
Backport 718: Add back v1beta1 manifests

xing-yang:
(cherry picked from commit 5c47a8ceb4d950c42729eee43e4f74a5b3ab8e6a)

**What type of PR is this?**
/kind api-change

Cherry-pick #718 

Fixes #

```release-note
Cherry-pick #718: Add VolumeSnapshot v1beta1 manifests back. VolumeSnapshot v1beta1 APIs are no longer served.
Action Item: Please update to VolumeSnapshot v1 APIs as soon as possible.
```

### DIFF
--- a/client/hack/README.md
+++ b/client/hack/README.md
@@ -54,6 +54,8 @@ Once you run the script, you will get an output as follows:
 
 ## update-crd.sh
 
+NOTE: We need to keep both v1beta1 and v1 snapshot APIs but set served and storage version of v1beta1 to false. Please copy back the v1beta1 manifest back to the files as this script will remove it.
+
 This is the script to update CRD yaml files under /client/config/crd/ based on types.go file.
 
 Make sure to run this script after making changes to /client/apis/volumesnapshot/v1/types.go.
--- a/client/config/crd/snapshot.storage.k8s.io_volumesnapshotclasses.yaml
+++ b/client/config/crd/snapshot.storage.k8s.io_volumesnapshotclasses.yaml
@@ -78,6 +78,54 @@ spec:
     served: true
     storage: true
     subresources: {}
+  - additionalPrinterColumns:
+    - jsonPath: .driver
+      name: Driver
+      type: string
+    - description: Determines whether a VolumeSnapshotContent created through the VolumeSnapshotClass should be deleted when its bound VolumeSnapshot is deleted.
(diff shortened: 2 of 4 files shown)


pohly:
Can someone explain why it gets done this way? I don't see that, neither here nor in the original PR, and it isn't obvious.

client/hack/README.md
```diff
@@ -54,6 +54,8 @@ Once you run the script, you will get an output as follows:
 
 ## update-crd.sh
 
+NOTE: We need to keep both v1beta1 and v1 snapshot APIs but set served and storage version of v1beta1 to false. Please copy back the v1beta1 manifest back to the files as this script will remove it.
+
```
pohly, line 58: Shouldn't that be fixed in the script?

xing-yang: @msau42 brought this to my attention.  See discussion here: https://github.com/kubernetes/kubernetes/pull/101694#issuecomment-831235149

xing-yang: We have already deleted v1beta1 from API definition.  So the auto generated code will not include v1beta1 any more.  We just keep the v1beta1 manifest identical as v5.0.1 and will not change it.

xing-yang: Every time you run the script, it also generates some changes such as new indentations even if there is no concrete change. It is hard to review them.  Just keeping v1beta1 frozen will make it easier to review.

xing-yang:
> Can someone explain why it gets done this way? I don't see that, neither here nor in the original PR, and it isn't obvious.

@msau42 brought this to my attention. See discussion here: https://github.com/kubernetes/kubernetes/pull/101694#issuecomment-831235149

ggriffiths (approved):
/lgtm

client/hack/README.md
pohly: This seems like a sub-optimal solution to me. Instead of manually keeping the CRD unchanged, a better approach might have been to bring back the Go types. Now you have a generated CRD that cannot be regenerated anymore, should the need ever arise.

Anyway, just my two cents.

xing-yang: I have asked API machinery about this and got confirmation that It is okay to delete the API definitions and only keep the manifests.
If needed, we can always refer back to the go types in earlier branches.

PR #724, opened 2022-05-28
Fix volumesnapshot webhook-example RBAC rules

astraw99:
**What type of PR is this?**
/kind bug

Adds RBAC `volumesnapshots` and `volumesnapshotcontents` to the validating webhook example, so that it can be deployed with proper access to the resources list/watch.

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/webhook-example/rbac-snapshot-webhook.yaml
+++ b/deploy/kubernetes/webhook-example/rbac-snapshot-webhook.yaml
@@ -17,7 +17,7 @@ metadata:
   name: snapshot-webhook-runner
 rules:
   - apiGroups: ["snapshot.storage.k8s.io"]
-    resources: ["volumesnapshotclasses"]
+    resources: ["volumesnapshotclasses", "volumesnapshots", "volumesnapshotcontents"]
     verbs: ["get", "list", "watch"]
 ---
 kind: ClusterRoleBinding


RaunakShah:
Can you update the description with what was failing and how this change fixes the issue?

deploy/kubernetes/webhook-example/rbac-snapshot-webhook.yaml
```diff
@@ -17,7 +17,7 @@ metadata:
   name: snapshot-webhook-runner
 rules:
   - apiGroups: ["snapshot.storage.k8s.io"]
-    resources: ["volumesnapshotclasses"]
+    resources: ["volumesnapshotclasses", "volumesnapshots", "volumesnapshotcontents"]
```
xing-yang, line 20: These additional RBAC rules should not be need here.

astraw99: The webhook will validate these 3 resources: `["volumesnapshotclasses", "volumesnapshots", "volumesnapshotcontents"]`, then how to get the access if not needed here?

xing-yang: The webhook does not do a Get or List of "VolumeSnapshots" or "VolumeSnapshotContents".  It works differently from other regular controllers.

"VolumeSnapshotClasses" is different.  There's logic added recently to check it.

ggriffiths (changes requested):
It looks like this change may not needed, can you provide info on what error you're facing in an issue?

PR #737, opened 2022-07-06
deploy/kubernetes: Update snapshot-validation-webhook

BenoitKnecht:
/kind bug

Since 4f783f10f added validation for `volumesnapshotclasses`, the
`snapshot-validation-webhook` needs to handle this kind of resource. But
version 5.0.1 doesn't, which leads to failures when updating or creating
`volumesnapshotclasses`.

This commit updates the `snapshot-validation-webhook` version to 6.0.1 in
`deploy/kubernetes/webhook-example/admission-configuration-template`.

This is a fairly trivial change, in a directory labeled as an example, so I don't think it warrants a release note.


BenoitKnecht:
> Can you also update images for snapshot-controller, csi-snapshotter, and csi-provisioner?

Done (and `hostpathplugin` as well).

PR #758, opened 2022-09-06
make the formatting errors based on go 1.19

humblec:
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

> /kind cleanup

```release-note
NONE
```

Additional Note for reviewer:

The errors in absence of this PR cause tests to fail in https://github.com/kubernetes-csi/csi-release-tools/pull/203

PR #760, opened 2022-09-14
update client package with go 1.19 formatting changes

humblec:

> /kind cleanup

```release-note
NONE
```
Additional Note for reviewer:
Eventhough we tried to minimize the changes for the unit test to pass with go 1.19 here https://github.com/kubernetes-csi/external-snapshotter/pull/758, it is not enough . 
The test is failing here https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-release-tools/203/pull-kubernetes-csi-release-tools-external-snapshotter/1569716664243589120 . This PR holds the required changes, however this has changes in client package too. Also the formatting we discussed here https://github.com/kubernetes-csi/external-snapshotter/pull/758#discussion_r964520500 getting adjusted to different representation by the gofmt tool as in this PR.

### DIFF
--- a/client/apis/volumesnapshot/v1/types.go
+++ b/client/apis/volumesnapshot/v1/types.go
@@ -123,11 +123,11 @@ type VolumeSnapshotSource struct {
 // VolumeSnapshotStatus and VolumeSnapshotContentStatus. Fields in VolumeSnapshotStatus
 // are updated based on fields in VolumeSnapshotContentStatus. They are eventual
 // consistency. These fields are duplicate in both objects due to the following reasons:
-// - Fields in VolumeSnapshotContentStatus can be used for filtering when importing a
-//   volumesnapshot.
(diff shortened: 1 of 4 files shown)


pohly (approved):
/lgtm

I remember that @xing-yang said "don't touch client", but these are all just comment changes and thus have no impact on the code.

humblec:
> /lgtm
> 
> I remember that @xing-yang said "don't touch client", but these are all just comment changes and thus have no impact on the code.

@pohly @xing-yang Looks like we need the changes to be populated even when it was "source comments" in  "types.go" and "clientset" The release tool PR fails on the same line https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-release-tools/203/pull-kubernetes-csi-release-tools-external-snapshotter/1570039353365762048 

It seems to me that, if we try update-generated code and crd , it brings other changes too.

PR #777, opened 2022-11-03
Parameterize startup CRD wait retries

mattcary:
Add a command line flag (retry-crd-interval-max) for the maximum interval to wait for CRDs to appear.

This allows providers to avoid unnecessary crashlooping on startup when it is not unusual for CRDs to take longer than 7 seconds to appear.

/kind feature

```release-note
Add --retry-crd-interval-max flag to the snapshot-controller in order to allow customization of CRD detection on startup.
```

/assign @msau42

### DIFF
--- a/cmd/snapshot-controller/main.go
+++ b/cmd/snapshot-controller/main.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"math"
 	"net"
 	"net/http"
 	"os"
@@ -71,6 +72,8 @@ var (
 	retryIntervalMax              = flag.Duration("retry-interval-max", 5*time.Minute, "Maximum retry interval of failed volume snapshot creation or deletion. Default is 5 minutes.")
 	enableDistributedSnapshotting = flag.Bool("enable-distributed-snapshotting", false, "Enables each node to handle snapshotting for the local volumes created on that node")
 	preventVolumeModeConversion   = flag.Bool("prevent-volume-mode-conversion", false, "Prevents an unauthorised user from modifying the volume mode when creating a PVC from an existing VolumeSnapshot.")
+
+	retryCRDIntervalMax = flag.Duration("retry-crd-interval-max", 5*time.Second, "Maximum retry interval to wait for CRDs to appear. The default is 5 seconds.")
 )
 
 var version = "unknown"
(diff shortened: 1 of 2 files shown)


cmd/snapshot-controller/main.go
```diff
@@ -100,11 +103,18 @@ func ensureCustomResourceDefinitionsExist(client *clientset.Clientset) error {
 		return true, nil
 	}
 
-	// with a Factor of 1.5 we wait up to 7.5 seconds (the 10th attempt)
+	maxMs := retryCRDIntervalMax.Milliseconds()
```
xing-yang, line 106: Can you add new comments here?

cmd/snapshot-controller/main.go
```diff
@@ -100,11 +103,18 @@ func ensureCustomResourceDefinitionsExist(client *clientset.Clientset) error {
...
+	if maxMs < 100 {
+		maxMs = 100
```
xing-yang, line 108: What is the relationship between this 100 ms and Duration of 100 * time.Millisecond on line 115?

cmd/snapshot-controller/main.go
```diff
@@ -100,11 +103,18 @@ func ensureCustomResourceDefinitionsExist(client *clientset.Clientset) error {
...
+	}
+	steps := int(math.Ceil(math.Log(float64(maxMs)/100) / math.Log(1.5)))
```
xing-yang, line 110: This 1.5 here is the same as Factor 1.5 on line 116?  Can you use the same parameter if they are the same?

cmd/snapshot-controller/main.go
mattcary: Done

cmd/snapshot-controller/main.go
mattcary: Explained in the comment, LMK if it's not clear.

cmd/snapshot-controller/main.go
mattcary: Yup, explained in the comment.

mattcary:
Thanks @xing-yang , lmk if it's still unclear.

mattcary:
> In the release note, can you add "in the snapshot-controller" at the end?
> 
> Can you also document this new flag in README? https://github.com/kubernetes-csi/external-snapshotter#snapshot-controller-command-line-options

Done & done!

PR #796, opened 2022-12-06
fix certificate generation in webhook example

haslersn:
After bb957259da3b the script was broken. This commit fixes this by respecting the format of TLS-type secrets in the script, too.

 /kind bug

No release note needed as the breaking commit is not yet released.

```release-note
Fixes an issue introduced by PR 793 by respecting the format of TLS-type secrets in the script.
```


deepakkinni:
This works.

/approve

xing-yang:
@haslersn Can you add a release note?

haslersn:
@xing-yang see my original post. The release note in bb957259da3b0ab6c5e27594ce15095de105676e already encompasses this change.

xing-yang:
@haslersn Please add to the release note of this PR what bug is fixed here.  We need that to make sure when doing cherry-picking, both PRs are selected.

PR #797, opened 2022-12-12
Update dependency go modules for k8s v1.26.0

sunnylovestiramisu:
Ran kubernetes-csi/csi-release-tools go-get-kubernetes.sh -p 1.26.0


client/go.mod
```diff
@@ -1,11 +1,11 @@
...
 
-go 1.18
+go 1.19
 
 require (
-	k8s.io/api v0.26.0-rc.0
-	k8s.io/apimachinery v0.26.0-rc.0
-	k8s.io/client-go v0.26.0-rc.0
+	k8s.io/api v0.26.0
+	k8s.io/apimachinery v0.26.1-rc.0
```
xing-yang, line 7: Why is this v0.26.1-rc.0, not v0.26.0?

client/go.mod
```diff
@@ -31,11 +31,11 @@ require (
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
 	golang.org/x/mod v0.6.0 // indirect
-	golang.org/x/net v0.1.1-0.20221027164007-c63010009c80 // indirect
+	golang.org/x/net v0.3.1-0.20221206200815-1e63c2f08a10 // indirect
```
xing-yang, line 34: This one is causing the trivy scan failure. Is this golang.org/x/net v0.3.1 referenced in other sidecar's go.mod?
Can you check who is using this lib?
One way to workaround this problem is to manually add a replace statement:
replace (
  golang.org/x/net => v0.4.0
)

client/go.mod
```diff
@@ -1,11 +1,11 @@
...
+go 1.19
 
 require (
-	k8s.io/api v0.26.0-rc.0
-	k8s.io/apimachinery v0.26.0-rc.0
-	k8s.io/client-go v0.26.0-rc.0
+	k8s.io/api v0.26.0
+	k8s.io/apimachinery v0.26.0
+	k8s.io/client-go v0.26.0
 	k8s.io/code-generator v0.25.2
```
sunnylovestiramisu, line 9: Does this one need to be 0.26.0 as well?

go.mod
```diff
@@ -17,11 +17,11 @@ require (
...
-	k8s.io/client-go v0.26.0-rc.0
-	k8s.io/component-base v0.26.0-rc.0
-	k8s.io/component-helpers v0.26.0-rc.0
+	k8s.io/api v0.26.0
+	k8s.io/apimachinery v0.26.1-rc.0
+	k8s.io/client-go v0.26.0
+	k8s.io/component-base v0.26.0
+	k8s.io/component-helpers v0.26.0
 	k8s.io/klog/v2 v2.80.1
 	k8s.io/kubernetes v1.25.4
```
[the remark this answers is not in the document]
sunnylovestiramisu: And because it is depending on k8s should we keep the replace generated by the script?

client/go.mod
xing-yang: This one is okay. Leave this one as is.

go.mod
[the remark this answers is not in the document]
xing-yang: Yes, this should be v1.26.0 as well.  I don't think we need the replace as it points to the same version any way.

client/go.mod
sunnylovestiramisu: Let me try to update this

go.mod
```diff
@@ -17,13 +17,13 @@ require (
 	github.com/prometheus/common v0.37.0
 	github.com/spf13/cobra v1.6.1
 	google.golang.org/grpc v1.50.1
-	k8s.io/api v0.26.0-rc.0
-	k8s.io/apimachinery v0.26.0-rc.0
-	k8s.io/client-go v0.26.0-rc.0
-	k8s.io/component-base v0.26.0-rc.0
-	k8s.io/component-helpers v0.26.0-rc.0
+	k8s.io/api v0.26.0
+	k8s.io/apimachinery v0.26.1-rc.0
```
xing-yang, line 21: Any reason why this one is v0.26.1-rc.0, not v0.26.0?

sunnylovestiramisu: Manually updated it, somehow the go-get-k8s.sh picked this one up

client/go.mod
xing-yang, line 34: @sunnylovestiramisu Can you submit another PR to make a similar change in client/go.mod as follows:
replace golang.org/x/net => golang.org/x/net v0.4.0

I'm getting a dependencies change error when generating the changelog.  I wonder if it is because the client and server go.mod are having different libs.

PR #829, opened 2023-04-16
Switch to using google.golang.org/protobuf rather than github.com/golang/protobuf

### LINKED ISSUE #804: module github.com/golang/protobuf is deprecated: Use the "google.golang.org/protobuf" module instead
**What happened**:
Depending on deprecated lib

Update to google.golang.org/protobuf

**How to reproduce it**:
N/A

Similar to https://github.com/kubernetes-csi/csi-lib-utils/issues/115, all sidecars can have this upgrade

ghakazian:
**What type of PR is this?**
/kind cleanup

This PR removes an explicit dependency on a deprecated package

Fixes #804

This is my first K8s PR so any constructive criticism would be welcome. Thanks!

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	github.com/evanphx/json-patch v4.12.0+incompatible
 	github.com/fsnotify/fsnotify v1.6.0
 	github.com/golang/mock v1.6.0
-	github.com/golang/protobuf v1.5.3
 	github.com/google/gofuzz v1.2.0
 	github.com/kubernetes-csi/csi-lib-utils v0.12.0
 	github.com/kubernetes-csi/csi-test/v4 v4.0.2
@@ -17,6 +16,7 @@ require (
 	github.com/prometheus/common v0.37.0
 	github.com/spf13/cobra v1.6.1
 	google.golang.org/grpc v1.51.0
+	google.golang.org/protobuf v1.28.1
 	k8s.io/api v0.27.0
 	k8s.io/apimachinery v0.27.0
 	k8s.io/client-go v0.27.0
@@ -38,6 +38,7 @@ require (
 	github.com/go-openapi/swag v0.22.3 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
+	github.com/golang/protobuf v1.5.3 // indirect
 	github.com/google/gnostic v0.6.9 // indirect
 	github.com/google/go-cmp v0.5.9 // indirect
 	github.com/google/uuid v1.3.0 // indirect
@@ -61,7 +62,6 @@ require (
 	golang.org/x/time v0.1.0 // indirect
 	google.golang.org/appengine v1.6.7 // indirect
 	google.golang.org/genproto v0.0.0-20221027153422-115e99e71e1c // indirect
-	google.golang.org/protobuf v1.28.1 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
--- a/pkg/snapshotter/snapshotter.go
+++ b/pkg/snapshotter/snapshotter.go
@@ -22,7 +22,6 @@ import (
 	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
-	"github.com/golang/protobuf/ptypes"
 	csirpc "github.com/kubernetes-csi/csi-lib-utils/rpc"
 
 	"google.golang.org/grpc"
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -38,6 +38,7 @@ require (
 	github.com/go-openapi/swag v0.22.3 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
+	github.com/golang/protobuf v1.5.3 // indirect
```
xing-yang, line 41: This is still used here.

ghakazian: Understood, it is still pulled in as an indirect dependency by a number of other components.

ghakazian: @xing-yang Any chance this can be approved? I removed the explicit dependency on  github.com/golang/protobuf; the remaining indirect reference comes from other components that still use github.com/golang/protobuf.
If I am missing other steps that I should do as part of this PR to remove the indirect dependency, please let me know. Thanks.

pkg/snapshotter/snapshotter.go
```diff
@@ -74,10 +73,8 @@ func (s *snapshot) CreateSnapshot(ctx context.Context, snapshotName string, volu
 	}
 
 	klog.V(5).Infof("CSI CreateSnapshot: %s driver name [%s] snapshot ID [%s] time stamp [%v] size [%d] readyToUse [%v]", snapshotName, driverName, rsp.Snapshot.SnapshotId, rsp.Snapshot.CreationTime, rsp.Snapshot.SizeBytes, rsp.Snapshot.ReadyToUse)
-	creationTime, err := ptypes.Timestamp(rsp.Snapshot.CreationTime)
-	if err != nil {
-		return "", "", time.Time{}, 0, false, err
-	}
+
+	creationTime := rsp.Snapshot.CreationTime.AsTime()
```
xing-yang, line 77: Any error check for this?  Could "creationTime" be nil?

pkg/snapshotter/snapshotter.go
```diff
@@ -138,9 +135,6 @@ func (s *snapshot) GetSnapshotStatus(ctx context.Context, snapshotID string, sna
 		return false, time.Time{}, 0, fmt.Errorf("can not find snapshot for snapshotID %s", snapshotID)
 	}
 
-	creationTime, err := ptypes.Timestamp(rsp.Entries[0].Snapshot.CreationTime)
-	if err != nil {
-		return false, time.Time{}, 0, err
-	}
+	creationTime := rsp.Entries[0].Snapshot.CreationTime.AsTime()
```
xing-yang, line 138: Same as above.

pkg/snapshotter/snapshotter_test.go
```diff
@@ -68,11 +68,8 @@ func createMockServer(t *testing.T) (*gomock.Controller, *driver.MockCSIDriver,
 func TestCreateSnapshot(t *testing.T) {
 	defaultName := "snapshot-test"
 	defaultID := "testid"
-	createTimestamp := ptypes.TimestampNow()
-	createTime, err := ptypes.Timestamp(createTimestamp)
-	if err != nil {
-		t.Fatalf("Failed to convert timestamp to time: %v", err)
-	}
+	createTimestamp := timestamppb.Now()
+	createTime := createTimestamp.AsTime()
```
xing-yang, line 72: Same as above.

pkg/snapshotter/snapshotter_test.go
```diff
@@ -338,11 +335,8 @@ func TestDeleteSnapshot(t *testing.T) {
 func TestGetSnapshotStatus(t *testing.T) {
 	defaultID := "testid"
 	size := int64(1000)
-	createTimestamp := ptypes.TimestampNow()
-	createTime, err := ptypes.Timestamp(createTimestamp)
-	if err != nil {
-		t.Fatalf("Failed to convert timestamp to time: %v", err)
-	}
+	createTimestamp := timestamppb.Now()
+	createTime := createTimestamp.AsTime()
```
xing-yang, line 339: Same as above.

go.mod
xing-yang: Can you check who is making this indirect dep?  You can try this "go mod why -m <lib>".

pkg/snapshotter/snapshotter.go
ghakazian: Response same as above, AsTime() doesn't return a nillable type or an error to check.

go.mod
ghakazian: It looks like it is being pulled in by github.com/container-storage-interface/spec/lib/go/csi

ghakazian:
@xing-yang Looking at output from `go mod why -m `, it looks like the dependency is being pulled in by github.com/container-storage-interface/spec.

pkg/snapshotter/snapshotter_test.go
ghakazian: Response same as above, AsTime() doesn't return a nillable type or an error to check.

pkg/snapshotter/snapshotter_test.go
ghakazian: Response same as above, AsTime() doesn't return a nillable type or an error to check.

go.mod
xing-yang: Thanks for checking!

PR #846, opened 2023-06-08
Fix CRDs missing metadata fields

### LINKED ISSUE #845: ObjectMeta fields missing from CRDs
The CRD yaml files in `client/config/crd` ([e.g.](https://github.com/kubernetes-csi/external-snapshotter/blob/master/client/config/crd/snapshot.storage.k8s.io_volumesnapshotclasses.yaml)) do not allow for standard metadata, according to their schemae.

However, looking at the struct definitions under `client/apis` ([e.g.](https://github.com/kubernetes-csi/external-snapshotter/blob/master/client/apis/volumesnapshot/v1/types.go#LL211C4-L211C4)), they correctly embed (and thus support) `metav1.ObjectMeta` -- and the docs reflect this as well.

This isn't problematic at first glance, because the controller recognizes and knows how to handle `ObjectMeta` in an applied resource. We only noticed it because we have a CI linter that validates applied resources against the schema as defined in the CRD.

I'll submit a PR to resolve this discrepancy.

bjschafer:
**What type of PR is this?**

/kind bug

The CRD yaml files in client/config/crd ([e.g.](https://github.com/kubernetes-csi/external-snapshotter/blob/master/client/config/crd/snapshot.storage.k8s.io_volumesnapshotclasses.yaml)) do not allow for standard metadata, according to their schemae.

However, looking at the struct definitions under client/apis ([e.g.](https://github.com/kubernetes-csi/external-snapshotter/blob/master/client/apis/volumesnapshot/v1/types.go#LL211C4-L211C4)), they correctly embed (and thus support) metav1.ObjectMeta -- and the docs reflect this as well.

This isn't problematic at first glance, because the controller recognizes and knows how to handle ObjectMeta in an applied resource. We only noticed it because we have a CI linter that validates applied resources against the schema as defined in the CRD, with strict validation enabled.

Fixes #845 

```release-note
NONE
```

### DIFF
--- a/client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
+++ b/client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
@@ -3,8 +3,8 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.3
     api-approved.kubernetes.io: "https://github.com/kubernetes-csi/external-snapshotter/pull/814"
+    controller-gen.kubebuilder.io/version: v0.12.0
   creationTimestamp: null
   name: volumegroupsnapshotclasses.groupsnapshot.storage.k8s.io
 spec:
@@ -53,6 +53,8 @@ spec:
               object represents. Servers may infer this from the endpoint the client
               submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
             type: string
+          metadata:
+            type: object
           parameters:
             additionalProperties:
               type: string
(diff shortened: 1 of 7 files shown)


client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
```diff
@@ -3,9 +3,9 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.3
     api-approved.kubernetes.io: "https://github.com/kubernetes-csi/external-snapshotter/pull/814"
   creationTimestamp: null
+    controller-gen.kubebuilder.io/version: v0.12.0
```
nixpanic, line 8: This was moved from the `annotations` section to here, that looks wrong. The same happened in the other files.

nixpanic (commented):
/lgtm

xing-yang:
@yuxiangqian, can you take a look?

/assign @yuxiangqian

PR #864, opened 2023-06-20
check if the csi driver is supported the volume group snapshot cap 

carlory:
**What type of PR is this?**

 /kind feature

check if the csi driver is supported the volume group snapshot cap when the --enable-volume-group-snapshots arg is true

waiting for csi-lib-utils 0.14.0 to be released 

/hold 

```release-note
print a warning message at the startup phase if the csi-driver doesn't support GroupControllerCreateVolumeGroupSnapshot when the --enable-volume-group-snapshots flag is true.
```


cmd/csi-snapshotter/main.go
```diff
@@ -229,6 +230,15 @@ func main() {
 	snapShotter := snapshotter.NewSnapshotter(csiConn)
 	var groupSnapshotter group_snapshotter.GroupSnapshotter
 	if *enableVolumeGroupSnapshots {
+		supportsCreateVolumeGroupSnapshot, err := supportsGroupControllerCreateVolumeGroupSnapshot(ctx, csiConn)
+		if err != nil {
+			klog.Errorf("error determining if driver supports create/delete group snapshot operations: %v", err)
+			os.Exit(1)
+		}
+		if !supportsCreateVolumeGroupSnapshot {
```
nixpanic, line 238: the support for this is optional, it probably is not correct to exit in case there is no support for VolumeGroupSnapshots.

carlory: In my opinion, storage vendors should implement the `group controller plugin` if they want to support the volume group snapshot feature. And the startup parameters of `csi-snapshotter` is specified by a storage vendor. So the storage vendor should know how to configure the `csi-snapshotter`. If they set ` --enable-volume-group-snapshots` to `true` but they don't implement the `group controller plugin`, I think it is their fault and they should fix it. 

By the way, [As @xing-yang said](https://github.com/container-storage-interface/spec/pull/545#discussion_r1235745577), the `controller plugin` is also optional for storage vendors. According to [the existing code](https://github.com/kubernetes-csi/external-snapshotter/blob/master/cmd/csi-snapshotter/main.go#L213), If the storage vendor doesn't implement the `controller plugin` or doesn't support the `csi.ControllerServiceCapability_RPC_CREATE_DELETE_SNAPSHOT` capability, the `csi-snapshotter` will exit.

If `group controller plugin` are GAed, another possible solution is that we deprecate the ` --enable-volume-group-snapshots` parameter and the `csi-snapshotter` will auto-detect the csi plugin whether supports the volume group snapshot feature at startup phase.

What do you think? @nixpanic @xing-yang

nixpanic: I agree that this PR improves the situation, and also with what you write. It is unfortunate that the `--enable-volume-group-snapshots` parameter is used, and there is no approach for dynamically detecting the support of the functionality.

It is not only storage vendors that deploy CSI-drivers, but also community projects. A mistake in the parameters for a sidecar can cause unclear problems for users. Exiting and causing a failure for a CSI-driver that otherwise works fine is not an optimal user experience in my opinion.

This is a little different from the generic snapshot support. If a CSI-driver does not implement snapshot functionality, it is rather obvious that the external-snapshotter sidecar does not need to be added to a Pod. The difference is that a CSI-driver may choose to implement GroupSnapshots with a new version, and adding `--enable-volume-group-snapshots` is forgotten.

So, to summarize my opinion:

> Adding detection for the capability is great! This also allows you to remove/deprecate the `--enable-volume-group-snapshots` parameter and simplify using the sidecar with a CSI-driver.

carlory: > I agree that this PR improves the situation, and also with what you write. It is unfortunate that the --enable-volume-group-snapshots parameter is used, and there is no approach for dynamically detecting the support of the functionality.

According to csi spec, all the plugins must be implement the indentity service RPCs. 

The GetPluginCapabilities RPC is one of them, so the external-snapshotter controller can query the GetPluginCapabilities RPC to check if the plugin supports the controller service and group controller service.

If the plugin supports the controller service, the external-snapshotter controller should check if the plugin supports the CREATE_DELETE_SNAPSHOT controller 
capability via csirpc.GetControllerCapabilities. If not, the external-snapshotter controller should fail to start.

If the plugin supports the group controller service, the external-snapshotter controller should check if the plugin supports 
the CREATE_DELETE_GET_VOLUME_GROUP_SNAPSHOT group controller capability via csirpc.GetGroupControllerCapabilities. If not, 
the external-snapshotter controller won't enable volume group snapshot feature and set the enableVolumeGroupSnapshots to false.

If the `--enable-volume-group-snapshots` flag is specified to true and the result of auto-detection is unsupported, the external-snapshotter controller should fail to start. Considering the `--enable-volume-group-snapshots` flag is used, we can log the warning message to remind the user that the flag 
will be deprecated in a future release. After 2 releases, we can remove the flag.

cc @nixpanic @xing-yang

cmd/csi-snapshotter/main.go
```diff
@@ -229,6 +230,15 @@ func main() {
...
+			klog.Errorf("CSI driver %s does not support GroupControllerCreateVolumeGroupSnapshot", driverName)
+			os.Exit(1)
```
xing-yang, line 240: This should not exit.  It is very likely that a CSI driver may combine the controller plugin and group controller plugin as one.

cmd/csi-snapshotter/main.go
xing-yang: It's unlikely that we can deprecate the `enable-volume-group-snapshots` flag. While volume snapshot is a feature that nearly every CSI driver can support, volume group snapshot is an advanced functionality.

cmd/csi-snapshotter/main.go
carlory: @xing-yang 

If this parameter `enable-volume-group-snapshots` is specified as true, but during the verification process, it is found that the group snapshot is not supported by the csi-plugins , then how should we deal with it? Print a warning log or exit directly?

xing-yang: Print a warning message but don't exit.

carlory:
@nixpanic @xing-yang please review again, thanks!

xing-yang:
Please add a release note.

carlory:
@xing-yang release note is added.

Issue #981, opened 2023-12-20
ensureCustomResourceDefinitionsExist takes too long when there is a network timeout


Fricounet:
**What happened**:
If the kube-apiserver is not reachable for whatever reason, the snapshot controller will behave like this:

```
I1220 15:19:51.451884       1 main.go:139] Version:
I1220 15:19:51.452949       1 main.go:182] Metrics path successfully registered at /metrics
E1220 15:20:21.454414       1 main.go:89] Failed to list v1 volumesnapshots with error=Get "https://172.17.0.1:443/apis/snapshot.storage.k8s.io/v1/volumesnapshots": dial tcp 172.17.0.1:443: i/o timeout
E1220 15:20:51.555570       1 main.go:89] Failed to list v1 volumesnapshots with error=Get "https://172.17.0.1:443/apis/snapshot.storage.k8s.io/v1/volumesnapshots": dial tcp 172.17.0.1:443: i/o timeout
E1220 15:21:21.706159       1 main.go:89] Failed to list v1 volumesnapshots with error=Get "https://172.17.0.1:443/apis/snapshot.storage.k8s.io/v1/volumesnapshots": dial tcp 172.17.0.1:443: i/o timeout
E1220 15:21:51.931777       1 main.go:89] Failed to list v1 volumesnapshots with error=Get "https://172.17.0.1:443/apis/snapshot.storage.k8s.io/v1/volumesnapshots": dial tcp 172.17.0.1:443: i/o timeout
E1220 15:22:22.269741       1 main.go:89] Failed to list v1 volumesnapshots with error=Get "https://172.17.0.1:443/apis/snapshot.storage.k8s.io/v1/volumesnapshots": dial tcp 172.17.0.1:443: i/o timeout
E1220 15:22:52.776373       1 main.go:89] Failed to list v1 volumesnapshots with error=Get "https://172.17.0.1:443/apis/snapshot.storage.k8s.io/v1/volumesnapshots": dial tcp 172.17.0.1:443: i/o timeout
E1220 15:23:23.536218       1 main.go:89] Failed to list v1 volumesnapshots with error=Get "https://172.17.0.1:443/apis/snapshot.storage.k8s.io/v1/volumesnapshots": dial tcp 172.17.0.1:443: i/o timeout
[...]

PR #987, opened 2023-12-22
Fix the max duration to wait for CRDs in the snapshot-controller

### LINKED ISSUE #980: Param retryCRDIntervalMax possibly not working as expected
I _think_ the naming of the argument `retryCRDIntervalMax` is a bit confusing as `interval` could mean:
1) the total duration we wait for the [ensureCustomResourceDefinitionsExist](https://github.com/kubernetes-csi/external-snapshotter/blob/master/cmd/snapshot-controller/main.go#L83) to complete
2) the max duration between 2 steps of the Backoff in the same function. Especially since the [description](https://github.com/kubernetes-csi/external-snapshotter/blob/master/cmd/snapshot-controller/main.go#L77) of the argument mentions `retry interval`

However, based on the description of the PR https://github.com/kubernetes-csi/external-snapshotter/pull/777 that introduced the change as well as [this comment](https://github.com/kubernetes-csi/external-snapshotter/blob/master/cmd/snapshot-controller/main.go#L126-L127) mentioning "retry **duration**", I believe that the true meaning of the argument is 1).
Also, if it were 2), computing the [number of steps](https://github.com/kubernetes-csi/external-snapshotter/blob/master/cmd/snapshot-controller/main.go#L134) with this argument makes little sense IMHO

If that is the case, I think there is a bug in the code computing the number of steps:
```
// The maximum retry duration = initial duration * retry factor ^ # steps. Rearranging, this gives
// # steps = log(maximum retry / initial duration) / log(retry factor).
```
Basically, this computation is not for the **number of steps** (which is how I interpret the `# steps`) we need to reach the retry duration but is the **Nth step** at which point the backoff between two steps of the wait lasts the retry duration :sweat_smile:  

The true computation of the total duration is just the sum of a geometric serie with `n` the number of steps needed:

```math
\left\{ 
\begin{aligned}
n=1 \Rightarrow  &T_{total} = 0 \ (no\ backoff\ needed\ since\ there\ is\ only\ one\ step) \\
n\ge2 \Rightarrow  &T_{total} = \sum_{k=0}^{n-2} initial\ duration * retry\ factor ^ k
\end{aligned}
\right.
```

If we rearrange the equation, it gives us (feel free to double check me):
```math
n = 1 + \frac {\ln(1 - \frac{T_{total}(1-retry\ factor)} {initial\ duration})} {\ln(retry\ factor)}
```

**What happened**:

If my interpretation of what the argument is used for is correct, I noticed that with the default value, the pod exits only after about 15sec (this [calculator](https://exponentialbackoffcalculator.com/) gives me 17sec to reach at least 5sec between steps)

The pod should exit after 5sec with the default value of the arg

**How to reproduce it**:
Have a cluster without the snapshots CRD and start the snapshot controller. It will fait after about 15sec of running

If my interpretation is correct, I'd like to submit a PR to fix it :D
If it is not and 2) is the correct interpretation, I think there is a discussion to be had on why we would want to have this flag compared to having a flag allowing us to directly configure the time we wait for the CRDs.

**Environment**:
- Snaphot controller version: v6.3.2
- OS (e.g. from /etc/os-release): linux
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Fricounet:
**What type of PR is this?**

/kind bug

Based on my understanding of the flag `retry-crd-interval-max` (see #980), this PR will ensure that the function `ensureCustomResourceDefinitionsExist` returns when the duration specified in `retryCRDIntervalMax` flag is reached. This works using by passing a context to the backoff so that we can simply timeout with context.WithTimeout (and not too complicated computations that may or may not do what we want).
Also limit the amount of resources returned by the list calls to 0 to speed up the operations because we only care that the CRD exists and not about the number of results.
It changes the default value of the flag to `30s` for 2 reasons:
- the timeout includes a bunch network calls to the apiserver. Network being notoriously slower than CPU, I believe 30s is the default timeout for api calls so it gives us time to wait for 1 complete timeout
- this will give more time for the CRDs to appear when they are created at the same time as the snapshot controller

With this PR, the new behavior is as follow:
- normal:
```
❯ kubectl logs snapshot-controller-5699b76477-vz6v5
I1222 13:43:13.513998       1 main.go:140] Version:
I1222 13:43:13.514943       1 main.go:183] Metrics path successfully registered at /metrics
I1222 13:43:13.553446       1 main.go:248] Metrics http server successfully started on :8080, /metrics
I1222 13:43:13.553709       1 leaderelection.go:250] attempting to acquire leader lease kube-system/snapshot-controller-leader...
I1222 13:43:43.435001       1 leaderelection.go:260] successfully acquired lease kube-system/snapshot-controller-leader
I1222 13:43:43.436397       1 snapshot_controller_base.go:212] Starting snapshot controller
```
- network timeout (fails after 10s and 1 step):
```
❯ kubectl logs snapshot-controller-5699b76477-crr4z
I1222 13:33:29.278744       1 main.go:140] Version:
I1222 13:33:29.280018       1 main.go:183] Metrics path successfully registered at /metrics
E1222 13:33:39.289596       1 main.go:94] Failed to list v1 volumesnapshots with error=Get "https://172.17.0.1:443/apis/snapshot.storage.k8s.io/v1/volumesnapshots": context deadline exceeded
E1222 13:33:39.289628       1 main.go:217] Exiting due to failure to ensure CRDs exist during startup: context deadline exceeded
```
- missing CRD (fails after 10s and 10 steps):
```
❯ kubectl logs snapshot-controller-5699b76477-xcpj9
I1222 13:39:24.679192       1 main.go:140] Version:
I1222 13:39:24.680276       1 main.go:183] Metrics path successfully registered at /metrics
E1222 13:39:24.693060       1 main.go:94] Failed to list v1 volumesnapshots with error=volumesnapshots.snapshot.storage.k8s.io is forbidden: User "system:serviceaccount:snapshot-controller" cannot list resource "volumesnapshots" in API group "snapshot.storage.k8s.io" at the cluster scope
E1222 13:39:24.794276       1 main.go:94] Failed to list v1 volumesnapshots with error=volumesnapshots.snapshot.storage.k8s.io is forbidden: User "system:serviceaccount:snapshot-controller" cannot list resource "volumesnapshots" in API group "snapshot.storage.k8s.io" at the cluster scope
E1222 13:39:24.945760       1 main.go:94] Failed to list v1 volumesnapshots with error=volumesnapshots.snapshot.storage.k8s.io is forbidden: User "system:serviceaccount:snapshot-controller" cannot list resource "volumesnapshots" in API group "snapshot.storage.k8s.io" at the cluster scope
E1222 13:39:25.172186       1 main.go:94] Failed to list v1 volumesnapshots with error=volumesnapshots.snapshot.storage.k8s.io is forbidden: User "system:serviceaccount:snapshot-controller" cannot list resource "volumesnapshots" in API group "snapshot.storage.k8s.io" at the cluster scope
E1222 13:39:25.511779       1 main.go:94] Failed to list v1 volumesnapshots with error=volumesnapshots.snapshot.storage.k8s.io is forbidden: User "system:serviceaccount:snapshot-controller" cannot list resource "volumesnapshots" in API group "snapshot.storage.k8s.io" at the cluster scope
E1222 13:39:26.019556       1 main.go:94] Failed to list v1 volumesnapshots with error=volumesnapshots.snapshot.storage.k8s.io is forbidden: User "system:serviceaccount:snapshot-controller" cannot list resource "volumesnapshots" in API group "snapshot.storage.k8s.io" at the cluster scope
E1222 13:39:26.781178       1 main.go:94] Failed to list v1 volumesnapshots with error=volumesnapshots.snapshot.storage.k8s.io is forbidden: User "system:serviceaccount:snapshot-controller" cannot list resource "volumesnapshots" in API group "snapshot.storage.k8s.io" at the cluster scope
E1222 13:39:27.921712       1 main.go:94] Failed to list v1 volumesnapshots with error=volumesnapshots.snapshot.storage.k8s.io is forbidden: User "system:serviceaccount:snapshot-controller" cannot list resource "volumesnapshots" in API group "snapshot.storage.k8s.io" at the cluster scope
E1222 13:39:29.631696       1 main.go:94] Failed to list v1 volumesnapshots with error=volumesnapshots.snapshot.storage.k8s.io is forbidden: User "system:serviceaccount:snapshot-controller" cannot list resource "volumesnapshots" in API group "snapshot.storage.k8s.io" at the cluster scope
E1222 13:39:32.196718       1 main.go:94] Failed to list v1 volumesnapshots with error=volumesnapshots.snapshot.storage.k8s.io is forbidden: User "system:serviceaccount:snapshot-controller" cannot list resource "volumesnapshots" in API group "snapshot.storage.k8s.io" at the cluster scope
E1222 13:39:34.681660       1 main.go:217] Exiting due to failure to ensure CRDs exist during startup: context deadline exceeded
```

Fixes #980 and #981 

Maybe this would need a separate flag like `retry-crd-max-duration` instead of reusing the previous unclear `retry-crd-interval-max` flag but I'm not sure. Open to suggestions on that.

```release-note
Fixed the max duration to wait for CRDs to appear especially in case of the apiserver being unreachable
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ Read more about how to install the example webhook [here](deploy/kubernetes/webh
 
 ##### Volume Snapshot Classes
 
-* There can only be a single default volume snapshot class for a particular driver. 
+* There can only be a single default volume snapshot class for a particular driver.
 
 ### Distributed Snapshotting
 
@@ -139,11 +139,11 @@ The distributed snapshotting feature is provided to handle snapshot operations f
 
 #### Snapshot controller option
 
-* `--enable-distributed-snapshotting`: This option lets the snapshot controller know that distributed snapshotting is enabled and the snapshotter sidecar will be running on each node. Off by default. 
+* `--enable-distributed-snapshotting`: This option lets the snapshot controller know that distributed snapshotting is enabled and the snapshotter sidecar will be running on each node. Off by default.
 
 #### CSI external snapshotter sidecar option
 
-* `--node-deployment`: Enables the snapshotter sidecar to handle snapshot operations for the volumes local to the node on which it is deployed. Off by default. 
+* `--node-deployment`: Enables the snapshotter sidecar to handle snapshot operations for the volumes local to the node on which it is deployed. Off by default.
 
 Other than this, the NODE_NAME environment variable must be set where the CSI snapshotter sidecar is deployed. The value of NODE_NAME should be the name of the node where the sidecar is running.
 
@@ -174,7 +174,7 @@ Other than this, the NODE_NAME environment variable must be set where the CSI sn
 
 * `--retry-interval-max`: Maximum retry interval of failed volume snapshot creation or deletion. Default value is 5 minutes.
 
-* `--retry-crd-interval-max`: Maximum retry interval for detecting the snapshot CRDs on controller startup. Default is 5 seconds.
+* `--retry-crd-interval-max`: Maximum retry duration for detecting the snapshot CRDs on controller startup. Default is 30 seconds.
 
 * `--enable-distributed-snapshotting` : Enables each node to handle snapshots for the volumes local to that node. Off by default. It should be set to true only if `--node-deployment` parameter for the csi external snapshotter sidecar is set to true. See https://github.com/kubernetes-csi/external-snapshotter/blob/master/README.md#distributed-snapshotting for details.
 
--- a/cmd/snapshot-controller/main.go
+++ b/cmd/snapshot-controller/main.go
@@ -74,46 +74,49 @@ var (
 	preventVolumeModeConversion   = flag.Bool("prevent-volume-mode-conversion", true, "Prevents an unauthorised user from modifying the volume mode when creating a PVC from an existing VolumeSnapshot.")
 	enableVolumeGroupSnapshots    = flag.Bool("enable-volume-group-snapshots", false, "Enables the volume group snapshot feature, allowing the user to create a snapshot of a group of volumes.")
 
-	retryCRDIntervalMax = flag.Duration("retry-crd-interval-max", 5*time.Second, "Maximum retry interval to wait for CRDs to appear. The default is 5 seconds.")
+	retryCRDIntervalMax = flag.Duration("retry-crd-interval-max", 30*time.Second, "Maximum time to wait for CRDs to appear. The default is 30 seconds.")
 )
 
 var version = "unknown"
 
-// Checks that the VolumeSnapshot v1 CRDs exist.
+// Checks that the VolumeSnapshot v1 CRDs exist. It will wait at most the duration specified by retryCRDIntervalMax
 func ensureCustomResourceDefinitionsExist(client *clientset.Clientset, enableVolumeGroupSnapshots bool) error {
-	condition := func() (bool, error) {
+	condition := func(ctx context.Context) (bool, error) {
 		var err error
+		// List calls should return faster with a limit of 0.
+		// We do not care about what is returned and just want to make sure the CRDs exist.
+		listOptions := metav1.ListOptions{Limit: 0}
 
 		// scoping to an empty namespace makes `List` work across all namespaces
-		_, err = client.SnapshotV1().VolumeSnapshots("").List(context.TODO(), metav1.ListOptions{})
+		_, err = client.SnapshotV1().VolumeSnapshots("").List(ctx, listOptions)
 		if err != nil {
 			klog.Errorf("Failed to list v1 volumesnapshots with error=%+v", err)
 			return false, nil
 		}
 
-		_, err = client.SnapshotV1().VolumeSnapshotClasses().List(context.TODO(), metav1.ListOptions{})
+		_, err = client.SnapshotV1().VolumeSnapshotClasses().List(ctx, listOptions)
 		if err != nil {
 			klog.Errorf("Failed to list v1 volumesnapshotclasses with error=%+v", err)
 			return false, nil
 		}
-		_, err = client.SnapshotV1().VolumeSnapshotContents().List(context.TODO(), metav1.ListOptions{})
+		_, err = client.SnapshotV1().VolumeSnapshotContents().List(ctx, listOptions)
 		if err != nil {
 			klog.Errorf("Failed to list v1 volumesnapshotcontents with error=%+v", err)
 			return false, nil
 		}
 		if enableVolumeGroupSnapshots {
-			_, err = client.GroupsnapshotV1alpha1().VolumeGroupSnapshots("").List(context.TODO(), metav1.ListOptions{})
+			_, err = client.GroupsnapshotV1alpha1().VolumeGroupSnapshots("").List(ctx, listOptions)
 			if err != nil {
 				klog.Errorf("Failed to list v1alpha1 volumegroupsnapshots with error=%+v", err)
 				return false, nil
 			}
 
-			_, err = client.GroupsnapshotV1alpha1().VolumeGroupSnapshotClasses().List(context.TODO(), metav1.ListOptions{})
+			_, err = client.GroupsnapshotV1alpha1().VolumeGroupSnapshotClasses().List(ctx, listOptions)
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -172,7 +172,7 @@ Other than this, the NODE_NAME environment variable must be set where the CSI sn
 
 * `--retry-interval-start`: Initial retry interval of failed volume snapshot creation or deletion. It doubles with each failure, up to retry-interval-max. Default value is 1 second.
 
-* `--retry-interval-max`: Maximum retry interval of failed volume snapshot creation or deletion. Default value is 5 minutes.
+* `--retry-interval-max`: Maximum time to wait for CRDs to appear. Default is 10 seconds.
```
xing-yang, line 175: This retry-interval-max is also for the failed volume snapshot creation/deletion retry interval, not just for controller startup. 10 seconds is too short for those operations.

xing-yang:
Thanks for submitting the PR. Regarding this issue, we discussed having a readiness probe to check if the controller is up and that the controller should keep retrying rather than exiting.

cc @msau42 @jsafrane

README.md
Fricounet: Sorry, I had not modified the correct part of the README, it should e better now :sweat_smile:

xing-yang:
@mattcary Can you take a look?

cmd/snapshot-controller/main.go
```diff
@@ -74,46 +74,49 @@ var (
 	preventVolumeModeConversion   = flag.Bool("prevent-volume-mode-conversion", true, "Prevents an unauthorised user from modifying the volume mode when creating a PVC from an existing VolumeSnapshot.")
 	enableVolumeGroupSnapshots    = flag.Bool("enable-volume-group-snapshots", false, "Enables the volume group snapshot feature, allowing the user to create a snapshot of a group of volumes.")
 
-	retryCRDIntervalMax = flag.Duration("retry-crd-interval-max", 5*time.Second, "Maximum retry interval to wait for CRDs to appear. The default is 5 seconds.")
+	retryCRDIntervalMax = flag.Duration("retry-crd-interval-max", 10*time.Second, "Maximum time to wait for CRDs to appear. The default is 10 seconds.")
```
mattcary, line 77: Would 30s be a better default? Sometimes CRDs and the controller are all started at the same time, so a bit more slack on allowing everything to come up would be better.

This longer delay might only make sense if we have a readiness probe --- will a 30s default cause problems for helm as you mentioned in #981?

mattcary:
Thanks! I also prefer the straightforward timeout rather than complicated (incorrect!) math.

Just want to close on the default & confirm about the state of the readiness probe.

cmd/snapshot-controller/main.go
Fricounet: 30s works for me. This shouldn't cause any problems related to #981 as long as `minReadySeconds` is above this value in the chart. Like this helm (or whatever deployment system waiting for pod readiness) will wait at least `minReadySeconds` before moving on.

I do think that adding a readiness probe is a better way to achieve the goal of 'waiting some time before sending the pod in crashloop'. I don't mind making another PR to add one so that we can maybe deprecate the flag `retryCRDIntervalMax` and just rely on the probe during startup.

mattcary:
sgtm on the plan for the probe.

Thank you!

/lgtm

I think Xing will have to approve.

README.md
```diff
@@ -174,7 +174,7 @@ Other than this, the NODE_NAME environment variable must be set where the CSI sn
 
 * `--retry-interval-max`: Maximum retry interval of failed volume snapshot creation or deletion. Default value is 5 minutes.
 
-* `--retry-crd-interval-max`: Maximum retry interval for detecting the snapshot CRDs on controller startup. Default is 5 seconds.
+* `--retry-crd-interval-max`: Maximum retry duration for detecting the snapshot CRDs on controller startup. Default is 10 seconds.
```
xing-yang, line 177: 30 seconds

xing-yang:
Can you also update your PR description?

Fricounet:
done, I also squashed commits for a cleaner history

PR #837, opened 2023-05-25
implement GroupSnapshotter.GetGroupSnapshotStatus()

nixpanic:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

GroupSnapshots are not fully implemented yet.

The commit in this PR makes it work nicely:

```yaml
---
apiVersion: groupsnapshot.storage.k8s.io/v1alpha1
kind: VolumeGroupSnapshotContent
metadata:
  name: manual-vgs-content
spec:
  deletionPolicy: Retain
  driver: hostpath.csi.k8s.io
  volumeGroupSnapshotRef:
    apiVersion: groupsnapshot.storage.k8s.io/v1alpha1
    kind: VolumeGroupSnapshot
    name: manual-vgs
    namespace: default
  volumeGroupSnapshotClassName: csi-hostpath-vgsc
  source:
    groupSnapshotHandles:
      volumeGroupSnapshotHandle: 41a78530-bb86-11ee-93b4-0a580a810320
      volumeSnapshotHandles:
        - 41a78c22-bb86-11ee-93b4-0a580a810320
---
apiVersion: groupsnapshot.storage.k8s.io/v1alpha1
kind: VolumeGroupSnapshot
metadata:
  name: manual-vgs
spec:
  volumeGroupSnapshotClassName: csi-hostpath-vgsc
  source:
    volumeGroupSnapshotContentName: manual-vgs-content
```

Results after applying the above YAML for the VolumeGroupSnapshotContent:
```console
$ kubectl get --show-managed-fields=false -oyaml volumegroupsnapshotcontent/manual-vgs-content
apiVersion: groupsnapshot.storage.k8s.io/v1alpha1
kind: VolumeGroupSnapshotContent
metadata:
  creationTimestamp: "2024-01-25T13:35:53Z"
  finalizers:
  - groupsnapshot.storage.kubernetes.io/volumegroupsnapshotcontent-bound-protection
  generation: 2
  name: manual-vgs-content
  resourceVersion: "1708778"
  uid: ca3b4443-1fdd-4a34-bb2f-eaa5db101439
spec:
  deletionPolicy: Retain
  driver: hostpath.csi.k8s.io
  source:
    groupSnapshotHandles:
      volumeGroupSnapshotHandle: 41a78530-bb86-11ee-93b4-0a580a810320
      volumeSnapshotHandles:
      - 41a78c22-bb86-11ee-93b4-0a580a810320
  volumeGroupSnapshotClassName: csi-hostpath-vgsc
[...]

PR #995, opened 2024-01-26
Selector in VolumeGroupSnapshotSource API should be optional

nixpanic:
**What type of PR is this?**

/kind api-change

When using a VolumeGroupSnapshotContent for a pre-preprovisioned VolumeGroupSnapshotSource, there is no correct value for the `selector` property. However, the original API requires the `selector` to be set, even though it is not used.

This problem was identified in #837.

```release-note
The VolumeGroupSnapshotSource.Selector is now an optional attribute, so that a pre-provisioned VolumeGroupSnapshotContent can be specified which does not require a matching label-selector.
```


nixpanic:
/cc xing-yang RaunakShah

Please have a look, combined with #837 the VolumeGroupSnapshot feature should be pretty much ready for users to try out.

client/apis/volumegroupsnapshot/v1alpha1/types.go
```diff
@@ -52,8 +52,8 @@ type VolumeGroupSnapshotSource struct {
 	// is created, the existing group snapshots won't be modified.
 	// Once a VolumeGroupSnapshotContent is created and the sidecar starts to process
 	// it, the volume list will not change with retries.
-	// Required.
-	Selector metav1.LabelSelector `json:"selector" protobuf:"bytes,1,opt,name=selector"`
+	// +optional
+	Selector *metav1.LabelSelector `json:"selector" protobuf:"bytes,1,opt,name=selector"`
```
xing-yang, line 56: It should have "omitempty".

nixpanic: That's been done now too.

RaunakShah (approved):
Thanks for catching this!

PR #1015, opened 2024-02-22
Fix detecting default annotation on VolumeGroupSnapshotClass

### LINKED ISSUE #968: Is the check for default volumegroupsnapshotclass looking for the correct annotation?
**What happened**:

If I create a volumegroupsnapshot without specifying the `volumeGroupSnapshotClassName`, the controller tries to find the default volumegroupsnapshot class but I'm not sure it's looking for the correct annotation.

From here: https://github.com/kubernetes-csi/external-snapshotter/blob/0ce7955655dc6fd2533a941ebe4b64257b418b9c/pkg/utils/util.go#L85

It looks to me like the expected annotation to use on a VolumeGroupSnapshotClass is: `groupsnapshot.storage.kubernetes.io/is-default-class`

But in fact it's still checking for the snapshot annotation: `snapshot.storage.kubernetes.io/is-default-class` instead.

It looks like the common IsDefaultAnnotation() function is called here: https://github.com/kubernetes-csi/external-snapshotter/blob/0ce7955655dc6fd2533a941ebe4b64257b418b9c/pkg/common-controller/groupsnapshot_controller_helper.go#L143

Which ends up looking for the snapshot annotation :

https://github.com/kubernetes-csi/external-snapshotter/blob/0ce7955655dc6fd2533a941ebe4b64257b418b9c/pkg/utils/util.go#L263-L269

**How to reproduce it**:

**Environment**:
- Driver version: Using latest code in master
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

nixpanic:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

The annotation to use the selected VolumeGroupSnapshotClass by default is expected to be `groupsnapshot.storage.kubernetes.io/is-default-class=true`, but the check was done on `snapshot.storage.kubernetes.io/is-default-class` instead.

Fixes #968

```release-note
Adding the annotation `groupsnapshot.storage.kubernetes.io/is-default-class=true` on a VolumeGroupSnapshotClass will cause that VolumeGroupSnapshotClass to be used by default.
```

### DIFF
--- a/pkg/common-controller/groupsnapshot_controller_helper.go
+++ b/pkg/common-controller/groupsnapshot_controller_helper.go
@@ -140,7 +140,7 @@ func (ctrl *csiSnapshotCommonController) SetDefaultGroupSnapshotClass(groupSnaps
 
 	defaultClasses := []*crdv1alpha1.VolumeGroupSnapshotClass{}
 	for _, groupSnapshotClass := range list {
-		if utils.IsDefaultAnnotation(groupSnapshotClass.ObjectMeta) && pvDriver == groupSnapshotClass.Driver {
+		if utils.IsDefaultAnnotation(groupSnapshotClass.TypeMeta, groupSnapshotClass.ObjectMeta) && pvDriver == groupSnapshotClass.Driver {
 			defaultClasses = append(defaultClasses, groupSnapshotClass)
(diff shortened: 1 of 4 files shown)


nixpanic:
wow, `size/L` is rather unexpected. The actual change is 17 lines, the added unit-test blows things up a little
image

PR #876, opened 2023-07-12
Update VolumeSnapshot and VolumeSnapshotContent using JSON patch

shubham-pampattiwar:
**What type of PR is this?**
 /kind bug

**What this PR does / why we need it**:
This PR replaces the Update calls for VolumeSnapshot and VolumeSnapshotContent with JSON patch calls. Addresses the "object has been modified" issue we see a lot in the snapshot-controller/snapshotter.

**Which issue(s) this PR fixes**:

Partial fix (only for Update() calls) https://github.com/kubernetes-csi/external-snapshotter/issues/748

**Special notes for your reviewer**:
N/A

**Does this PR introduce a user-facing change?**:

```release-note
Update VolumeSnapshot and VolumeSnapshotContent using JSON patch
```

PR #1019, opened 2024-02-22
Use patch instead of update for GroupSnapshots, VolumeSnapshots, PVCs

### LINKED ISSUE #748: snapshot-controller logs report failure frequently
**What happened**:

In our csi-driver snapshot controller logs shows below error. There is no impact on functionality. But there are too many of these errors.

```
snapshot controller failed to update ...xxxxx
the object has been modified; please apply your changes to the latest version and try again
```

**How to reproduce it**:

**Environment**: I tested in IKs 1.22,1.23,1.24
- Driver version: Our csi driver is not GA
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:
[gcr.io/k8s-staging-sig-storage/csi-snapshotter:v6.0.1](http://gcr.io/k8s-staging-sig-storage/csi-snapshotter:v6.0.1)
[gcr.io/k8s-staging-sig-storage/snapshot-controller:v6.0.1](http://gcr.io/k8s-staging-sig-storage/snapshot-controller:v6.0.1)

kaovilai:
Signed-off-by: Tiger Kaovilai <tkaovila@redhat.com>

**What type of PR is this?**
>
> /kind api-change
/kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR cleans up some of the unit test reactor code, and eliminate update calls that I can see, fixing unit tests to accommodate the changes.
Previously update call was required simply cause unit test is borked. Patch calls was modifying original input causing compare failures.

Fixes #748

```release-note

```

This PR extend https://github.com/kubernetes-csi/external-snapshotter/pull/876 work to later added update calls.

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -1506,23 +1506,12 @@ func (ctrl *csiSnapshotCommonController) addSnapshotFinalizer(snapshot *crdv1.Vo
 	var updatedSnapshot *crdv1.VolumeSnapshot
 	var err error
 
-	// var patches []utils.PatchOp
 	finalizersToAdd := []string{}
 	// If finalizers exist already, add new ones to the end of the array
 	if addSourceFinalizer {
-		// patches = append(patches, utils.PatchOp{
-		// 	Op:    "add",
-		// 	Path:  "/metadata/finalizers/-",
-		// 	Value: utils.VolumeSnapshotAsSourceFinalizer,
-		// })
 		finalizersToAdd = append(finalizersToAdd, utils.VolumeSnapshotAsSourceFinalizer)
 	}
 	if addBoundFinalizer {
-		// patches = append(patches, utils.PatchOp{
-		// 	Op:    "add",
-		// 	Path:  "/metadata/finalizers/-",
-		// 	Value: utils.VolumeSnapshotBoundFinalizer,
-		// })
 		finalizersToAdd = append(finalizersToAdd, utils.VolumeSnapshotBoundFinalizer)
 	}
 	patches := utils.PatchOpsToAddFinalizers(snapshot, finalizersToAdd...)
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -937,12 +937,13 @@ func (ctrl *csiSnapshotCommonController) ensurePVCFinalizer(snapshot *crdv1.Volu
 		if err != nil {
 			return newControllerUpdateError(pvcClone.Name, err.Error())
 		}
-		_, err = ctrl.client.CoreV1().PersistentVolumeClaims(pvcClone.Namespace).Patch(context.TODO(), pvcClone.Name, types.JSONPatchType, patchBytes, metav1.PatchOptions{})
+		pvcCloneAfter, err := ctrl.client.CoreV1().PersistentVolumeClaims(pvcClone.Namespace).Patch(context.TODO(), pvcClone.Name, types.JSONPatchType, patchBytes, metav1.PatchOptions{})
 		if err != nil {
 			klog.Errorf("cannot add finalizer on claim [%s/%s] for snapshot [%s/%s]: [%v]", pvc.Namespace, pvc.Name, snapshot.Namespace, snapshot.Name, err)
 			return newControllerUpdateError(pvcClone.Name, err.Error())
 		}
 		klog.Infof("Added protection finalizer to persistent volume claim %s/%s", pvc.Namespace, pvc.Name)
+		fmt.Printf("pvcCloneAfter: %+v", pvcCloneAfter)
 	}
 
 	return nil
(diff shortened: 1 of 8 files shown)


pkg/common-controller/snapshot_controller.go
```diff
@@ -1570,21 +1570,34 @@ func (ctrl *csiSnapshotCommonController) removeSnapshotFinalizer(snapshot *crdv1
...
+	newSnapshot := newObj.(*crdv1.VolumeSnapshot)
+	fmt.Printf("afterpatch: original snapshot finalizer address %s\n", &snapshot.Finalizers)
+	fmt.Printf("afterpatch: clone snapshot finalizer address %s\n", &snapshotClone.Finalizers)
+	fmt.Printf("afterpatch: new snapshot finalizer address %s\n", &newSnapshot.Finalizers)
+	// output from test TestSnapshotFinalizer, deepcopy doesn't work ???
+	// b4patch: original snapshot finalizer address &[snapshot.storage.kubernetes.io/volumesnapshot-as-source-protection snapshot.storage.kubernetes.io/volumesnapshot-bound-protection]
+	// b4patch: clone snapshot finalizer address &[snapshot.storage.kubernetes.io/volumesnapshot-as-source-protection snapshot.storage.kubernetes.io/volumesnapshot-bound-protection]
+	// afterpatch: original snapshot finalizer address &[]
+	// afterpatch: clone snapshot finalizer address &[snapshot.storage.kubernetes.io/volumesnapshot-as-source-protection snapshot.storage.kubernetes.io/volumesnapshot-bound-protection]
+	// afterpatch: new snapshot finalizer address &[]
```
kaovilai, line 1600: no idea what's going on here..

pkg/utils/patch.go
```diff
@@ -97,3 +102,101 @@ func PatchVolumeGroupSnapshotContent(
...
+	if len(obj.Finalizers) == 0 {
+		// to satisfy some tests that requires nil rather than []string{}
+		obj.Finalizers = nil
+	}
+	return obj, err
+}
+
+func PatchOpsToRemoveFinalizers(object metav1.Object, finalizers ...string) []PatchOp {
+	patches := []PatchOp{}
+	if len(finalizers) == 0 {
```
kaovilai, line 149: This might cause confusion, so I'll remove this part.

pkg/utils/patch.go
```diff
@@ -97,3 +102,94 @@ func PatchVolumeGroupSnapshotContent(
...
+		}
+	}
+	slices.Sort(indexes)
+	slices.Reverse(indexes)
+	for _, i := range indexes {
+		patches = append(patches, PatchOp{
+			Op:   "remove",
+			Path: "/metadata/finalizers/" + fmt.Sprint(i),
+		})
+	}
```
jsafrane, line 171: This is racy. Consider another controller / user that added a finalizer to an object when this loop runs. Since the PATCH removes indices, it will remove a wrong item.

Is there a way how to remove a _value_ and not an index via `Patch()`? If not, then please stick to `Update()`

kaovilai: I believe there is.
One is to remove all then add all back in one call. Let me try that.

kaovilai: If this is still considered racy, then will have to wait for https://github.com/json-patch/json-patch2/issues/18 and will move back to update. lmk.

kaovilai: It probably still is racy.. hmm. but at least it external-snapshotter won't hang. User could see the finalizer they added isn't there. but at least this won't have the "wrong index" issue.

Another would be to use update but test it can get out of the "out of date, please apply again"

shubham-pampattiwar:
Thank you @kaovilai !!

pkg/utils/patch.go
kaovilai: Used delete all + add which would not be stuck in [`the object has been modified; please apply your changes to the latest version and try again`](https://github.com/kubernetes-sigs/kubebuilder/blob/66c3572c9b4b6b918c129eaa5b27abe842c54c2a/docs/book/src/reference/good-practices.md?plain=1#L51)

Any other watchers should wait until the objects have finalizers external-snapshotter added before doing it's own patching/updating to avoid race condition removing "user" added labels.

https://github.com/kubernetes-csi/external-snapshotter/pull/1019/commits/2e56da847422c99bc5c5dc88a65e80cd74436fc5

gnufied: If a controller is running into "the object has been modified" errors, then we should find out why is that happening. IMO - using unilateral patch is kind of a band-aid (which is race prone and dangerous).

kaovilai: I expect many more projects to add volumesnapshots to their watch and perform processing. If you insists patch is dangerous here, then there needs to be tests that ensures update fail calls are recoverable within few seconds, not over 10 minutes which is what we have been seeing.

https://github.com/vmware-tanzu/velero/issues/7356

kaovilai: @jsafrane @gnufied see if you like https://github.com/kubernetes-csi/external-snapshotter/pull/1023 better.

jsafrane: If I understand it correctly, the current version does not fix the race. It blindly removes all finalizers and add those that were known at the time the controller processed VolumeSnapshot / VolumeSnapshotContent.
It will again erase any finalizers added in parallel to the controller.

> While it is a bandaid, it is better than nothing

No. You are breaking finalizers of someone else. This is not a good behavior.

I think the whole fear of "the object has been modified" error is unjustified. It tells the controller that it has been working with stale data. The controller should check what has changed and try again, if it's still applicable. Or it may discover that the work is not needed any longer. In this case, the VolumeSnapshot / Content should be re-queued with exp. backoff.

> there needs to be tests that ensures update fail calls in external-snapshotter are recoverable within few seconds, not over 10 minutes which is what we have been seeing.

If that's true then this is the part that needs to be fixed. And not work around it using patch blindly everywhere. Can you reliably reproduce the issue, e.g. in an unit test? It should be easy to debug what's causing the delay then.

kaovilai: Yes we've been hitting this issue everywhere in our CI and other velero users have been hitting it as well.

kaovilai:
```
		{
			name:             "2-4 - successful remove Snapshot finalizer after update conflict",
			initialSnapshots: newSnapshotArray("snap2-4", "snapuid2-4", "claim2-4", "", classSilver, "", &False, nil, nil, nil, false, true, nil),
			initialClaims:    newClaimArray("claim2-4", "pvc-uid2-4", "1Gi", "volume2-4", v1.ClaimBound, &classEmpty),
			test:             testRemoveSnapshotFinalizerAfterUpdateConflict,
			expectSuccess:    true,
			errors: []reactorError{
				{"update", "volumesnapshots", errors.NewConflict(crdv1.Resource("volumesnapshots"), "snap2-4", nil)},
			},
		},
```
Added this case here.
https://github.com/kubernetes-csi/external-snapshotter/pull/1023#pullrequestreview-1902552534

kaovilai: > reliably reproduce the issue

50% of the time prior to https://github.com/kubernetes-csi/external-snapshotter/pull/876
After 876 it's much improved but added update calls after that OCP QE noted still cause issues sometimes.

kaovilai:
Closing due to other priorities as the sense is this isn't going to be approved as is and to see if #876 recently cherrypicked to our nightlies would be sufficient. Feel free to takeover.

Test improvements are moved to #1024

PR #1064, opened 2024-04-13
delete the content if creation failed and snapshot deleted

huww98:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

This is useful for Retain policy snapshot. When volume snapshot is created with invalid config, then deleted, all things should be cleaned up, instead of leaving a content to retry creation infinitely.

Fixes #

Some cleanups are included, please see commit message.

A new RBAC permission is added.

```release-note
No retry of creation is performed after the VolumeSnapshot is deleted.
```

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -149,31 +149,6 @@ func (ctrl *csiSnapshotCommonController) syncContent(content *crdv1.VolumeSnapsh
 			ctrl.snapshotQueue.Add(snapshotName)
 		}
 	}
-
-	// NOTE(xyang): Do not trigger content deletion if
-	// snapshot is nil. This is to avoid data loss if
-	// the user copied the yaml files and expect it to work
-	// in a different setup. In this case snapshot is nil.
-	// If we trigger content deletion, it will delete
-	// physical snapshot resource on the storage system
-	// and result in data loss!
-	//
-	// Trigger content deletion if snapshot is not nil
-	// and snapshot has deletion timestamp.
-	// If snapshot has deletion timestamp and finalizers, set
-	// AnnVolumeSnapshotBeingDeleted annotation on the content.
-	// This may trigger the deletion of the content in the
-	// sidecar controller depending on the deletion policy
-	// on the content.
-	// Snapshot won't be deleted until content is deleted
-	// due to the finalizer.
-	if snapshot != nil && utils.IsSnapshotDeletionCandidate(snapshot) {
-		// Do not need to use the returned content here, as syncContent will get
-		// the correct version from the cache next time. It is also not used after this.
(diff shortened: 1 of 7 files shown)


huww98:
/cc @xing-yang
as I see your NOTE in comment. And you also appear in `git blame`

pkg/common-controller/snapshot_update_test.go
```diff
@@ -318,36 +318,37 @@ func TestSync(t *testing.T) {
...
 			initialSecrets:   []*v1.Secret{secret()},
 			errors: []reactorError{
-				// Inject error to the forth client.VolumesnapshotV1().VolumeSnapshots().Update call.
-				{"update", "volumesnapshotcontents", errors.New("mock update error")},
+				// Inject error to the forth client.VolumesnapshotV1().VolumeSnapshotContents().Patch call.
+				{"patch", "volumesnapshotcontents", errors.New("mock update error")},
 			},
-			expectSuccess: false,
-			test:          testSyncContentError,
+			expectSuccess: true,
```
huww98, line 347: I think we should always do `expectSuccess: true`? Specify false means don't check for failure, not asserting error is returned. But that is a big project. Let's leave it to another PR.

pkg/common-controller/snapshot_controller.go
```diff
@@ -149,31 +149,6 @@ func (ctrl *csiSnapshotCommonController) syncContent(content *crdv1.VolumeSnapsh
...
-	//
-	// Trigger content deletion if snapshot is not nil
-	// and snapshot has deletion timestamp.
-	// If snapshot has deletion timestamp and finalizers, set
-	// AnnVolumeSnapshotBeingDeleted annotation on the content.
-	// This may trigger the deletion of the content in the
-	// sidecar controller depending on the deletion policy
-	// on the content.
-	// Snapshot won't be deleted until content is deleted
-	// due to the finalizer.
```
xing-yang, line 169: What is the reason to remove this whole block?

huww98: https://github.com/kubernetes-csi/external-snapshotter/blob/3a35a5e48640085658d07b61eea3e0bdc3a757c9/pkg/common-controller/snapshot_controller.go#L332-L337
This code should already handle it.

In the commit message of second commit, I wrote:
> The original annotate logic in syncContent() is not needed because all deleting snapshot will go through above logic. And natually, if a content is never bound, is will never get the vs-deleted annotation.

huww98: Hi @xing-yang What do you think about the overall direction of this PR? Should we continue to work on this? These change is not required for this PR. If you agree, I can defer it to another PR.

PR #1094, opened 2024-05-22
Use a marker for the Namespace VSC additional printer column

leonardoce:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
The Namespace additional printer column is now generated from a marker in the codebase instead of directly added in the CRD.
```

### DIFF
--- a/pkg/utils/patch.go
+++ b/pkg/utils/patch.go
@@ -3,8 +3,8 @@ package utils
 import (
 	"context"
 	"encoding/json"
-	crdv1alpha1 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1alpha1"
 
+	crdv1alpha1 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1alpha1"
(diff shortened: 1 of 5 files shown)


pkg/utils/patch.go
```diff
@@ -3,8 +3,8 @@ package utils
 import (
 	"context"
 	"encoding/json"
-	crdv1alpha1 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1alpha1"
 
+	crdv1alpha1 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1alpha1"
```
xing-yang, line 7: This change is not necessary?  Can you remove it?

xing-yang:
Can you add a release note?

pkg/utils/patch.go
leonardoce: It's just a cleanup, grouping together all the internal imports. Do you want me to open a PR just for that, or we leave it as it is?

xing-yang: Oh, I see what you are doing.  That's fine then.

PR #1100, opened 2024-05-28
Correctly detect newly created Volume[Group]SnapshotClasses

### LINKED ISSUE #1099: Controller does not detect new Volume[Group]SnapshotClass set as default
**What happened**:

If multiple default `Volume[Group]SnapshotClasses` are created before the external-snapshotter controller Pod starts, the controller will detect them correctly and when webhook is not installed dynamically provisioned VolumeSnapshot and VolumeGroupSnapshots will fail with:

```
apiVersion: groupsnapshot.storage.k8s.io/v1alpha1
kind: VolumeGroupSnapshot
metadata:
  name: new-groupsnapshot-demo
spec:
  source:
    selector:
      matchLabels:
        cnpg.io/instanceName: cluster-example-1
status:
  error:
    message: Failed to set default group snapshot class with error 2 default snapshot
      classes were found
    time: "2024-05-28T08:24:42Z"
```

This doesn't happen if the external-snapshotter controller starts with a correct configuration (having just one default class per CSI driver) and a new one is added later.
In that case, the Volume[Group]Snapshot will be provisioned by the CSI driver specified in the Volume[Group]SnapshotClass that was present before the controller started.

I expect the behavior to be the same, regardless of when the Volume[Group]SnapshotClass was created.

**How to reproduce it**:

On an empty cluster with just the hostpath provisioner installed, create one default VolumeSnapshotClass.
Install the external-snapshotter controller and, when it's ready, create a new VolumeSnapshotClass.

You'll see the VolumeSnapshot objects being created from the first VolumeSnapshotClass you created and set as default.

leonardoce:
**What type of PR is this?**

/kind bug

>
> /kind api-change
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR makes the controller process newly created default VolumeSnapshotClasses and VolumeGroupSnapshotClasses, correctly using them or raising an error when appropriate.

Fixes #1099 

The controller is using an informer to get VolumeGroupSnapshotClasses and VolumeSnapshotClasses.
Unfortunately, objects returned by that will not have the API Kind info set (see https://github.com/kubernetes/client-go/issues/308).

This prevented the `IsDefaultAnnotation` function from working correctly.

This patch avoids relying on the object Kind to know whether a VolumeGroupSnapshotClass or a VolumeSnapshotClass is set as default.

The semantics of the `groupsnapshot.storage.kubernetes.io/is-default-class` have been preserved, with `true` being true and everything else being `false` (this is slightly different from Kubernetes).

```release-note
The controller will now detect new default VolumeSnapshotClasses and VolumeGroupSnapshotClasses.
If multiple classes exist for the same CSI driver, VolumeSnapshot and VolumeGroupSnapshots
will be marked as failed when provisioned dynamically.
```

### DIFF
--- a/pkg/common-controller/groupsnapshot_controller_helper.go
+++ b/pkg/common-controller/groupsnapshot_controller_helper.go
@@ -141,7 +141,7 @@ func (ctrl *csiSnapshotCommonController) SetDefaultGroupSnapshotClass(groupSnaps
 
 	defaultClasses := []*crdv1alpha1.VolumeGroupSnapshotClass{}
 	for _, groupSnapshotClass := range list {
-		if utils.IsDefaultAnnotation(groupSnapshotClass.TypeMeta, groupSnapshotClass.ObjectMeta) && pvDriver == groupSnapshotClass.Driver {
+		if utils.IsVolumeGroupSnapshotClassDefaultAnnotation(groupSnapshotClass.ObjectMeta) && pvDriver == groupSnapshotClass.Driver {
 			defaultClasses = append(defaultClasses, groupSnapshotClass)
 			klog.V(5).Infof("get defaultGroupClass added: %s, driver: %s", groupSnapshotClass.Name, pvDriver)
 		}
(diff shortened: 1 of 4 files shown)


xing-yang:
Please rebase.

leonardoce:
Done @xing-yang

PR #1133, opened 2024-08-07
[snapshot-controller] Retry PVC finalizer removal on conflict

Fricounet:
**What type of PR is this?**

> /kind bug

The previous `removePVCFinalizer` function was using the PVC stored in the informer which, in cases where the PVC had been modified since, lead to conflict errors when trying to remove the PVC finalizer through an update.
Example error:
```
E0807 13:09:20.573657       1 snapshot_controller.go:183] error check and remove PVC finalizer for snapshot [snapshot]: snapshot controller failed to update pvc-sts-0 on API server: Operation cannot be fulfilled on persistentvolumeclaims "pvc-sts-0": the object has been modified; please apply your changes to the latest version and try again
```
Unfortunately, the error was not retried so the PVC ends up stuck.

Now the `removePVCFinalizer` function uses the `RetryOnConflict` helper to make sure the update goes through.

Fixes #

```release-note
Fixes a race condition where the PVC finalizer could end up not being removed by the snapshot-controller if the update had a conflict.
```

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes/scheme"
 	ref "k8s.io/client-go/tools/reference"
+	"k8s.io/client-go/util/retry"
 	corev1helpers "k8s.io/component-helpers/scheduling/corev1"
 	klog "k8s.io/klog/v2"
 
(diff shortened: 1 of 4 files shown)


pkg/common-controller/snapshot_controller.go
```diff
@@ -930,19 +931,22 @@ func (ctrl *csiSnapshotCommonController) ensurePVCFinalizer(snapshot *crdv1.Volu
...
-	}
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		// Get snapshot source which is a PVC
+		newPvc, err := ctrl.client.CoreV1().PersistentVolumeClaims(pvc.Namespace).Get(context.TODO(), pvc.Name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		newPvc = newPvc.DeepCopy()
+		newPvc.ObjectMeta.Finalizers = utils.RemoveString(newPvc.ObjectMeta.Finalizers, utils.PVCFinalizer)
+		_, err = ctrl.client.CoreV1().PersistentVolumeClaims(newPvc.Namespace).Update(context.TODO(), newPvc, metav1.UpdateOptions{})
```
xing-yang, line 942: Using a "patch" instead of an "update" can reduce the conflicts.

Fricounet: I do agree that a using a "patch" would get rid of all the conflicts however since the `finalizers` are stored in a list, using a patch seems dangerous to me because if the finalizer list changes between the get and update calls (because another controller removed its finalizer for instance), it means that we would end up removing the wrong finalizer (since we need the index to remove an element from a list using a patch).

Is there something I'm missing that could let us use a patch instead without the drawback above?

pkg/common-controller/snapshot_controller.go
```diff
@@ -930,19 +931,22 @@ func (ctrl *csiSnapshotCommonController) ensurePVCFinalizer(snapshot *crdv1.Volu
...
+		if err != nil {
+			return newControllerUpdateError(newPvc.Name, err.Error())
```
akalenyu, line 944: folks, I suspect generating a new error here (newControllerUpdateError) results in IsConflict never evaluating to true, and thus the original problem still existing. I hit this on an environment with 8.2.0 which contains this commit.

I can open a PR to tackle

akalenyu: https://github.com/kubernetes-csi/external-snapshotter/pull/1299

Fricounet: Hi @akalenyu sorry about this! I actually did notice it a while back and made a [PR](https://github.com/DataDog/kubernetes-csi-external-snapshotter/pull/8) in our fork to make sure this fixed the behavior for good (which it did) but I forgot to upstream it :(

PR #139, opened 2019-07-11
Update VolumeSnapshot CRD version to v1beta

xing-yang:
**What type of PR is this?**
/kind api-change

**What this PR does / why we need it**:
This PR changes VolumeSnapshot CRD version from v1alpha1 to v1beta1.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
Changes VolumeSnapshot CRD version from v1alpha1 to v1beta1.
```

PR #1150, opened 2024-08-30
Update VolumeGroupSnapshot to v1beta1

### LINKED ISSUE #1119: Update VolumeGroupSnapshot CRDs to v1beta1

We are moving VolumeGroupSnapshot to Beta. The CRDs need to be updated accordingly.

leonardoce:
**What type of PR is this?**
>
/kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Update the definition of CRDs to move VolumeGroupSnapshot, VolumeGroupSnapshotContent, and VolumeGroupSnapshotClass from v1alpha1 to v1beta1.

cc: @xing-yang 

Fixes #1119

This PR entirely removes the capability to serve the `v1alpha` version. Existing clients will be broken.
This is consistent with the [relative PR comment](https://github.com/kubernetes-csi/external-snapshotter/pull/1150#discussion_r1852973872) and the related [Kubernetes documentation](https://github.com/kubernetes-csi/external-snapshotter/pull/1150#discussion_r1853037522).

The underlying structure of `v1alpha` and `v1beta1` is consistent.

```release-note
`VolumeGroupSnapshot`, `VolumeGroupSnapshotContent`, and `VolumeGroupSnapshotClass`
are now available in `v1beta1` version. The support for the `v1alpha1` version have been removed.
```


leonardoce:
I rebased this PR and added a commit to add and set the new field `VolumeSnapshotHandlePairList ` of the `VolumeGroupSnapshotContent` resource, as described in the [latest iteration of the relative KEP](https://github.com/kubernetes/enhancements/tree/e1d4725d07a07c40da7ded6e1dd44c0f9aaeaeb9/keps/sig-storage/3476-volume-group-snapshot#volumegroupsnapshotcontent).

Thank you!

cc: @xing-yang

client/apis/volumegroupsnapshot/v1beta1/types.go
```diff
@@ -341,24 +347,24 @@ type VolumeGroupSnapshotContentStatus struct {
 	// On Unix, the command date +%s%N returns the current time in nanoseconds
 	// since 1970-01-01 00:00:00 UTC.
 	// +optional
-	CreationTime *int64 `json:"creationTime,omitempty" protobuf:"varint,2,opt,name=creationTime"`
+	CreationTime *int64 `json:"creationTime,omitempty" protobuf:"varint,3,opt,name=creationTime"`
```
nixpanic, line 350: This is a breaking change in the gRPC protocol, has that been considered? Existing indexes should ideally not be modified to keep it backwards compatible.

leonardoce: You are right. I can definitely use a new field index for the new field, and keep the ones we already use steady.

leonardoce: Thank you for spotting this. I updated the commit and fixed it.

yati1998 (commented):
LGTM

xing-yang:
@leonardoce We discussed this PR during today's CSI Implementation meeting. @jsafrane suggested that we break this PR into two PRs.
- PR 1: Add new field and populate the new field but keeping the APIs at alpha.
- PR 2: Move the APIs to Beta.
There are other things we need to do between PR 1 and PR2. See detailed plan here: https://docs.google.com/document/d/1jjliw0tFhCPnT9ixlYpj61k5u8_-awf1qSCWrkunshQ/edit?tab=t.0

leonardoce:
Ok. I'll close it and file the one to add the new field

yati1998:
@leonardoce can I know why this PR is in draft? I think we were good on this part?

leonardoce:
Hi @yati1998.
According to this plan: https://docs.google.com/document/d/1jjliw0tFhCPnT9ixlYpj61k5u8_-awf1qSCWrkunshQ/edit?tab=t.0#heading=h.fizenae5evi0
We need at least https://github.com/kubernetes-csi/external-snapshotter/pull/1177 to be merged before this one.

gnufied:
@leonardoce can you update this PR to resolve the conflicts?

xing-yang:
Can you please rebase?

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
```diff
@@ -92,3 +92,77 @@ spec:
...
+    - description: Determines whether a VolumeGroupSnapshotContent created through
+        the VolumeGroupSnapshotClass should be deleted when its bound VolumeGroupSnapshot
+        is deleted.
+      jsonPath: .deletionPolicy
+      name: DeletionPolicy
+      type: string
+    - jsonPath: .metadata.creationTimestamp
+      name: Age
+      type: date
+    name: v1alpha1
```
xing-yang, line 108: I don't think we need to keep the v1alpha1 APIs.  We didn't keep the v1alpha1 VolumeSnapshot APIs when we moved it to beta: https://github.com/kubernetes-csi/external-snapshotter/pull/139

yati1998:
@gnufied can you please review this as well?

examples/kubernetes/groupsnapshot-v1beta1.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-apiVersion: groupsnapshot.storage.k8s.io/v1alpha1
+apiVersion: groupsnapshot.storage.k8s.io/v1beta1
```
xing-yang, line 2: Can you fix the label?
`groupsnapshot.storage.k8s.io/volumeGroupSnapshotName: postgresql`

leonardoce: Hi! If you are referring to `app.kubernetes.io/name: postgresql`, that's just the selector this example uses to get the PVC list. I don't think we need to change it 👍

xing-yang:
@leonardoce Please ignore my comments:)

gnufied:
We should compile this branch rebased with master and then run e2e tests in a loop for sometime. I have often done this for testing pre-release code which needs more testing. This will give us greater confidence about this feature. This does not require making a release.

client/apis/volumegroupsnapshot/v1beta1/types.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 // +kubebuilder:object:generate=true
-package v1alpha1
```
gnufied, line 17: Why is `VolumeGroupSnapshot.Status.CreationTime` --> `metav1.Time` and why is `VolumeGroupSnapshotContentStatus.CreationTime` --> int64 ?

I know that change was not introduced in this PR, but since we are moving this to beta, we have to be careful about all the decisions we make.

leonardoce: The same is happening with independent VolumeSnapshots:

https://github.com/kubernetes-csi/external-snapshotter/blob/07425bd341a786aa58eecb1d2dcec81b5e6dcfe8/client/apis/volumesnapshot/v1/types.go#L158

vs:

https://github.com/kubernetes-csi/external-snapshotter/blob/07425bd341a786aa58eecb1d2dcec81b5e6dcfe8/client/apis/volumesnapshot/v1/types.go#L408 

The content object looks more directly related to the GRPC CSI call.

gnufied: This is how it is displayed in CLI then:

```
  creationTime: "2024-11-22T22:14:35Z"
  creationTime: 1732313675689050917
```

@xing-yang can you confirm if this intentional? This doesn't seem right.

gnufied: We can always convert `int64` to `metav1.Time` so as this is nicely handled by k8s tooling. I am not sure if this has to be int64...

xing-yang: In CSI spec, the timestamp is .google.protobuf.Timestamp which represents seconds of UTC time since Unix epoch 1970-01-01T00:00:00Z. So in VolumeSnapshotContent and VolumeGroupSnapshotContent, it is a direct mapping to what is returned from the CSI driver.
VolumeSnapshot and VolumeGroupSnapshot are in user namespace, we try to show them nicely in the readable format of `metav1.Time`.

client/apis/volumegroupsnapshot/v1beta1/types.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
...
+package v1beta1
```
gnufied, line 18: Do we need to drop `PVCVolumeSnapshotPair` from here? We are not using it by the looks of it.

client/apis/volumegroupsnapshot/v1beta1/types.go
gnufied: I still do not understand why though. `metav1.Time` also represents timestamp equally well. In k8s APIs, `metav1.Time` is almost universally used for representing time. Is there a loss of information or something we will incur if we use `metav1.Time`?

client/apis/volumegroupsnapshot/v1beta1/types.go
leonardoce: In my opinion, yes, and `PVVolumeSnapshotContentPair` is a leftover too.

gnufied:
I noticed that following field in `volumeGroupSnapshot.Status` is user editable after setting it:

```
BoundVolumeGroupSnapshotContentName *string `json:"boundVolumeGroupSnapshotContentName,omitempty" protobuf:"bytes,1,opt,name=boundVolumeGroupSnapshotContentName"`
```

Should this be immutable after creation?

client/apis/volumegroupsnapshot/v1beta1/types.go
xing-yang: VolumeSnapshotContent and VolumeGroupSnapshotContent are admin API objects. We don't need to show them in user readable format.  I think they should stay true to the original values returned by the CSI driver.
Since we've already made a decision for this in VolumeSnapshotContent, it also looks inconsistent now if we are showing them differently in VolumeGroupSnapshotContent.

xing-yang:
> I noticed that following field in volumeGroupSnapshot.Status is user editable after setting it:

> BoundVolumeGroupSnapshotContentName *string `json:"boundVolumeGroupSnapshotContentName,omitempty"  protobuf:"bytes,1,opt,name=boundVolumeGroupSnapshotContentName"`
> Should this be immutable after creation?

This field is in the Status. Are we supposed to make fields in Status immutable?  We have not done that for VolumeSnapshot APIs.

gnufied:
>This field is in the Status. Are we supposed to make fields in Status immutable? We have not done that for VolumeSnapshot APIs.

It would have been nice to have. But not a blocker I think.

client/apis/volumegroupsnapshot/v1beta1/types.go
```diff
@@ -354,16 +345,6 @@ type VolumeGroupSnapshotContentStatus struct {
...
-// VolumeSnapshotContent names
-type PVVolumeSnapshotContentPair struct {
-	// PersistentVolumeRef is a reference to the persistent volume resource
-	PersistentVolumeRef core_v1.LocalObjectReference `json:"persistentVolumeRef,omitempty" protobuf:"bytes,1,opt,name=persistentVolumeRef"`
-
-	// VolumeSnapshotContentRef is a reference to the volume snapshot content resource
-	VolumeSnapshotContentRef core_v1.LocalObjectReference `json:"volumeSnapshotContentRef,omitempty" protobuf:"bytes,2,opt,name=volumeSnapshotContentRef"`
-}
-
 // VolumeGroupSnapshotContentSource represents the CSI source of a group snapshot.
```
gnufied, line 348: Not visible in PR - Can `VolumeHandle` and `SnapshotHandle` inside `VolumeSnapshotHandlePair` be empty? Or they both have to be specified. I didn't see these fields marked as `required` and neither see any validation kinda logic preventing saving of these fields without any values. Correct me if I missed something.

xing-yang: VolumeSnapshotHandlePair itself is not required, but if a VolumeSnapshotHandlePair is specified, then both VolumeHandle and SnapshotHandle should be required. Otherwise, it is not very useful.
We can add CEL for them.

client/apis/volumegroupsnapshot/v1beta1/types.go
xing-yang: Since VolumeSnapshot and VolumeGroupSnapshot already have timestamps shown in user readable format, my preference is to keep the timestamps in VolumeSnapshotContent and VolumeGroupSnapshotContent the same as what is returned by the CSI driver as they are admin facing, but let's see what @jsafrane and @msau42 think.

client/apis/volumegroupsnapshot/v1beta1/types.go
gnufied: > We can add CEL for them.
Yeah we should IMO.

Madhu-1:
@gnufied @xing @leonardoce i have tested this PR + #1219 https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.31/detail/mini-e2e_k8s-1.31/201/pipeline/ https://github.com/ceph/ceph-csi/pull/4974 (we already have framework for volumegroupsnapshot)

This is what it does

* Create 3 PVC
* Create a groupsnapshot
* Create a clones from the 3 snapshots (as part of group)
* Create pod using the clones
* Delete the pod
* Delete the clones
* Delete the group
* Delete the PVC

Repeat the above in a loop of 20 , let me know if you guys wants me to test anything else?

client/apis/volumegroupsnapshot/v1beta1/types.go
xing-yang: @leonardoce Can you make changes to make VolumeHandle and SnapshotHandle required in VolumeSnapshotHandlePair?

leonardoce: Sure! I think it's better to file a different PR for that and keep this as a giant substitution. If we use this PR to review the API, it will become really difficult to rebase

xing-yang: Discussed with @gnufied.  He prefers this change to be made as part of this PR.

client/apis/volumegroupsnapshot/v1beta1/types.go
xing-yang: @leonardoce can you please update the comments with the highlighted text?

type VolumeGroupSnapshotStatus struct {
..........
	// CreationTime is the timestamp when the point-in-time group snapshot is taken
	// by the underlying storage system.
	// If not specified, it may indicate that the creation time of the group snapshot
	// is unknown.
	// The format of this field is a Unix nanoseconds time encoded as an int64.
	// On Unix, the command date +%s%N returns the current time in nanoseconds
	// since 1970-01-01 00:00:00 UTC.
        **// This field is updated based on the CreationTime field in VolumeGroupSnapshotContentStatus**
	// +optional
	CreationTime *metav1.Time `json:"creationTime,omitempty" protobuf:"bytes,2,opt,name=creationTime"`

type VolumeGroupSnapshotContentStatus struct {
..........
	// CreationTime is the timestamp when the point-in-time group snapshot is taken
	// by the underlying storage system.
	// If not specified, it indicates the creation time is unknown.
	// If not specified, it means the readiness of a group snapshot is unknown.
	// The format of this field is a Unix nanoseconds time encoded as an int64.
	// On Unix, the command date +%s%N returns the current time in nanoseconds
	// since 1970-01-01 00:00:00 UTC.
        **// This field is the source for the CreationTime field in VolumeGroupSnapshotStatus**
	// +optional
	CreationTime *int64 `json:"creationTime,omitempty" protobuf:"varint,2,opt,name=creationTime"`

client/apis/volumegroupsnapshot/v1beta1/types.go
gnufied: The main reason I wanted all API changes in same PR is, so as we don't have to do follow up API fixes.

leonardoce: @gnufied I see both fields are already marked as required (meaning they are not optional) in the CRD.
See:

https://github.com/kubernetes-csi/external-snapshotter/blob/b79aef97261427cf2765410dc7f9d6647b7b3928/client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotcontents.yaml#L290-L312

Do you still need me to add some CEL rules for them? Thanks!

client/apis/volumegroupsnapshot/v1beta1/types.go
leonardoce: I improved the comments @xing-yang.

client/apis/volumegroupsnapshot/v1beta1/types.go
xing-yang: That should be sufficient.  The API server will reject it if VolumeHandle or SnapshotHandle is not set in VolumeSnapshotHandlePair.

xing-yang: Can you add "// Required." as comments for both VolumeHandle and SnapshotHandle in types.go?

leonardoce: I did it and rebased the PR too.

iPraveenParihar:
Hey, I tested this using the CSI HostPath Driver in a GitHub Action. The test, which can be viewed [[here](https://github.com/iPraveenParihar/csi-driver-host-path/actions/runs/12114803563/job/33771961716)](https://github.com/iPraveenParihar/csi-driver-host-path/actions/runs/12114803563/job/33771961716), includes the following steps:

1. Create 3 PVCs and mount them to a Pod.
2. Create a VolumeGroupSnapshot.
3. Restore PVCs from VolumeSnapshots, mount them to a Pod.
4. Delete the VolumeGroupSnapshot.
5. Delete the Pods and PVCs.

This sequence is repeated 20 times.

cc @Madhu-1

Madhu-1:
@gnufied The above is the results for e2e tests

leonardoce:
I rebased it over the new master and tested it. My smoke tests look good.

gbartolini:
This is great news!!! Amazing work, well done! Congratulations!

client/apis/volumegroupsnapshot/v1beta1/types.go
msau42: I believe we concluded last week to use `metav1.Time` for CreationTime?

xing-yang: Opened an issue: https://github.com/kubernetes-csi/external-snapshotter/issues/1234
cc @leonardoce

xing-yang: @msau42 PR is submitted here: https://github.com/kubernetes-csi/external-snapshotter/pull/1235

msau42:
API review lgtm

jonathon2nd:
I was wondering how I can install the update when I am getting the following error
`CustomResourceDefinition.apiextensions.k8s.io "volumegroupsnapshots.groupsnapshot.storage.k8s.io" is invalid: status.storedVersions[0]: Invalid value: "v1alpha1": must appear in spec.versions`

See https://github.com/piraeusdatastore/helm-charts/issues/54

pl4nty:
@jonathon2nd `kubectl patch crd x.groupsnapshot.storage.k8s.io --subresource='status' --type='merge' -p '{"status":{"storedVersions": ["v1beta1"]}}'` where x is each of `volumegroupsnapshots`, `volumegroupsnapshotclasses`, and `volumegroupsnapshotcontents`

Issue #1035, opened 2024-03-14
VolumeGroupSnapshot deletion intermediate failures


Madhu-1:
**What happened**:

The volumegroupsnapshot deletion is kind of stuck because the volumesnapshotcontent are already deleted https://github.com/kubernetes-csi/external-snapshotter/blob/fcf78d3d6964632ed7f8b85aa045d667b1da47d4/pkg/sidecar-controller/groupsnapshot_helper.go#L242-L249

**What you expected to happen**:

The volumegroupsnapshot deletion should happen

**How to reproduce it**:

It's sometimes happens not always

* Create VolumeGrousnapshot
* Delete VolumeGroupSnapshot
* 
**Anything else we need to know?**:

**Environment**:
- Driver version:
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

# Logs

```
I0314 11:31:02.878590       1 connection.go:244] GRPC call: /csi.v1.Controller/ControllerGetCapabilities
I0314 11:31:02.878604       1 connection.go:245] GRPC request: {}
I0314 11:31:02.881202       1 connection.go:251] GRPC response: {"capabilities":[{"Type":{"Rpc":{"type":1}}},{"Type":{"Rpc":{"type":5}}},{"Type":{"Rpc":{"type":9}}},{"Type":{"Rpc":{"type":7}}},{"Type":{"Rpc":{"type":13}}}]}
I0314 11:31:02.881426       1 connection.go:252] GRPC error: <nil>
I0314 11:31:02.881516       1 snapshot_controller.go:291] checkandUpdateContentStatusOperation: driver rook-ceph.cephfs.csi.ceph.com, snapshotId 0001-0009-rook-ceph-0000000000000001-5c98229b-0f9d-4448-9ccf-6822d3b87e81, creationTime 0001-01-01 00:00:00 +0000 UTC, size 0, readyToUse true, groupSnapshotID 
[...]

PR #1164, opened 2024-09-26
add check continue VGS deletion

yati1998:
this commit adds check to continue deleting the
volumegroupsnapshot in case there are no snapshotcontent found under that group.

https://github.com/kubernetes-csi/external-snapshotter/issues/1035

### DIFF
--- a/pkg/sidecar-controller/groupsnapshot_helper.go
+++ b/pkg/sidecar-controller/groupsnapshot_helper.go
@@ -251,6 +251,10 @@ func (ctrl *csiSnapshotSideCarController) deleteCSIGroupSnapshotOperation(groupS
 		for _, contentRef := range groupSnapshotContent.Status.PVVolumeSnapshotContentList {
 			snapshotContent, err := ctrl.contentLister.Get(contentRef.VolumeSnapshotContentRef.Name)
 			if err != nil {
+				if errors.IsNotFound(err) {
+					klog.Infof("snapshot content %v not found", contentRef.VolumeSnapshotContentRef.Name)
+					continue
+				}
 				return fmt.Errorf("failed to get snapshot content %s from snapshot content store: %v", contentRef.VolumeSnapshotContentRef.Name, err)
 			}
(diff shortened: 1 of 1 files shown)


yati1998:
@Madhu-1 @xing-yang please take a look.
I tested the changes and now the volumegroupdeletion is not getting stuck even after multiple deletions

xing-yang:
@yati1998 Is there an issue opened for this bug fix?  If there is one, please add to your PR description.

pkg/sidecar-controller/groupsnapshot_helper.go
```diff
@@ -251,6 +251,10 @@ func (ctrl *csiSnapshotSideCarController) deleteCSIGroupSnapshotOperation(groupS
 		for _, contentRef := range groupSnapshotContent.Status.PVVolumeSnapshotContentList {
 			snapshotContent, err := ctrl.contentLister.Get(contentRef.VolumeSnapshotContentRef.Name)
 			if err != nil {
+				if errors.IsNotFound(err) {
+					klog.Infof("snapshot content %v not found", contentRef.VolumeSnapshotContentRef.Name)
```
xing-yang, line 255: This should be a warning. If snapshot content not found, that means it is somehow forcefully deleted. A snapshot content can't be deleted if group snapshot handle is still in its status.
Can you tell me how you reproduced this problem?

yati1998: yes even I thought the same, but this issue was reproducible when we repeatedly create and delete a VGS. As per my debug and testing, since the creation and deletion was taking place with same name within seconds, it seems there was a sync issue due to which the snapshot content got deleted . There is a script that we used to reproduce this issue, let me know if you want I can share it with you.

yati1998: Hey, I think we dont need this check anymore, it seems the VGS is getting successfully deleted.

PR #1169, opened 2024-10-23
Add and set the new VolumeSnapshotHandlePairList field

leonardoce:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
/kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:

Adds and sets the new field VolumeSnapshotHandlePairList of the VolumeGroupSnapshotContent resource, as described in the [latest iteration of the relative KEP](https://github.com/kubernetes/enhancements/tree/e1d4725d07a07c40da7ded6e1dd44c0f9aaeaeb9/keps/sig-storage/3476-volume-group-snapshot#volumegroupsnapshotcontent).

This is a prerequisite to promote the volume group snapshot API to v1beta1.

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

This has been extracted from https://github.com/kubernetes-csi/external-snapshotter/pull/1150

This is an example of how a VolumeGroupSnapshotContent looks:

```
❯ kubectl get volumegroupsnapshotcontent groupsnapcontent-cea7182b-4b04-4614-a3e8-a5b09c669e5c  -o yaml
apiVersion: groupsnapshot.storage.k8s.io/v1alpha1
kind: VolumeGroupSnapshotContent
metadata:
  creationTimestamp: "2024-10-23T15:24:11Z"
  finalizers:
  - groupsnapshot.storage.kubernetes.io/volumegroupsnapshotcontent-bound-protection
  generation: 1
  name: groupsnapcontent-cea7182b-4b04-4614-a3e8-a5b09c669e5c
  resourceVersion: "458156"
  uid: 740d4c07-9efc-4c6b-b46a-3c4733a391d3
spec:
  deletionPolicy: Delete
  driver: hostpath.csi.k8s.io
  source:
    volumeHandles:
    - 75b1b2e9-8cc4-11ef-b606-c20bfb388315
    - 75b196d5-8cc4-11ef-b606-c20bfb388315
  volumeGroupSnapshotClassName: csi-hostpath-groupsnapclass
  volumeGroupSnapshotRef:
    apiVersion: groupsnapshot.storage.k8s.io/v1alpha1
    kind: VolumeGroupSnapshot
    name: new-groupsnapshot-demo
    namespace: default
    resourceVersion: "458124"
[...]

PR #1171, opened 2024-10-24
Centralize VolumeGroupSnapshot members resource creation in the snapshot-controller

leonardoce:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Previously, the csi-snapshotter was responsible for creating individual
VolumeSnapshots and VolumeSnapshotContents for each member when
dynamically provisioning a VolumeGroupSnapshot.

This commit relocates this logic to the snapshot-controller, bringing
all resource creation processes under a single authority.

Fixes #

```release-note
Move the logic of creating individual VolumeSnapshot and VolumeSnapshotContent resources for dynamically created VolumeGroupSnapshot from csi-snapshotter sidecar to snapshot-controller.
```


xing-yang:
Hi @leonardoce, can you please split this into 2 separate PRs, one depending on the other?  Thanks!

pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -485,6 +493,191 @@ func (ctrl *csiSnapshotCommonController) syncUnreadyGroupSnapshot(groupSnapshot
...
+					Name:      volumeSnapshotName,
+					Namespace: volumeSnapshotNamespace,
+				},
+				DeletionPolicy: groupSnapshotContent.Spec.DeletionPolicy,
+				Driver:         groupSnapshotContent.Spec.Driver,
+				Source: crdv1.VolumeSnapshotContentSource{
+					SnapshotHandle: &snapshotHandle,
+				},
+				// TODO: Populate this field when volume mode conversion is enabled by default
+				SourceVolumeMode: nil,
```
xing-yang, line 576: We should set SourceVolumeMode now as volume mode conversion is already GA.  I know that this is like this before. We can make this change in a followup PR.
See example here: https://github.com/kubernetes-csi/external-snapshotter/blob/7e7fea24e7e22b3b321025477ab709adbafe6f90/pkg/common-controller/snapshot_controller.go#L698
We are retrieving PV a little later in this function.  We can probably set it there.

xing-yang:
Can you add a release note?

pkg/common-controller/groupsnapshot_controller_helper.go
leonardoce: I also did that. Should it fit in a different PR, just tell me I'll split this.

deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
```diff
@@ -38,7 +38,7 @@ rules:
     verbs: ["patch"]
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshots"]
-    verbs: ["get", "list", "watch", "update", "patch", "delete"]
+    verbs: ["create", "get", "list", "watch", "update", "patch", "delete"]
```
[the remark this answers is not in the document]
xing-yang: Also remove "create" from "volumesnapshotcontents" rbac rules.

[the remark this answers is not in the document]
leonardoce: Done @xing-yang. I tested everything again and it works fine.
Thank you!

leonardoce:
Since #1169 is merged, I rebased over the latest master and marked it as ready for review.

xing-yang:
Can you reword the release notes as follows? 
Move the logic of creating individual VolumeSnapshot and VolumeSnapshotContent resources for dynamically created VolumeGroupSnapshot from csi-snapshotter sidecar to snapshot-controller.

leonardoce:
I did it @xing-yang. Thank you!

pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -485,6 +493,204 @@ func (ctrl *csiSnapshotCommonController) syncUnreadyGroupSnapshot(groupSnapshot
...
+			klog.Errorf(
+				"updateGroupSnapshotContentStatus: unable to find PV for volumeHandle:[%s] and CSI driver:[%s]",
+				volumeHandle,
+				groupSnapshotContent.Spec.Driver)
+		}
+
+		volumeSnapshotContentName := CreateSnapshotContentNameForVolumeGroupSnapshotContent(
+			string(groupSnapshotContent.UID), volumeHandle)
+
+		volumeSnapshotName := CreateSnapshotNameForVolumeGroupSnapshotContent(string(groupSnapshotContent.UID), volumeHandle)
```
jsafrane, line 568: `CreateSnapshotNameForVolumeGroupSnapshotContent` takes PV UUID as the second parameter, but you pass volume handle. It will still be unique, but please keep the parameter names + the actual values in sync.

pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -485,6 +493,204 @@ func (ctrl *csiSnapshotCommonController) syncUnreadyGroupSnapshot(groupSnapshot
...
+
+		pv := utils.GetPersistentVolumeFromHandle(pvs, groupSnapshotContent.Spec.Driver, volumeHandle)
+		if pv == nil {
+			klog.Errorf(
+				"updateGroupSnapshotContentStatus: unable to find PV for volumeHandle:[%s] and CSI driver:[%s]",
+				volumeHandle,
+				groupSnapshotContent.Spec.Driver)
+		}
+
+		volumeSnapshotContentName := CreateSnapshotContentNameForVolumeGroupSnapshotContent(
```
jsafrane, line 565: CreateSnapshotContentNameForVolumeGroupSnapshotContent takes PV UUID as the second parameter, but you pass volume handle. It will still be unique, but please keep the parameter names + the actual values in sync.

pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -485,6 +493,204 @@ func (ctrl *csiSnapshotCommonController) syncUnreadyGroupSnapshot(groupSnapshot
...
+			persistentVolume:    pv,
+		}
+
+		volumeSnapshotContent := &crdv1.VolumeSnapshotContent{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: volumeSnapshotContentName,
+			},
+			Spec: crdv1.VolumeSnapshotContentSpec{
+				VolumeSnapshotRef: v1.ObjectReference{
+					Kind:      "VolumeSnapshots",
```
jsafrane, line 582: `Kind` is `VolumeSnapshot`

pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -485,6 +493,204 @@ func (ctrl *csiSnapshotCommonController) syncUnreadyGroupSnapshot(groupSnapshot
...
+				},
+				DeletionPolicy: groupSnapshotContent.Spec.DeletionPolicy,
+				Driver:         groupSnapshotContent.Spec.Driver,
+				Source: crdv1.VolumeSnapshotContentSource{
+					SnapshotHandle: &snapshotHandle,
+				},
+			},
+			Status: &crdv1.VolumeSnapshotContentStatus{
+				VolumeGroupSnapshotHandle: &groupSnapshotContent.Name,
+			},
```
jsafrane, line 594: Who files the rest of status, like `RestoreSize`, `CreationTime` etc? If it's a subsequent controller sync, then please add a comment about it.

pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -485,6 +493,204 @@ func (ctrl *csiSnapshotCommonController) syncUnreadyGroupSnapshot(groupSnapshot
...
+		"createSnapshotsForGroupSnapshotContent[%s]: creating volumesnapshots and volumesnapshotcontent for group snapshot content",
+		groupSnapshotContent.Name)
+
+	for _, snapshot := range groupSnapshotContent.Status.VolumeSnapshotHandlePairList {
+		snapshotHandle := snapshot.SnapshotHandle
+		volumeHandle := snapshot.VolumeHandle
+
+		pv := utils.GetPersistentVolumeFromHandle(pvs, groupSnapshotContent.Spec.Driver, volumeHandle)
+		if pv == nil {
+			klog.Errorf(
```
jsafrane, line 559: This is an interesting case. A snapshot exists, but its PV has been deleted. How useful is to have a  VolumeSnapshot  / VolumeSnapshotContent for it, if the user cannot possibly guess from which PVC the snapshot is?

My gut feeling would be to record also the source PVC + PV name in `VolumeSnapshotHandlePairList` to be able to reconstruct the VolumeSnapshotContent / VolumeSnapshot with a link to the PV / PVC

This should be discussed in the KEP to keep track of the decision.

pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -485,6 +493,204 @@ func (ctrl *csiSnapshotCommonController) syncUnreadyGroupSnapshot(groupSnapshot
...
+		if groupSnapshotSecret != nil {
+			klog.V(5).Infof("createSnapshotsForGroupSnapshotContent: set annotation [%s] on volume snapshot content [%s].", utils.AnnDeletionSecretRefName, volumeSnapshotContent.Name)
+			metav1.SetMetaDataAnnotation(&volumeSnapshotContent.ObjectMeta, utils.AnnDeletionSecretRefName, groupSnapshotSecret.Name)
+
+			klog.V(5).Infof("createSnapshotsForGroupSnapshotContent: set annotation [%s] on volume snapshot content [%s].", utils.AnnDeletionSecretRefNamespace, volumeSnapshotContent.Name)
+			metav1.SetMetaDataAnnotation(&volumeSnapshotContent.ObjectMeta, utils.AnnDeletionSecretRefNamespace, groupSnapshotSecret.Namespace)
+		}
+
+		label := make(map[string]string)
+		label["volumeGroupSnapshotName"] = groupSnapshotContent.Spec.VolumeGroupSnapshotRef.Name
```
jsafrane, line 610: The label must have some `k8s.io` prefix, like `snapshot.storage.k8s.io/volumeGroupSnapshotName`

pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -485,6 +493,204 @@ func (ctrl *csiSnapshotCommonController) syncUnreadyGroupSnapshot(groupSnapshot
...
+				Name:       volumeSnapshotName,
+				Namespace:  volumeSnapshotNamespace,
+				Labels:     label,
+				Finalizers: []string{utils.VolumeSnapshotInGroupFinalizer},
+			},
+			Spec: crdv1.VolumeSnapshotSpec{
+				Source: crdv1.VolumeSnapshotSource{
+					VolumeSnapshotContentName: &volumeSnapshotContentName,
+				},
+			},
```
jsafrane, line 622: Again, please add a comment about who/when files the Status.

pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -485,6 +493,204 @@ func (ctrl *csiSnapshotCommonController) syncUnreadyGroupSnapshot(groupSnapshot
...
+				"createSnapshotsForGroupSnapshotContent: creating volumesnapshot %w", err)
+		}
+	}
+
+	// Phase 2: set the backlinks
+	klog.V(4).Infof(
+		"createSnapshotsForGroupSnapshotContent[%s]: linking volumesnapshots and volumesnapshotcontent for group snapshot content",
+		groupSnapshotContent.Name)
+
+	for _, snapshot := range groupSnapshotContent.Status.VolumeSnapshotHandlePairList {
```
jsafrane, line 643: Why a separate loop over VolumeSnapshotHandlePairList? It seems to me that all this loop does could be done in the first VolumeSnapshotHandlePairList loop few lines above.

pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -485,6 +493,204 @@ func (ctrl *csiSnapshotCommonController) syncUnreadyGroupSnapshot(groupSnapshot
...
+		}
+
+		_, err := ctrl.clientset.SnapshotV1().VolumeSnapshotContents().Create(context.TODO(), volumeSnapshotContent, metav1.CreateOptions{})
+		if err != nil {
+			return groupSnapshotContent, fmt.Errorf(
+				"createSnapshotsForGroupSnapshotContent: creating volumesnapshotcontent %w", err)
+		}
+
+		_, err = ctrl.clientset.SnapshotV1().VolumeSnapshots(volumeSnapshotNamespace).Create(context.TODO(), volumeSnapshot, metav1.CreateOptions{})
+		if err != nil {
```
jsafrane, line 632: Picking a very random error here.

What happens when any Create / Patch calls here fail? To me it _seems_ that a next sync loop will pick _different_ object names (because they are based on the current time) and create *new* contents and snapshots from scratch. Leaving the old ones orphaned. That does not look good.

I think the `GroupSnapshotContent.Name` and `GroupSnapshot.Name` are unique enough (are they?) and all VolumeSnapshotContents could be based on its name (like `groupSnapshotContent.Name + "-" + pv.Name`). Similarly, for VolumeSnapshots. The names should be stable for a given source PV / volumeHandle and VolumeGroupSnapshot.

And then the code should ignore `AlreadyExists` errors.

pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -485,6 +493,204 @@ func (ctrl *csiSnapshotCommonController) syncUnreadyGroupSnapshot(groupSnapshot
...
+		utils.GroupSnapshotterSecretParams,
+		groupSnapshotClass.Parameters,
+		groupSnapshotContent.GetObjectMeta().GetName(), nil)
+	if err != nil {
+		return groupSnapshotContent, fmt.Errorf(
+			"createSnapshotsForGroupSnapshotContent: failed to get secret reference for group snapshot content %s: %v",
+			groupSnapshotContent.Name, err)
+	}
+
+	pvs, err := ctrl.client.CoreV1().PersistentVolumes().List(context.TODO(), metav1.ListOptions{})
```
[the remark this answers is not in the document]
leonardoce: Agree with you. In this PR, I'm just moving the existing logic. Should we address it here, in your opinion?

pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -485,6 +493,204 @@ func (ctrl *csiSnapshotCommonController) syncUnreadyGroupSnapshot(groupSnapshot
...
+				Finalizers: []string{utils.VolumeSnapshotInGroupFinalizer},
+			},
+			Spec: crdv1.VolumeSnapshotSpec{
+				Source: crdv1.VolumeSnapshotSource{
+					VolumeSnapshotContentName: &volumeSnapshotContentName,
+				},
+			},
+		}
+
+		_, err := ctrl.clientset.SnapshotV1().VolumeSnapshotContents().Create(context.TODO(), volumeSnapshotContent, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
leonardoce: Yes. This comes out of the original implementation. Should we fix it here?

pkg/common-controller/groupsnapshot_controller_helper.go
leonardoce: Yes, good catch. This actually comes from https://github.com/kubernetes-csi/external-snapshotter/blame/33b4e6120ccc6427cf1ccc87b2314768ee37d754/pkg/sidecar-controller/groupsnapshot_helper.go#L452

pkg/common-controller/groupsnapshot_controller_helper.go
leonardoce: I asked myself the same question. Should we change it here?
https://github.com/kubernetes-csi/external-snapshotter/blob/33b4e6120ccc6427cf1ccc87b2314768ee37d754/pkg/sidecar-controller/groupsnapshot_helper.go#L477

pkg/common-controller/groupsnapshot_controller_helper.go
leonardoce: I think this deserves to be fixed here, even if this issue existed before. What's your opinion?

pkg/common-controller/groupsnapshot_controller_helper.go
leonardoce: I think this deserves to be fixed here, even if this issue existed before. What's your opinion?

pkg/common-controller/groupsnapshot_controller_helper.go
leonardoce: I agree with you, but I'm unaware of the underlying reasoning.

That's also why I changed the name of the original function `GetSnapshotContentNameForVolumeGroupSnapshotContent` to `CreateSnapshotContentNameForVolumeGroupSnapshotContent`.

Also, most bookkeeping could be simplified by an idempotent name.
I'm sure @xing-yang knows the context.

xing-yang: That's a good point. This timestamp is even in the KEP.  This came from the beginning when we still have VolumeGroup in the design.  Though it does not look like we need it any more.  @leonardoce can you get rid of it for both VolumeSnapshot name and VolumeSnapshotContent name?

pkg/common-controller/groupsnapshot_controller_helper.go
xing-yang: We retrieve PV here to populate groupSnapshotContent.Status.PVVolumeSnapshotContentList and set volumeSnapshotContent.Spec.SourceVolumeMode.
- The SourceVolumeMode will have an impact on applications that rely on changing volume mode when creating PVC from VolumeSnapshot. Without it being set, it falls back to the default behavior (for backward compatibility) which is not to check it. 
- Regarding PVVolumeSnapshotContentList, we have an action item to see if we can get rid of it after making all other changes.
I'm not sure if it is worth it to add source PVC and PV name in VolumeSnapshotHandlePairList.  That sounds like the wrong place to store this info. We can discuss more about it.

pkg/common-controller/groupsnapshot_controller_helper.go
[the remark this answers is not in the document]
leonardoce: I added the informer and the indexer in a separate commit, so we can decide whether to include it in this PR or in a separate one.

pkg/common-controller/groupsnapshot_controller_helper.go
leonardoce: Done.

pkg/common-controller/groupsnapshot_controller_helper.go
leonardoce: There is no good reason besides it being easier on my eyes.
Now the names of members [VolumeSnapshot and VolumeSnapshotContent are consistent](https://github.com/kubernetes-csi/external-snapshotter/pull/1171#discussion_r1826145289), so there's no reason for two loops anymore.

pkg/common-controller/groupsnapshot_controller_helper.go
leonardoce: Sure! Now, the names are consistent.

pkg/common-controller/groupsnapshot_controller_helper.go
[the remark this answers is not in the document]
leonardoce: Xing asked me to put that on a separate PR. I'm on it.

leonardoce:
I rebased it on the latest `master` and now I'm filing a separate PR for the added informer and lister.

pkg/common-controller/groupsnapshot_controller_helper.go
xing-yang: Let me open an issue and we'll get back to this one later.

jsafrane:
The amount of issues and subsequent PRs worries me a bit, but I can see it's moving towards a better controller.

PR #1182, opened 2024-11-04
Informer lister

leonardoce:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

The VolumeGroupSnapshot feature requires looking up the PV object corresponding to a volumeHandle for a known CSI Driver. At this moment, the code uses a List call on the API server, and List calls are particularly expensive.

This patch adds an informer whose lister and indexer are used to alleviate the cost of that call.

Fixes #

This PR depends on https://github.com/kubernetes-csi/external-snapshotter/pull/1171 to be applied. This is why it is marked as a draft.

```release-note
NONE
```


pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -659,6 +652,39 @@ func (ctrl *csiSnapshotCommonController) createSnapshotsForGroupSnapshotContent(
 	return newGroupSnapshotObj, nil
 }
 
+// findPersistentVolumeByCSIDriverHandle looks us an existing PersistentVolume
```
xing-yang, line 655: looks us -> looks at

jsafrane:
/lgtm
/approve

Nicely written!

Please remember to update all other `PersistentVolumes().Get()` to use the informer or add a comment there why it's not possible.

PR #1073, opened 2024-04-24
Adding validation rules into CRDs for validation

cici37:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
This is to add validation rules to CRDs which did the same validation as written in https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/validation-webhook/validation.go

**Which issue(s) this PR fixes**:

Fixes #1072 

**Special notes for your reviewer**:
The feature is available since Kubernetes 1.25. Is there any version compatibility concern there? Could I remove the existing validating webhook in the same PR? Thank you!

**Does this PR introduce a user-facing change?**:

```release-note
Adds validation rules into CRDs. Minimum required Kubernetes version is 1.25 for these validation rules.
```

PR #1184, opened 2024-11-05
Remove volumeGroupSnapshotRef immutable validation rule

xing-yang:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

We need to remove the validation rule that ensures volumeGroupSnapshotRef is immutable. 
This is because that UID needs to be set by the snapshot controller after volumeGroupSnapshotRef is populated with name and namespace for pre-provisioned VolumeGroupSnapshotContent.

This immutable validation rule was added as part of this PR https://github.com/kubernetes-csi/external-snapshotter/pull/1073 to use CEL to replace the validation added in the validation webhook:
```
	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="volumeGroupSnapshotRef is immutable"
```
However, in the original validation webhook, we never had this rule:
https://github.com/kubernetes-csi/external-snapshotter/blob/v7.0.2/pkg/validation-webhook/validation.go#L52
With this rule, the snapshot controller won't be able to set UID for volumeGroupSnapshotRef after it was populated with name and namespace initially. So we need to remove this rule.

Fixes #

```release-note
Remove the rule that ensures volumeGroupSnapshotRef is immutable. UID needs to be set by the snapshot controller after volumeGroupSnapshotRef is populated with name and namespace for pre-provisioned VolumeGroupSnapshotContent.
```

### DIFF
--- a/client/apis/volumegroupsnapshot/v1alpha1/types.go
+++ b/client/apis/volumegroupsnapshot/v1alpha1/types.go
@@ -281,7 +281,6 @@ type VolumeGroupSnapshotContentSpec struct {
 	// This field is immutable after creation.
 	// Required.
 	// +kubebuilder:validation:XValidation:rule="has(self.name) && has(self.__namespace__)",message="both volumeGroupSnapshotRef.name and volumeGroupSnapshotRef.namespace must be set"
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="volumeGroupSnapshotRef is immutable"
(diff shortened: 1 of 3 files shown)


leonardoce:
I tested the volume group snapshot static and dynamic provisioning, and everything is fine.

PR #1194, opened 2024-11-14
include feature gate in snapshotter

yati1998:
this commit includes feature gate method
to enable the volumeGroupSnapshot

```release-note
enable-volumegroup-snapshot flag has been replaced by feature-gates flag.
enables feature gate to enable volumegroupsnapshot instead of using flags like --feature-gates=CSIVolumeGroupSnapshot=true
By default the feature gate is disabled
```


pkg/features/features.go
```diff
@@ -0,0 +1,36 @@
...
+
+package features
+
+import (
+	"k8s.io/apiserver/pkg/util/feature"
+	"k8s.io/component-base/featuregate"
+)
+
+const (
+	VolumeGroupSnapshot featuregate.Feature = "volumeGroupSnapshot"
```
nixpanic, line 25: Probably should be called `"CSIVolumeGroupSnapshot"` to be in line with https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

cmd/snapshot-controller/main.go
```diff
@@ -73,7 +75,7 @@ var (
 	retryIntervalMax              = flag.Duration("retry-interval-max", 5*time.Minute, "Maximum retry interval of failed volume snapshot creation or deletion. Default is 5 minutes.")
 	enableDistributedSnapshotting = flag.Bool("enable-distributed-snapshotting", false, "Enables each node to handle snapshotting for the local volumes created on that node")
 	preventVolumeModeConversion   = flag.Bool("prevent-volume-mode-conversion", true, "Prevents an unauthorised user from modifying the volume mode when creating a PVC from an existing VolumeSnapshot.")
-	enableVolumeGroupSnapshots    = flag.Bool("enable-volume-group-snapshots", false, "Enables the volume group snapshot feature, allowing the user to create a snapshot of a group of volumes.")
+	//enableVolumeGroupSnapshots    = flag.Bool("enable-volume-group-snapshots", false, "Enables the volume group snapshot feature, allowing the user to create a snapshot of a group of volumes.")
```
nixpanic, line 78: Not sure what usually is done, but deprecating an option usually involves keeping the it, but ignoring it. That prevents errors when a deployment has it set, and the container-image is updated.

nixpanic (commented):
/ok-to-test

cmd/snapshot-controller/main.go
yati1998: it is showing a warning if we keep this variable because it is no more used anywhere. So I think it would be fine removing this

yati1998:
@xing-yang @nixpanic please do review the PR,
I have tested it, this is working good with feature gate

yati1998:
```
 containers:
  - args:
    - --feature-gates=CSIVolumeGroupSnapshot=true
    - --v=5
    - --leader-election=true
    image: quay.io/ypadia/snapshotcontroller:featuregate2
    imagePullPolicy: IfNotPresent
    name: snapshot-controller
```

cmd/csi-snapshotter/main.go
```diff
@@ -28,9 +28,11 @@ import (
 
 	"google.golang.org/grpc"
 
+	flag "github.com/spf13/pflag"
```
leonardoce, line 31: Do we really need to have both `flag` and `github.com/spf13/pflag` at the same time?
`utilflag.NewMapStringBool` can work with plain `flag` too.

cmd/csi-snapshotter/main.go
```diff
@@ -98,10 +103,18 @@ var (
 )
 
 func main() {
+	flag.Var(utilflag.NewMapStringBool(&featureGates), "feature-gates", "A set of key=value pairs that describe feature gates for alpha/experimental features. "+
```
leonardoce, line 106: see previous point

cmd/csi-snapshotter/main.go
```diff
@@ -98,10 +103,18 @@ var (
...
+		"Options are:\n"+strings.Join(utilfeature.DefaultFeatureGate.KnownFeatures(), "\n"))
+
 	klog.InitFlags(nil)
+	flag.CommandLine.AddGoFlagSet(goflag.CommandLine)
```
leonardoce, line 110: see previous point

cmd/csi-snapshotter/main.go
```diff
@@ -98,10 +103,18 @@ var (
...
 	flag.Set("logtostderr", "true")
 	flag.Parse()
 
+	if err := utilfeature.DefaultMutableFeatureGate.SetFromMap(featureGates); err != nil {
+		klog.Fatal(err)
```
leonardoce, line 115: We could include more information than just the error message. I.e. we could explain that this is happening while we're parsing the `-feature-gates` option

cmd/snapshot-controller/main.go
leonardoce: Let's remove this line if we don't have the `enable-volume-group-snapshots` flag.

cmd/snapshot-controller/main.go
```diff
@@ -147,10 +153,18 @@ func ensureCustomResourceDefinitionsExist(client *clientset.Clientset, enableVol
...
+	flag.Var(utilflag.NewMapStringBool(&featureGates), "feature-gates", "A set of key=value pairs that describe feature gates for alpha/experimental features. "+
+		"Options are:\n"+strings.Join(utilfeature.DefaultFeatureGate.KnownFeatures(), "\n"))
+
 	klog.InitFlags(nil)
+	flag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 	flag.Set("logtostderr", "true")
 	flag.Parse()
 
+	if err := utilfeature.DefaultMutableFeatureGate.SetFromMap(featureGates); err != nil {
+		klog.Fatal(err)
```
leonardoce, line 165: See previous point

cmd/csi-snapshotter/main.go
yati1998: yeah but `AddGoFlagSet` required pflag

cmd/csi-snapshotter/main.go
yati1998: I tried using flag package for this, but it shows error

leonardoce: Which error are you referring to? Can you share it here?

cmd/snapshot-controller/main.go
leonardoce: I agree with @nixpanic and I'd like to avoid breaking existing deployments.
If we found it specified, we could warn the user that the option is deprecated.

cmd/csi-snapshotter/main.go
leonardoce: That's exactly what I was asking. Why do we need two flag sets?

pkg/features/features.go
```diff
@@ -0,0 +1,36 @@
...
+
+package features
+
+import (
+	"k8s.io/apiserver/pkg/util/feature"
+	"k8s.io/component-base/featuregate"
+)
+
+const (
+	VolumeGroupSnapshot featuregate.Feature = "CSIVolumeGroupSnapshot"
```
leonardoce, line 25: `VolumeGroupSnapshot` is a public constant. I'd add a comment describing it.

pkg/features/features.go
```diff
@@ -0,0 +1,37 @@
...
+	"k8s.io/component-base/featuregate"
+)
+
+const (
+	// Enable usage of volume group snapshot
+	VolumeGroupSnapshot featuregate.Feature = "CSIVolumeGroupSnapshot"
+)
+
+func init() {
+	feature.DefaultMutableFeatureGate.Add(defaultKubernetesFeatureGates)
```
xing-yang, line 30: VolumeGroupSnapshot is a new API.  It needs to be disabled by default when moving to Beta.

cmd/snapshot-controller/main.go
```diff
@@ -228,10 +240,10 @@ func main() {
 		workqueue.NewItemExponentialFailureRateLimiter(*retryIntervalStart, *retryIntervalMax),
 		*enableDistributedSnapshotting,
 		*preventVolumeModeConversion,
-		*enableVolumeGroupSnapshots,
+		utilfeature.DefaultFeatureGate.Enabled(features.VolumeGroupSnapshot),
```
xing-yang, line 243: VolumeGroupSnapshot needs to be disabled by default when moving to Beta initially.

pkg/features/features.go
yati1998: by default it is false only right? check line no 36?

cmd/snapshot-controller/main.go
yati1998: this just checks if it is enabled or not, but in features, default value is false only

cmd/csi-snapshotter/main.go
```diff
@@ -234,7 +245,7 @@ func main() {
 
 	snapShotter := snapshotter.NewSnapshotter(csiConn)
 	var groupSnapshotter group_snapshotter.GroupSnapshotter
-	if *enableVolumeGroupSnapshots {
```
leonardoce, line 237: We deprecated `enable-volume-group-snapshots`, didn't we?
If somebody used that option, I'd file a log message explaining that we no longer care about it.

gnufied: We can break alpha features in releases. We should be okay. :-)

cmd/csi-snapshotter/main.go
```diff
@@ -90,6 +93,7 @@ var (
```
xing-yang, line 93: Please remove this flag enableVolumeGroupSnapshots and add a release note that this is replaced by the feature gate.

cmd/csi-snapshotter/main.go
leonardoce: Yep, but a log message is three lines away and may save several headaches.
In my opinion, we should either log a message or remove the `enable-volume-group-snapshots` option entirely.

yati1998: yes as per xing's comment , I have removed the flag itself, mentioned in relase note

yati1998:
@gnufied can you give your final reviews as well?

xing-yang:
In the release notes, can you add that the feature gate is disabled by default?

yati1998:
> In the release notes, can you add that the feature gate is disabled by default?

updated!! please review it again

xing-yang:
You need to change `--feature-gates=CSIVolumeGroupSnapshot=false` to `--feature-gates=CSIVolumeGroupSnapshot=true` in the release note?

huww98:
We should cleanup `--enable-volume-group-snapshots` in readme and error message. There are still 6 mentions.

xing-yang:
@yati1998 Can you take care of this?

yati1998:
> @yati1998 Can you take care of this?

noted

PR #1200, opened 2024-11-20
Remove PVCVolumeSnapshotRefList from VolumeGroupSnapshot API

### LINKED ISSUE #1202: Remove PVCVolumeSnapshotRefList from VolumeGroupSnapshot API
**What happened**:
`PVCVolumeSnapshotRefList` was originally added to the Status of VolumeGroupSnapshot API so that a user knows which individual VolumeSnapshot is corresponding to which PVC when doing a restore: https://github.com/kubernetes-csi/external-snapshotter/issues/969

For dynamic provisioning of VolumeGroupSnapshot, we changed `volumeSnapshotContentName` to `persistentVolumeClaimName` in the Source field of individual VolumeSnapshot objects, and changed `snapshotHandle` to `volumeHandle` in the Source field of individual VolumeSnapshotContent objects. So we no longer need `PVCVolumeSnapshotRefList` in the Status of VolumeGroupSnapshot for the purpose of finding the matching PVC to do restore from a VolumeSnapshot.

For static provisioning, the VolumeSnapshot Source only contains `volumeSnapshotContentName` and the VolumeSnapshotContent Source only contains `snapshotHandle`. This means the original individual volume snapshot API objects don't contain any information about the PVC name or volume handle. It is assumed that the admin should have this information. Therefore the VolumeGroupSnapshot API does not need to provide extra information about the source PVCs.

Therefore it is now safe to remove `PVCVolumeSnapshotRefList` from VolumeGroupSnapshot API.

**How to reproduce it**:

**Environment**:
- Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

leonardoce:
**What type of PR is this?**
>
/kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR removes the `PVCVolumeSnapshotRefList` entry from the VolumeGroupSnapshot API. This is no longer needed since we now set `persistentVolumeClaimName` in the source of member VolumeSnapshots

Fixes #1202 

```release-note
The `volumegroupsnapshot.status.pvcVolumeSnapshotRefList` field has been removed. VolumeShapshots members of a dynamically provisioned VolumeGroupSnapshot will have their `persistentVolumeClaimName` set, allowing the consumer to map the PVC being snapshotted with the corresponding snapshot.
```

### DIFF
--- a/client/apis/volumegroupsnapshot/v1alpha1/types.go
+++ b/client/apis/volumegroupsnapshot/v1alpha1/types.go
@@ -108,12 +108,6 @@ type VolumeGroupSnapshotStatus struct {
 	// group snapshot creation. Upon success, this error field will be cleared.
 	// +optional
 	Error *snapshotv1.VolumeSnapshotError `json:"error,omitempty" protobuf:"bytes,4,opt,name=error,casttype=VolumeSnapshotError"`
-
-	// VolumeSnapshotRefList is the list of PVC and VolumeSnapshot pairs that
-	// is part of this group snapshot.
-	// The maximum number of allowed snapshots in the group is 100.
-	// +optional
-	PVCVolumeSnapshotRefList []PVCVolumeSnapshotPair `json:"pvcVolumeSnapshotRefList,omitempty" protobuf:"bytes,5,opt,name=pvcVolumeSnapshotRefList"`
 }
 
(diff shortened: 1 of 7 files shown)


xing-yang:
Can you add a note about the label in the PR description?

PR #1221, opened 2024-11-25
use golang 1.22.9

Madhu-1:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

There is no hard requirement to use the latest 1.23 golang release. this commit update the version to use
latest 1.22.9 release. The update to 1.23 was done to fix CVE, 1.22.9 also fixes the same problem and not many peoples/companies are using latest version of 1.23 for building

Fixes #

```release-note
use golang 1.22.9 for external-snapshotter
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/kubernetes-csi/external-snapshotter/v8
 
-go 1.23.1
+go 1.22.9
 
 require (
 	github.com/container-storage-interface/spec v1.11.0


Madhu-1:
https://github.com/kubernetes-csi/external-snapshotter/blob/master/client/go.mod#L3-L5 we are using 1.22 here

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kubernetes-csi/external-snapshotter/v8
 
-go 1.23.1
+go 1.22.9
```
xing-yang, line 3: Kubernetes already moved to 1.23.3: https://github.com/kubernetes/kubernetes/blob/master/.go-version
We can't move back.

PR #1238, opened 2024-12-09
Fix Snapshot Controller's unbounded VolumeSnapshot list call on startup

AndrewSirenko:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

As of v6.2.2, the snapshot controller performs a list operation on startup as a way of validating the Snapshot CRD is installed on the cluster. v7.0.0 attempts to put a limit on this list call of 0, but [that is treated as an unbounded request by etcd](https://github.com/kubernetes/kubernetes/blob/249ad2a6137cc8f1e0ccb7f0aef9ff4ba38927b9/vendor/go.etcd.io/etcd/client/v3/kubernetes/interface.go#L75). 

For clusters with many VolumeSnapshots, the API Server may struggle to handle an unbounded list request within 60s, leading to continuous restarts in the snapshot controller. 

@wongma7 proposes a better longterm solution [here](https://github.com/kubernetes-csi/external-snapshotter/pull/1238#discussion_r1876893801), but this PR introduces a short-term fix of properly setting the list response limit to 1.  

Fixes #

Within k8s API server, paging gets set to false [here](https://github.com/kubernetes/apiserver/blob/3658357fea9fa8b36173d072f2d548f135049e05/pkg/storage/etcd3/store.go#L713) because limit <= 0

We pass through limit = 0 to store.getList [here](https://github.com/kubernetes/apiserver/blob/3658357fea9fa8b36173d072f2d548f135049e05/pkg/storage/etcd3/store.go#L749)

For etcd, limit of 0 allegedly means no limitation [documented here](https://github.com/kubernetes/kubernetes/blob/249ad2a6137cc8f1e0ccb7f0aef9ff4ba38927b9/vendor/go.etcd.io/etcd/client/v3/kubernetes/interface.go#L75)

```release-note
Fix unbounded volumesnapshots list call on Snapshot Controller startup
```

### DIFF
--- a/cmd/snapshot-controller/main.go
+++ b/cmd/snapshot-controller/main.go
@@ -88,9 +88,9 @@ var version = "unknown"
 func ensureCustomResourceDefinitionsExist(client *clientset.Clientset, enableVolumeGroupSnapshots bool) error {
 	condition := func(ctx context.Context) (bool, error) {
 		var err error
-		// List calls should return faster with a limit of 0.
+		// List calls should return faster with a limit of 1.
 		// We do not care about what is returned and just want to make sure the CRDs exist.
-		listOptions := metav1.ListOptions{Limit: 0}
+		listOptions := metav1.ListOptions{Limit: 1}
 
 		// scoping to an empty namespace makes `List` work across all namespaces
 		_, err = client.SnapshotV1().VolumeSnapshots("").List(ctx, listOptions)


cmd/snapshot-controller/main.go
```diff
@@ -88,9 +88,9 @@ var version = "unknown"
 func ensureCustomResourceDefinitionsExist(client *clientset.Clientset, enableVolumeGroupSnapshots bool) error {
 	condition := func(ctx context.Context) (bool, error) {
 		var err error
-		// List calls should return faster with a limit of 0.
+		// List calls should return faster with a limit of 1.
 		// We do not care about what is returned and just want to make sure the CRDs exist.
-		listOptions := metav1.ListOptions{Limit: 0}
+		listOptions := metav1.ListOptions{Limit: 1}
```
xing-yang, line 93: List snapshots is an expensive call.  We should try to avoid listing all the snapshots.
The purpose for this call here is to ensure the CRD is installed. We don't really need to list the snapshots.

xing-yang: I see your question here: https://kubernetes.slack.com/archives/C0EG7JC6T/p1733776266305359
It's odd that {Limit: 0} would return 10,000 entries.

AndrewSirenko: Agreed, at the very least can we wait for an informer sync of volumesnapshots instead of making a list call?

wongma7: According to https://github.com/kubernetes-csi/external-snapshotter/pull/504 which added this ensureCustomResourceDefinitionsExist waiting for informer sync was not sufficient bc the controller showed as Ready even before informer had synced, so ensureCustomResourceDefinitionsExist was added to force the controller to immediately crash. So I guess it's not as simple as removing this function , I think the most correct solution would be to implement readyz probe which waited for informer sync. But that is a lot of effort so I think this PR is the best short term fix.

(As for the limit=0 behavior, indeed it is counterintuitive & I could not find where it's documented in apiserver code if at all It seems that since resourceversion is not set in this list call then the lImit=0 option gets plumbed to etcd list call here and it *does* document that "// If WithLimit is given a 0 limit, it is treated as no limit." https://github.com/etcd-io/etcd/blob/854bdd646c8ce50b879f79f403726c7ab0dc726c/client/v3/op.go#L345. Anyway, the details are unimportant, the point is we observe that limit=0 does the opposite of intended!!)

wongma7: Just saw that external-provisioner has a healt hz check https://github.com/kubernetes-csi/external-provisioner/commit/52c3575b970ec0a9a1972579e5b6a99f5860a228#diff-db1b4e664fc5c6203530702df4e7c6eade0fd481ef5badffbc0bb4a92648d36e. So in short mb snapshot-controller should have also a healthz check which starts serving OK only after informer synced.

AndrewSirenko:
Tested on local kops cluster with 10.291k volumesnapshots. Added a log statement:

With this PR (limit=1)
"Found volumesnapshots with" len=1
---
Without PR limit=0
"Found volumesnapshots with" len=10291

xing-yang:
lgtm

mauriciopoppe:
/lgtm

Thanks!

cmd/snapshot-controller/main.go
xing-yang: Discussed in the CSI Implementation meeting today and decided we should not go with the "health check" fix. If the CRDs are not deployed, informer will returns errors. It's not going to be synced.  So the current fix is good.

PR #1284, opened 2025-03-22
Add --poll-snapshot-period for periodic readiness requeue check when creating a VolumeSnapshotContent

### LINKED ISSUE #1282: Improve csi-snapshotter VolumeSnapshotContent requeue fairness

This enhancement is to improve the `requeue` behavior for syncing VolumeSnapshotContent resources.

VolumeSnapshotContent resources are reconciled via the [contentQueue](https://github.com/kubernetes-csi/external-snapshotter/blob/59d72977e7a24ccdfabed3a20720e41beddc8216/pkg/sidecar-controller/snapshot_controller_base.go#L112). For snapshots that are long running, this can be very useful to reduce the amount of polling required to determine if a snapshot is `readyToUse=true`. However the exponential nature of this backoff can result in the contentQueue rate limiter quickly reaching the maximum. The current default is `1 second`, and the current maximum is `300 seconds`. This only requires [9 requeue events] to reach the maximum. This limit can quickly be reached today, if a VolumeSnapshotContent is updated. Updates (especially re-entrant updates) trigger resync and requeue, which can quickly bump up the rate limiter retry number, resulting in long requeue wait times.

There are two things that should be fixed here:
1) Prevent updates from bumping the `requeue` rate limiter limit: Ideally, a additional call to `contentQueue.AddRateLimited()` should not increase the rate limiter exponent if an item is already scheduled to be requeued. It should either maintain the same requeue schedule, or be adjusted to requeue further into the future, but with the same backoff exponent.
2) Reduce the number of re-entrant updates. This can reduce the number of requeues (which can lead to the problem above). Some updates are necessary for tracking the lifecycle VolumeSnapshotContent. However it appears that the `snapshot.storage.kubernetes.io/volumesnapshot-being-created` annotation [can be removed early during](https://github.com/kubernetes-csi/external-snapshotter/blob/59d72977e7a24ccdfabed3a20720e41beddc8216/pkg/sidecar-controller/snapshot_controller.go#L391), prior to the snapshot actually being marked as `readyToUse`.

A quick fix alternative is just to decrease the max exponential backoff of `contentQueue` to a lower default (eg: 30 seconds, 60 seconds). This can be used by a CO to reduce the likelihood of higher latency VolumeSnapshotContent reconciliation.

**Additional context**

pwschuurman:
**What type of PR is this?**
/kind feature

Today, the csi-snapshotter runs with a reconciler pattern when processing VolumeSnapshotContent resources. This follows the familiar `requeue, err` pattern that is followed by other controller frameworks, like controller-runtime.
 * Error Handling: When an error is encountered, VolumeSnapshotContent resources will be requeued with exponential decay. This behavior is desired, as presumably RPC calls to a CSI driver are due to congestion, and the request rate should be reduced.
 * Requeue Processing: In the case of dynamic provisioning, the creation of a VolumeSnapshotContent may require some time until a snapshot is ready in the underlying storage system. This is where the `requeue` field is useful, as the reconciler can periodically attempt to check the status of a snapshot.

The problem with the current requeue implementation is that it relies on the same exponential backoff rate limiter queue as the error handling scenario. This is a problem, as it can lead to non-deterministic ready times, and can lead to significantly higher effective snapshot ready latencies for snapshots that take longer to process (eg: larger snapshots).

To fix this, the requeue logic should use a constant retry period, to allow for more reliable snapshot reconciliation. This PR adds a new command line argument `--poll-snapshot-period` to the csi-snapshotter, with a default of `10 seconds`.

Fixes #1282

```release-note
Requeue VolumeSnapshotContent resources based on --poll-snapshot-period in csi-snapshotter
```

### DIFF
--- a/cmd/csi-snapshotter/main.go
+++ b/cmd/csi-snapshotter/main.go
@@ -93,6 +93,7 @@ var (
 	metricsAddress              = flag.String("metrics-address", "", "(deprecated) The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled. Only one of `--metrics-address` and `--http-endpoint` can be set.")
 	httpEndpoint                = flag.String("http-endpoint", "", "The TCP network address where the HTTP server for diagnostics, including metrics and leader election health check, will listen (example: `:8080`). The default is empty string, which means the server is disabled. Only one of `--metrics-address` and `--http-endpoint` can be set.")
 	metricsPath                 = flag.String("metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
+	pollSnapshotPeriod          = flag.Duration("poll-snapshot-period", 10*time.Second, "Poll interval to check if a snapshot is ready. Default is 10 seconds.")
 	retryIntervalStart          = flag.Duration("retry-interval-start", time.Second, "Initial retry interval of failed volume snapshot creation or deletion. It doubles with each failure, up to retry-interval-max. Default is 1 second.")
 	retryIntervalMax            = flag.Duration("retry-interval-max", 5*time.Minute, "Maximum retry interval of failed volume snapshot creation or deletion. Default is 5 minutes.")
 	enableNodeDeployment        = flag.Bool("node-deployment", false, "Enables deploying the sidecar controller together with a CSI driver on nodes to manage snapshots for node-local volumes.")
@@ -297,6 +298,7 @@ func main() {
 		snapshotContentfactory.Groupsnapshot().V1beta1().VolumeGroupSnapshotContents(),
 		snapshotContentfactory.Groupsnapshot().V1beta1().VolumeGroupSnapshotClasses(),
 		workqueue.NewTypedItemExponentialFailureRateLimiter[string](*retryIntervalStart, *retryIntervalMax),
+		*pollSnapshotPeriod,
 	)
 
 	run := func(context.Context) {
--- a/pkg/sidecar-controller/framework_test.go
+++ b/pkg/sidecar-controller/framework_test.go
@@ -580,6 +580,7 @@ func newTestController(kubeClient kubernetes.Interface, clientset clientset.Inte
 		informerFactory.Groupsnapshot().V1beta1().VolumeGroupSnapshotContents(),
 		informerFactory.Groupsnapshot().V1beta1().VolumeGroupSnapshotClasses(),
 		workqueue.NewTypedItemExponentialFailureRateLimiter[string](1*time.Millisecond, 1*time.Minute),
+		10*time.Second, /* pollSnapshotPeriod */
 	)
 
 	ctrl.eventRecorder = record.NewFakeRecorder(1000)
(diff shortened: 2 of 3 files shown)


sunnylovestiramisu:
Question: 

1. Why is the default 10 seconds? Is any data we looked at that is supporting this value?
2. Do you think it makes sense to separate two queues for this case? One for deterministic ready times and the other for non-deterministic ready times.

cmd/csi-snapshotter/main.go
```diff
@@ -93,6 +93,7 @@ var (
 	metricsAddress              = flag.String("metrics-address", "", "(deprecated) The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled. Only one of `--metrics-address` and `--http-endpoint` can be set.")
 	httpEndpoint                = flag.String("http-endpoint", "", "The TCP network address where the HTTP server for diagnostics, including metrics and leader election health check, will listen (example: `:8080`). The default is empty string, which means the server is disabled. Only one of `--metrics-address` and `--http-endpoint` can be set.")
 	metricsPath                 = flag.String("metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
+	pollSnapshotPeriod          = flag.Duration("poll-snapshot-period", 10*time.Second, "Poll interval to check if a snapshot is ready. Default is 10 seconds.")
```
xing-yang, line 96: Can you set the default to 0 so that drivers can opt-in to this new behavior?

pwschuurman: > Can you set the default to 0 so that drivers can opt-in to this new behavior?

I think changing the default is a benefit for end users here, so I would prefer to set it to a reasonable poll interval. We could increase the poll interval higher, if scalability is a concern. However if we do get a 429 from the APIServer, we'll encounter an error and then leverage exponential backoff.

I can make the default 0, and then refactor the logic to fallback on the exponential backoff when `pollSnapshotPeriod` is not set (eg: we need to adjust when `Forget()` is called if we do this).

@xing-yang What would you prefer (either keeping the logic as is but modifying the `pollSnapshotPeriod`, or refactoring to use the exponential rate limiting poll interval as a fallback).

pwschuurman:
@sunnylovestiramisu 

> Why is the default 10 seconds? Is any data we looked at that is supporting this value?

Good call out. I don't have any data for setting a default. A default of a minute is may be more reasonable. This is the CSI connection timeout: https://github.com/kubernetes-csi/external-snapshotter/blob/7e305533b2c96901f61d1e9ef3822b2b97df8d52/cmd/csi-snapshotter/main.go#L70

I think the CreateSnapshot RPC is unique in that it is one of the RPCs that can return OK, but may require a retry. Other RPCs (eg: CreateVolume) are pass/fail. For CSI Provisioner, it seems that we used to have the default poll interval set at 10 seconds (eg: some reference here https://github.com/kubernetes-csi/external-provisioner/issues/131).

> Do you think it makes sense to separate two queues for this case? One for deterministic ready times and the other for non-deterministic ready times.

So you're thinking in the event of an error, we add a PVC to an "error" processing queue (that has a ExponentialRateLimiter), and we have a "retry" processing queue? I think that is more to manage, and there is greater risk of concurrency related issues. We'd need to track resource keys across two queues, and we need to listen to resources available on both queues.

PR #751, opened 2022-08-20
add helm charts

kahirokunn:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:

helm can be conveniently installed with the helm command, etc.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

I also have a command that can update the helm chart based on `client/config/crd` and `deploy/kubernetes/snapshot-controller`

eg. `sh ./release-tools/release-helm-charts.sh`

**Does this PR introduce a user-facing change?**:

```release-note

Add Helm Charts

```

PR #622, opened 2021-12-01
Add Helm Chart manifests and GitHub Release Helm Chart for the snapshot-controller component 

kaskol10:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind design

**What this PR does / why we need it**:

The changes introduced with this PR allow for better integration snapshot-controller and Helm. Besides, publish the Helm Chart in the GitHub Releases Pages using a GitHub Action.

**Which issue(s) this PR fixes**:

Fixes [#551](https://github.com/kubernetes-csi/external-snapshotter/issues/551)

**Special notes for your reviewer**:
A GitHub branch called `gh-pages` would be needed to store the published charts. It's a pre-requisite for [chart-releaser-action](https://github.com/helm/chart-releaser-action#pre-requisites)
**Does this PR introduce a user-facing change?**:

```release-note
Added Helm Chart manifests and Helm Chart GitHub Release for snapshot-controller component
```

PR #1285, opened 2025-03-24
Create Helm Chart from Static Manifests

### LINKED ISSUE #551: Helm chart for CRDs and snapshot controller?
Are there any plans to offer snapshot CRDs, controller, etc. as a single Helm chart? If not, would you accept contributions in this area?

I realize that all these currently external components (CRDs, controller) will probably be incorporated in Kubernetes some time in the future, so maybe it doesn't make sense having to maintain a Helm chart for these.

hswong3i:
**What type of PR is this?**
/kind feature

This PR convert our static deployment manifests from `deploy/kubernetes/*` into Helm Charts under `charts/*`, by running a convert script under `deploy/util/chart-releaser.sh`.

Moreover, it handle the chart release with branch `gh-pages`, by integrate with https://github.com/helm/chart-releaser-action.

The sample helm repo could find from https://alvistack.github.io/kubernetes-csi-external-snapshotter/index.yaml, by:

    helm repo add external-snapshotter https://alvistack.github.io/kubernetes-csi-external-snapshotter
    helm repo update
    helm search repo external-snapshotter

Before each stable tag release, maintainers only required to:
1. Run the `./deploy/util/chart-releaser.sh` and update templates under `charts/*/templates/*`
2. Double confirm `charts/*/values.yml` with correct values, e.g. image tags
3. `git add --all --force charts/* && git commit`

Fixes #551
Fixes #622
Fixes #751

```release-note
Create Helm Chart from Static Manifests
```

### DIFF
--- a/charts/csi-snapshotter/Chart.yaml
+++ b/charts/csi-snapshotter/Chart.yaml
@@ -1,0 +1,24 @@
+apiVersion: v2
+name: csi-snapshotter
+description: A Helm chart for deploy the CSI snapshotter together with the hostpath CSI driver
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: v8.2.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: v8.2.0
--- a/charts/csi-snapshotter/templates/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
+++ b/charts/csi-snapshotter/templates/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
@@ -1,0 +1,91 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.kubernetes.io: https://github.com/kubernetes-csi/external-snapshotter/pull/1150
+    controller-gen.kubebuilder.io/version: v0.15.0
+  name: volumegroupsnapshotclasses.groupsnapshot.storage.k8s.io
+spec:
+  group: groupsnapshot.storage.k8s.io
+  names:
+    kind: VolumeGroupSnapshotClass
+    listKind: VolumeGroupSnapshotClassList
+    plural: volumegroupsnapshotclasses
+    shortNames:
+      - vgsclass
+      - vgsclasses
+    singular: volumegroupsnapshotclass
+  scope: Cluster
+  versions:
+    - additionalPrinterColumns:
+        - jsonPath: .driver
+          name: Driver
+          type: string
+        - description: Determines whether a VolumeGroupSnapshotContent created through the VolumeGroupSnapshotClass should be deleted when its bound VolumeGroupSnapshot is deleted.
+          jsonPath: .deletionPolicy
+          name: DeletionPolicy
+          type: string
+        - jsonPath: .metadata.creationTimestamp
+          name: Age
+          type: date
+      name: v1beta1
+      schema:
+        openAPIV3Schema:
+          description: |-
+            VolumeGroupSnapshotClass specifies parameters that a underlying storage system
+            uses when creating a volume group snapshot. A specific VolumeGroupSnapshotClass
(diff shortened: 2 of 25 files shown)


charts/csi-snapshotter/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: v8.2.0
```
TerryHowe, line 18: Are you expecting that the chart would never change for an csi-snapshotter version? So a chart change would go out as a bug fix maybe?

charts/csi-snapshotter/templates/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
```diff
@@ -0,0 +1,91 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
TerryHowe, line 2: I'm not sure what the official helm recommendation is on CRDs, but I like to put them in another helm chart in case there is a migration. Maybe if your CRD doesn't change much, you could handle migrations differently.

Helm does have a HIP around CRD handling https://github.com/helm/community/pull/379

charts/csi-snapshotter/values.yaml
```diff
@@ -0,0 +1,18 @@
+---
+csiSnapshotter:
+  replicaCount: 1
+  csiProvisioner:
+    image:
+      repository: registry.k8s.io/sig-storage/csi-provisioner
```
TerryHowe, line 6: I think I'm the only one that does this, but I like to separate the registry from the repository so I can change the registry to a mirror and not change the repository structure.

charts/csi-snapshotter/templates/setup-csi-snapshotter.yaml
```diff
@@ -0,0 +1,105 @@
...
+      app.kubernetes.io/name: csi-snapshotter
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: csi-snapshotter
+    spec:
+      serviceAccountName: csi-snapshotter
+      containers:
+        - name: csi-provisioner
+          image: {{ .Values.csiSnapshotter.csiProvisioner.image.repository }}:{{ .Values.csiSnapshotter.csiProvisioner.image.tag }}
```
TerryHowe, line 59: I like to use a helper for this so you can allow installation by sha for higher security
```
{{/*
Create image name
*/}}          
{{- define "template.image" -}}
{{- if eq (substr 0 7 .tag) "sha256:" -}}
{{- printf "/%s@%s" .repository .tag -}} 
{{- else -}}
{{- printf "/%s:%s" .repository .tag -}}
{{- end -}}       
{{- end -}} 
```

That example expects registry to be set at a higher level

charts/csi-snapshotter/Chart.yaml
hswong3i: Sorry for my late reply.

As mentioned in this PR GIT commit message, we just need to update with:

> Before each stable tag release, maintainers only required to:
>     1. Run the `./deploy/util/chart-releaser.sh` and update templates under `charts/*/templates/*`
>     2. Double confirm `charts/*/values.yml` with correct values, e.g. image tags
>     3. `git add --all --force charts/* && git commit`

The update of this line will also be managed with the provided `./deploy/util/chart-releaser.sh`.

charts/csi-snapshotter/templates/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
hswong3i: From today point of view, even if I am using cert-manager + helm, during upgrade we also need some special handling (see https://cert-manager.io/docs/installation/upgrade/#crds-managed-using-helm):

> CRDs managed using helm
> 
> If you have installed the CRDs together with the helm install command (using --set crds.enabled=true), Helm will upgrade the CRDs automatically when you upgrade the cert-manager Helm chart:
> 
> helm upgrade --reset-then-reuse-values --version <version> <release_name> jetstack/cert-manager

So I don't think we could do something specially better than that :-(

charts/csi-snapshotter/values.yaml
hswong3i: Back to the origin: with the provided `./deploy/util/chart-releaser.sh`, we develop could still focus on original `deploy/kubernetes/csi-snapshotter` workflow, which additionally provide a simpler method for end-user daily deployment with helm.

As a long if we are switching into helm-first workflow style, but static deployment manifest as output result, this suggestion on the `repository:` line would be very meaningful.

IMHO, let's KISS for this PR as a good start with helm support, so we could improve it later?

charts/csi-snapshotter/templates/setup-csi-snapshotter.yaml
hswong3i: Similar as above reply: if we are working with helm-first model, yes this suggestion looks nice; but for now with static-manifest-first model, I would vote for no and KISS.

hswong3i:
Failed with reopen, recreate PR as #1325

Issue #1305, opened 2025-06-03
PVC finalizer not removed if invalid VolumeSnapshots with missing ReadyToUse field exist in the namespace


vasu81in:
**What happened**:

The snapshot.storage.kubernetes.io/pvc-as-source-protection finalizer is not removed from a PersistentVolumeClaim (PVC) if there are other VolumeSnapshot objects in the same namespace referencing that PVC that never reach status.readyToUse: true. This causes delete PVC to get stuck in a Terminating state even though the valid snapshots have completed and been deleted.

**What you expected to happen**:
The finalizer should be removed from the PVC once the only valid VolumeSnapshot referencing it reaches ReadyToUse: true and is no longer present. Invalid VolumeSnapshot objects (e.g., those with incorrect or non-existent VolumeSnapshotClass) that never transition to ReadyToUse should not block finalizer removal from the PVC.

**How to reproduce it**:

Create multiple VolumeSnapshot objects pointing to the same PVC, but use a non-existent  VolumeSnapshotClass for some of them. These will never transition to ReadyToUse: true.

Create a valid VolumeSnapshot referencing the same PVC using a working VolumeSnapshotClass.

Observe that the pvc-as-source-protection finalizer gets added to the PVC.

The valid VolumeSnapshot will complete (ReadyToUse: true), and may even be deleted.

However, the PVC’s finalizer is never removed, due to the lingering invalid snapshots.

Deleting the invalid VolumeSnapshots causes the finalizer to be cleaned up, only if done in the correct order (invalid → valid).

**Anything else we need to know?**:

The IsSnapshotReady() function checks if status.readyToUse is true, but when the field is unset (e.g. due to invalid VolumeSnapshotClass or a snapshot controller not updating the status), the snapshot is treated as still in use. As a result, finalizer removal is blocked, even if the snapshot never succeeded and never will.

if snap.Spec.Source.PersistentVolumeClaimName != nil &&
    pvc.Name == *snap.Spec.Source.PersistentVolumeClaimName &&
    !utils.IsSnapshotReady(snap) {
    return true
}

This considers any snapshot referencing the PVC, and if its ReadyToUse field is false or missing, it blocks finalizer removal even for broken or misconfigured snapshots.

**Environment**:
- Kubernetes version v1.32.2
- snapshot-controller: v6.3.3
- OS: any

xing-yang:
[...]

PR #1309, opened 2025-06-09
Fix: PVC finalizer not removed due to broken VolumeSnapshots

vasumahalingam-nutanix:
**What type of PR is this?**
> /kind bug

Fixes a bug where a PVC’s finalizer `snapshot.storage.kubernetes.io/pvc-as-source-protection` is not removed when there are broken or misconfigured VolumeSnapshots that reference the PVC but never transition to `ReadyToUse`.

The snapshot controller currently considers any VolumeSnapshot with `ReadyToUse == false` as blocking PVC deletion — even if the snapshot is invalid and has already failed.

This patch ensures the controller only considers:
- VolumeSnapshots with `ReadyToUse == false`, **and**
- that still retain the `pvc-as-source-protection` finalizer

This avoids permanently stuck PVCs when broken snapshots linger in the namespace.

Fixes [#1305](https://github.com/kubernetes-csi/external-snapshotter/issues/1305)

- The change preserves the protective behavior for in-progress snapshots.
- The PVC can now be deleted once only broken snapshots remain, and they no longer hold the finalizer.

```release-note
Fix: PVC finalizer `snapshot.storage.kubernetes.io/pvc-as-source-protection` is now removed even if other VolumeSnapshots are stuck in a failed state, as long as those snapshots no longer hold the finalizer.

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -1050,7 +1050,12 @@ func (ctrl *csiSnapshotCommonController) isPVCBeingUsed(pvc *v1.PersistentVolume
 			klog.V(4).Infof("Skipping static bound snapshot %s when checking PVC %s/%s", snap.Name, pvc.Namespace, pvc.Name)
 			continue
 		}
-		if snap.Spec.Source.PersistentVolumeClaimName != nil && pvc.Name == *snap.Spec.Source.PersistentVolumeClaimName && !utils.IsSnapshotReady(snap) {
+
+		if snap.Spec.Source.PersistentVolumeClaimName != nil &&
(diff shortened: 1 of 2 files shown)


vasu81in:
Hi @gnufied , @sunnylovestiramisu  PTAL when you get a chance. This PR resolves an edge case where PVC deletion is blocked due to broken VolumeSnapshots that are no longer usable or hold the finalizer. It ensures correctness without affecting valid in-progress snapshots. Thanks!

vasumahalingam-nutanix:
Hi @gnufied , @sunnylovestiramisu PTAL when you get a chance. This PR resolves an edge case where PVC deletion is blocked due to broken VolumeSnapshots that are no longer usable or hold the finalizer. It ensures correctness without affecting valid in-progress snapshots. Thanks!

PR #1312, opened 2025-06-25
Introduce v1beta2 VolumeGroupSnapshot API

leonardoce:
**What type of PR is this?**
>

/kind api-change

> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This commit introduces the new VolumeGroupSnapshot v1beta2 API, updates the CRD, the generated code, and introduce a conversion webhook allowing idempotent conversion between v1beta1 and v1beta2.

Both v1beta1 and v1beta2 APIs are served, with v1beta2 being the stored version.

Fixes #

NA

```release-note
Introduce the `v1beta2` VolumeGroupSnapshot API as described by [KEP 5013](https://github.com/kubernetes/enhancements/pull/5013)
```

### DIFF
--- a/client/apis/volumegroupsnapshot/v1beta2/types.go
+++ b/client/apis/volumegroupsnapshot/v1beta2/types.go
@@ -315,12 +315,11 @@ type VolumeGroupSnapshotContentSpec struct {
 // The VolumeSnapshotInfo struct is added in v1beta2
 // VolumeSnapshotInfo contains information for a snapshot
 type VolumeSnapshotInfo struct {
-	// VolumeHandle is a unique id returned by the CSI driver to identify a volume
-	// on the storage system
+	// VolumeHandle specifies the CSI "volume_id" of the volume from which this snapshot
+	// was taken from.
 	VolumeHandle string `json:"volumeHandle,omitempty" protobuf:"bytes,1,opt,name=volumeHandle"`
 
-	// SnapshotHandle is a unique id returned by the CSI driver to identify a volume
-	// snapshot on the storage system
+	// SnapshotHandle is the CSI "snapshot_id" of this snapshot on the underlying storage system.
 	SnapshotHandle string `json:"snapshotHandle,omitempty" protobuf:"bytes,2,opt,name=snapshotHandle"`
 
 	// creationTime is the timestamp when the point-in-time snapshot is taken
@@ -332,7 +331,8 @@ type VolumeSnapshotInfo struct {
 	// +optional
 	ReadyToUse *bool `json:"readyToUse,omitempty" protobuf:"varint,4,opt,name=readyToUse"`
 
-	// RestoreSize represents the complete size of the snapshot in bytes.
+	// RestoreSize represents the minimum size of volume required to create a volume
+	// from this snapshot.
 	// +optional
 	RestoreSize *int64 `json:"restoreSize,omitempty" protobuf:"bytes,5,opt,name=restoreSize"`
 }
--- a/vendor/github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta2/types.go
+++ b/vendor/github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta2/types.go
@@ -315,12 +315,11 @@ type VolumeGroupSnapshotContentSpec struct {
 // The VolumeSnapshotInfo struct is added in v1beta2
 // VolumeSnapshotInfo contains information for a snapshot
 type VolumeSnapshotInfo struct {
-	// VolumeHandle is a unique id returned by the CSI driver to identify a volume
-	// on the storage system
+	// VolumeHandle specifies the CSI "volume_id" of the volume from which this snapshot
+	// was taken from.
 	VolumeHandle string `json:"volumeHandle,omitempty" protobuf:"bytes,1,opt,name=volumeHandle"`
 
-	// SnapshotHandle is a unique id returned by the CSI driver to identify a volume
-	// snapshot on the storage system
+	// SnapshotHandle is the CSI "snapshot_id" of this snapshot on the underlying storage system.
 	SnapshotHandle string `json:"snapshotHandle,omitempty" protobuf:"bytes,2,opt,name=snapshotHandle"`
 
 	// creationTime is the timestamp when the point-in-time snapshot is taken
@@ -332,7 +331,8 @@ type VolumeSnapshotInfo struct {
 	// +optional
 	ReadyToUse *bool `json:"readyToUse,omitempty" protobuf:"varint,4,opt,name=readyToUse"`
 
-	// RestoreSize represents the complete size of the snapshot in bytes.
+	// RestoreSize represents the minimum size of volume required to create a volume
+	// from this snapshot.
 	// +optional
 	RestoreSize *int64 `json:"restoreSize,omitempty" protobuf:"bytes,5,opt,name=restoreSize"`
 }


client/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -0,0 +1,441 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+// +kubebuilder:object:generate=true
+package v1beta2
+
+import (
+	core_v1 "k8s.io/api/core/v1"
```
Madhu-1, line 20: can this be corev1?

leonardoce: Yes, and in other places, the same package is referred to as `v1`.
We should decide on a name and stick with it.

In this PR, I believe it is better to keep this file as similar as the sibling in `v1beta1` as possible.
I decided to update only the new files without touching the existing ones: this will make reviewing easier.

Shall we choose a name and handle it in a different PR?

Madhu-1: sounds good.

client/clientset/versioned/typed/volumegroupsnapshot/v1beta1/doc.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2024 The Kubernetes Authors.
+Copyright 2025 The Kubernetes Authors.
```
xing-yang: Why is this changed?  Same comment in other places.

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
```diff
@@ -92,3 +92,77 @@ spec:
...
+              Parameters is a key-value map with storage driver specific parameters for
+              creating group snapshots.
+              These values are opaque to Kubernetes and are passed directly to the driver.
+            type: object
+        required:
+        - deletionPolicy
+        - driver
+        type: object
+    served: true
+    storage: false
```
xing-yang: v1beta2 should have storage: true.
Same in other places.

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
```diff
@@ -92,3 +92,77 @@ spec:
     served: true
     storage: true
```
xing-yang: v1beta1 should have storage: false.

xing-yang:
Do you need to update README to generate files?
https://github.com/kubernetes-csi/external-snapshotter/tree/master/client/hack#readme

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
leonardoce: It should be definitely set to `true` when v1beta2 will become the "default" API, but in this moment we're internally not using it and we've no conversion webhook.
I was thinking to do it later, if you agree with me.

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
leonardoce: same as https://github.com/kubernetes-csi/external-snapshotter/pull/1312#discussion_r2187275506

client/clientset/versioned/typed/volumegroupsnapshot/v1beta1/doc.go
leonardoce: I guess this happened because I regenerated these files.
We could avoid putting this changes into the patch if you think it makes easier to review it.

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
xing-yang: Can you add that this will be changed in follow up PRs in the PR description?

client/clientset/versioned/typed/volumegroupsnapshot/v1beta1/doc.go
xing-yang: Right.  VolumeGroupSnapshot v1beta1 and VolumeSnapshot v1 related files should not be changed.

client/apis/volumesnapshot/v1/zz_generated.deepcopy.go
```diff
@@ -2,7 +2,7 @@
 // +build !ignore_autogenerated
 
 /*
-Copyright 2024 The Kubernetes Authors.
+Copyright 2025 The Kubernetes Authors.
```
xing-yang: This should not be updated.

client/clientset/versioned/typed/volumesnapshot/v1/doc.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2024 The Kubernetes Authors.
+Copyright 2025 The Kubernetes Authors.
```
xing-yang: This and other volumesnapshot v1 related files should not be updated.

README.md
```diff
@@ -66,6 +65,8 @@ Therefore, it is strongly recommended that Kubernetes distributors bundle and de
 
 If your Kubernetes distribution does not bundle the snapshot controller, you may manually install these components by executing the following steps. Note that the snapshot controller YAML files in the git repository deploy into the default namespace for system testing purposes. For general use, update the snapshot controller YAMLs with an appropriate namespace prior to installing. For example, on a Vanilla Kubernetes cluster update the namespace from 'default' to 'kube-system' prior to issuing the kubectl create command.
 
+There is a new validating webhook server which provides tightened validation on snapshot objects. The cluster admin or Kubernetes distribution admin should install the webhook alongside the snapshot controllers and CRDs. More details [below](#validating-webhook).
```
jsafrane: This is not a validation webhook.
```suggestion
There is a new conversion webhook server which provides conversion between v1beta1 and v1beta2 group snapshot objects. The cluster admin or Kubernetes distribution admin should install the webhook alongside the snapshot controllers and CRDs if they want to provide group snapshot v1beta1 API. More details [below](#validating-webhook).
```

README.md
```diff
@@ -82,6 +83,44 @@ Install CSI Driver:
   * With the repo cloned locally: `kubectl kustomize deploy/kubernetes/csi-snapshotter | kubectl create -f -`
   * From the repo remotely: `kubectl kustomize https://github.com/kubernetes-csi/external-snapshotter/deploy/kubernetes/csi-snapshotter | kubectl create -f -`
 
+### Validating Webhook
```
jsafrane: Conversion webhook
(the text below needs lot of updates)

README.md
```diff
@@ -242,6 +281,8 @@ If you have already deployed v1alpha1 snapshot APIs and external-snapshotter sid
 
 ### Upgrade from v1beta1 to v1
 
+Validation webhook should be installed before upgrading to v1. Potential impacts of not installing the validation webhook before upgrading to v1 include being unable to delete invalid snapshot objects. See the section on Validation Webhook for details.
```
jsafrane, line 245: This should be more like v1beta1 to v1beta2 upgrade. IMHO we need a new chapter. Or remove v1beta1 to v1 completely, it happened a long time ago.

deploy/kubernetes/webhook-example/README.md
```diff
@@ -0,0 +1,75 @@
+# Validating Webhook
```
jsafrane: Conversion webhook

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
```diff
@@ -17,6 +17,15 @@ spec:
...
   scope: Cluster
+  conversion:
+    strategy: Webhook
+    webhook:
+      conversionReviewVersions: ["v1","v1beta1"]
+      clientConfig:
+        service:
+          namespace: default
+          name: snapshot-webhook-service
+          path: /convert
```
jsafrane: If VGSClass v1beta1 is the same as v1beta2, I would not declare the conversion. The API server should do the right thing, without any code in the webhook

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshots.yaml
```diff
@@ -16,6 +16,15 @@ spec:
...
   scope: Namespaced
+  conversion:
+    strategy: Webhook
+    webhook:
+      conversionReviewVersions: ["v1","v1beta1"]
+      clientConfig:
+        service:
+          namespace: default
+          name: snapshot-webhook-service
+          path: /convert
```
jsafrane: If VGS v1beta1 is the same as v1beta2, I would not declare the conversion. The API server should do the right thing, without any code in the webhook

pkg/webhook/convert.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil, statusErrorWithMessage("conversion from a version to itself should not call the webhook: %s", toVersion)
+	}
+
+	switch obj.GetAPIVersion() {
+	case "groupsnapshot.storage.k8s.io/v1beta1":
+		switch toVersion {
+		case "groupsnapshot.storage.k8s.io/v1beta2":
+			switch kind {
+			case "VolumeGroupSnapshot":
+			case "VolumeGroupSnapshotClass":
```
jsafrane: Remove these cases, the webhook should not convert VolumeGroupSnapshots and VolumeGroupSnapshotClasses. Throw `unexpected conversion kind %q` below.

pkg/webhook/fuzzer.go
```diff
@@ -0,0 +1,43 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
jsafrane:
```suggestion
Copyright 2025 The Kubernetes Authors.
```

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,111 @@
...
+var (
+	certFile                         string
+	keyFile                          string
+	kubeconfigFile                   string
+	port                             int
+	preventVolumeModeConversion      bool
+	enableVolumeGroupSnapshotWebhook bool
+)
+
+var CmdWebhook = &cobra.Command{
```
jsafrane: should be most of this file be in cmd/webhook? The controller and the sidecar have their cmdline params defined in cmd/

pkg/webhook/testdata/v1beta1_to_v1beta2/annotation_no_status_v1beta1.yaml
```diff
@@ -0,0 +1,32 @@
+---
+apiVersion: groupsnapshot.storage.k8s.io/v1beta1
+kind: VolumeGroupSnapshot
```
jsafrane: Should the webhook convert VolumeGroupSnapshotContent instead???

pkg/webhook/testdata/v1beta1_to_v1beta2/annotation_no_status_v1beta2.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: groupsnapshot.storage.k8s.io/v1beta2
+kind: VolumeGroupSnapshot
```
jsafrane: This object is neither VolumeGroupSnapshot nor VolumeGroupSnapshotContent

.github/workflows/trivy.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
           make
           docker build -t test/csi-snapshotter:latest -f ./cmd/csi-snapshotter/Dockerfile --output=type=docker --label revision=latest .
           docker build -t test/snapshot-controller:latest -f ./cmd/snapshot-controller/Dockerfile --output=type=docker --label revision=latest .
+          docker build -t test/snapshot-webhook:latest -f ./cmd/snapshot-webhook/Dockerfile --output=type=docker --label revision=latest .
```
xing-yang: snapshot-conversion-webhook

README.md
```diff
@@ -82,6 +83,44 @@ Install CSI Driver:
   * With the repo cloned locally: `kubectl kustomize deploy/kubernetes/csi-snapshotter | kubectl create -f -`
   * From the repo remotely: `kubectl kustomize https://github.com/kubernetes-csi/external-snapshotter/deploy/kubernetes/csi-snapshotter | kubectl create -f -`
 
+### Validating Webhook
+
+The snapshot validating webhook is an HTTP callback which responds to [admission requests](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/). It is part of a larger [plan](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1900-volume-snapshot-validation-webhook) to tighten validation for volume snapshot objects. This webhook introduces the [ratcheting validation](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1900-volume-snapshot-validation-webhook#backwards-compatibility) mechanism targeting the tighter validation. The cluster admin or Kubernetes distribution admin should install the webhook alongside the snapshot controllers and CRDs.
```
xing-yang: snapshot conversion webhook

README.md
```diff
@@ -82,6 +83,44 @@ Install CSI Driver:
...
+
+Along with the validation webhook, the volume snapshot controller will start [labeling](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1900-volume-snapshot-validation-webhook#automatic-labelling-of-invalid-objects) invalid snapshot objects which already existed. This is to enable quick identification of invalid snapshot objects in the system by running:
```
xing-yang: It looks like this was copied from the readme for the old validation webhook, but the content does not work for the conversion webhook though.

README.md
```diff
@@ -82,6 +83,44 @@ Install CSI Driver:
...
+```
+kubectl get volumesnapshots --selector=snapshot.storage.kubernetes.io/invalid-snapshot-resource: ""
+kubectl get volumesnapshotcontents --selector=snapshot.storage.kubernetes.io/invalid-snapshot-content-resource: ""
```
xing-yang: I don't think we need these.

README.md
```diff
@@ -82,6 +83,44 @@ Install CSI Driver:
...
+
+* `--tls-cert-file`: File containing the x509 Certificate for HTTPS. (CA cert, if any, concatenated after server cert). Required.
+
+* `--tls-private-key-file`: File containing the x509 private key matching --tls-cert-file. Required.
+
+* `--port`: Secure port that the webhook listens on (default 443)
+
+* `--kubeconfig <path>`: Path to Kubernetes client configuration that the webhook uses to connect to Kubernetes API server. When omitted, default token provided by Kubernetes will be used. This option is useful only when the snapshot controller does not run as a Kubernetes pod, e.g. for debugging.
+
+* `--prevent-volume-mode-conversion`: Boolean that prevents an unauthorised user from modifying the volume mode when creating a PVC from an existing VolumeSnapshot. Was present as an alpha feature in `v6.0.0`; Having graduated to beta, defaults to true.
```
xing-yang: ?

README.md
xing-yang: Yes, we can remove the v1beta1 to v1 section for VolumeSnapshot API now.
You can add a new section for VolumeGroupSnapshot.

pkg/utils/conversion.go
```diff
@@ -0,0 +1,40 @@
...
+limitations under the License.
+*/
+
+package utils
+
+import "google.golang.org/protobuf/types/known/timestamppb"
+
+// CsiTimestampToKubernetes converts a CSI timestamp in the format
+// used by the Snapshot API
+func CsiTimestampToKubernetes(creationTime *timestamppb.Timestamp) *int64 {
```
jsafrane: Abbreviations are all caps, `CSITimestampToKubernetes`

leonardoce:
I realized I didn't had a look at the documentation. I'm looking at it now.

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,111 @@
...
+		"File containing the x509 private key matching --tls-cert-file. Required.")
+	CmdWebhook.Flags().IntVar(&port, "port", 443,
+		"Secure port that the webhook listens on")
+	CmdWebhook.MarkFlagRequired("tls-cert-file")
+	CmdWebhook.MarkFlagRequired("tls-private-key-file")
+	// Add optional flag for kubeconfig
+	CmdWebhook.Flags().StringVar(&kubeconfigFile, "kubeconfig", "", "kubeconfig file to use for volumesnapshotclasses")
+	CmdWebhook.Flags().BoolVar(&preventVolumeModeConversion, "prevent-volume-mode-conversion",
+		true, "Prevents an unauthorised user from modifying the volume mode when creating a PVC from an existing VolumeSnapshot.")
+	CmdWebhook.Flags().BoolVar(&enableVolumeGroupSnapshotWebhook, "enable-volume-group-snapshot-webhook",
```
jsafrane: This variable is never used. IMHO it does not make much sense to make the conversion optional - what will the binary do if it's disabled? Just exit?

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,111 @@
...
+		"File containing the x509 Certificate for HTTPS. (CA cert, if any, concatenated after server cert). Required.")
+	CmdWebhook.Flags().StringVar(&keyFile, "tls-private-key-file", "",
+		"File containing the x509 private key matching --tls-cert-file. Required.")
+	CmdWebhook.Flags().IntVar(&port, "port", 443,
+		"Secure port that the webhook listens on")
+	CmdWebhook.MarkFlagRequired("tls-cert-file")
+	CmdWebhook.MarkFlagRequired("tls-private-key-file")
+	// Add optional flag for kubeconfig
+	CmdWebhook.Flags().StringVar(&kubeconfigFile, "kubeconfig", "", "kubeconfig file to use for volumesnapshotclasses")
+	CmdWebhook.Flags().BoolVar(&preventVolumeModeConversion, "prevent-volume-mode-conversion",
```
jsafrane: this variable is never used.

deploy/kubernetes/webhook-example/patch-ca-bundle.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# File originally from https://github.com/banzaicloud/admission-webhook-example/blob/blog/deployment/webhook-patch-ca-bundle.sh
+
+ROOT=$(cd $(dirname $0)/../../; pwd)
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+CA_BUNDLE=$(kubectl config view --raw -o json | jq -r '.clusters[0].cluster."certificate-authority-data"' | tr -d '"')
```
jsafrane: This did not work for me, the API server rejected the webhook's key.
Since the key+certificate is generated by create-cert.sh, would it be better to use that?

This works, with hardcoded secret name:
```suggestion
CA_BUNDLE=$( kubectl get secret snapshot-webhook-secret -o json | jq -r '.data."tls.crt"' )
```

deploy/kubernetes/webhook-example/webhook.yaml
```diff
@@ -0,0 +1,50 @@
...
+    spec:
+      containers:
+      - name: snapshot-webhook
+        image: registry.k8s.io/sig-storage/snapshot-webhook:v8.0.1 # change the image if you wish to use your own custom validation server image
+        imagePullPolicy: IfNotPresent
+        args:
+        - '--tls-cert-file=/etc/snapshot-webhook/certs/tls.crt'
+        - '--tls-private-key-file=/etc/snapshot-webhook/certs/tls.key'
+        # uncomment the following line to enable webhook for VolumeGroupSnapshot, VolumeGroupSnapshotContent and VolumeGroupSnapshotClass.
+        # - '--enable-volume-group-snapshot-webhook'
```
jsafrane: The webhook seems to covert objects even without this parameter. IMHO it does not make much sense to have it optional - what will the webhook do if the conversion is disabled? Exit?

pkg/webhook/webhook.go
leonardoce: Agree. I reused the structure of the current validation webhook as far as possible, but having everything in `cmd/` is clearer in my opinion.

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotcontents.yaml
```diff
@@ -17,6 +17,15 @@ spec:
...
   scope: Cluster
+  conversion:
+    strategy: Webhook
+    webhook:
+      conversionReviewVersions: ["v1","v1beta1"]
+      clientConfig:
+        service:
+          namespace: default
+          name: snapshot-webhook-service
+          path: /convert
```
jsafrane: If someone deploys the CRD as it is, it will require the conversion webhook to run.
That complicates my local tests - I just want v1beta2 and I want to run it quickly, without generating TLS keys and whatnot. Would it be possible to make the conversion opt-in (or opt-out) via kustomize? And is it worth the effort?

deploy/kubernetes/webhook-example/create-cert.sh
```diff
@@ -0,0 +1,127 @@
...
+            shift
+            ;;
+        *)
+            usage
+            ;;
+    esac
+    shift
+done
+
+[ -z ${service} ] && service=admission-webhook-example-svc
```
jsafrane, line 44:
```suggestion
service=${service:-admission-webhook-example-svc}
```

deploy/kubernetes/webhook-example/create-cert.sh
```diff
@@ -0,0 +1,127 @@
...
+[ -z ${secret} ] && secret=admission-webhook-example-certs
+[ -z ${namespace} ] && namespace=default
+
+if [ ! -x "$(command -v openssl)" ]; then
+    echo "openssl not found"
+    exit 1
+fi
+
+csrName=${service}.${namespace}
+tmpdir=$(mktemp -d)
```
jsafrane, line 54: remove the dir on exit?

deploy/kubernetes/webhook-example/patch-ca-bundle.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# File originally from https://github.com/banzaicloud/admission-webhook-example/blob/blog/deployment/webhook-patch-ca-bundle.sh
```
jsafrane: https://github.com/banzaicloud/admission-webhook-example does not exist and thus I can't check what license the script had :-(

pkg/utils/conversion.go
```diff
@@ -0,0 +1,40 @@
...
+		return nil
+	}
+
+	result := creationTime.AsTime().UnixNano()
+	return &result
+}
+
+// CsiSizeToKubernetes converts the size to the format used by
+// the Snapshot API
+func CsiSizeToKubernetes(sizeBytes int64) *int64 {
```
jsafrane: can it be replaced by `ptr.To(sizeBytes)`?

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
leonardoce: Agree. Removed

pkg/webhook/convert.go
leonardoce: Removed.

pkg/webhook/fuzzer.go
leonardoce: Removed the file. It was used only in the old validation webhook.

pkg/webhook/testdata/v1beta1_to_v1beta2/annotation_no_status_v1beta1.yaml
leonardoce: Yes!

.github/workflows/trivy.yaml
leonardoce: Moved everything from snapshot-webhook to snapshot-conversion-webhook.

deploy/kubernetes/webhook-example/patch-ca-bundle.sh
leonardoce: Yes. That's because that shell script blindly uses the first cluster in the kubeconfig. I directly used the one we had in the old validation webhook.
We can directly use the public part of the generated secret, yes.

deploy/kubernetes/webhook-example/webhook.yaml
leonardoce: removed!

pkg/webhook/webhook.go
leonardoce: Removed!

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshots.yaml
leonardoce: Done.

pkg/webhook/framework.go
```diff
@@ -0,0 +1,247 @@
...
+			http.Error(w, msg, http.StatusBadRequest)
+			return
+		}
+		convertReview.Response = doConversionV1beta1(convertReview.Request, convert)
+		convertReview.Response.UID = convertReview.Request.UID
+		klog.V(2).Info(fmt.Sprintf("sending response: %v", convertReview.Response))
+
+		// reset the request, it is not needed in a response.
+		convertReview.Request = &v1beta1.ConversionRequest{}
+		responseObj = convertReview
```
jsafrane: `ConversionReview` is GA since 1.16, IMHO we don't need to implement v1beta1 handling and register v1beta1   `conversionReviewVersions` in the CRD

deploy/kubernetes/webhook-example/patch-ca-bundle.sh
leonardoce: Applied your suggestion. Thanks!

pkg/utils/conversion.go
leonardoce: I'm using a function here because I need `CSISizeToKubernetes(0)` to be `nil` instead of a pointer to zero.
This is needed because, in my understanding, at the GRPC level we use `0` to represent an empty restoreSize (it is declared optional) meanwhile in the API we need `nil`.

On second thought, I could have avoided creating this function and put the code where it is really needed. Would you prefer that? Described as it is, the behavior of this function could be surprising.

deploy/kubernetes/webhook-example/webhook.yaml
```diff
@@ -0,0 +1,49 @@
...
+        app.kubernetes.io/name: snapshot-conversion-webhook
+    spec:
+      containers:
+      - name: snapshot-conversion-webhook
+        image: registry.k8s.io/sig-storage/snapshot-conversion-webhook:v8.0.1 # change the image if you wish to use your own custom conversion server image
+        imagePullPolicy: IfNotPresent
+        args:
+        - '--tls-cert-file=/etc/snapshot-conversion-webhook/certs/tls.crt'
+        - '--tls-private-key-file=/etc/snapshot-conversion-webhook/certs/tls.key'
+        # uncomment the following line to enable webhook for VolumeGroupSnapshot, VolumeGroupSnapshotContent and VolumeGroupSnapshotClass.
```
Madhu-1: This comment need to be removed

cmd/snapshot-conversion-webhook/main.go
```diff
@@ -0,0 +1,81 @@
...
+	if certFile == nil || *certFile == "" {
+		klog.Fatal("--tls-cert-file must be specified")
+	}
+	if keyFile == nil || *keyFile == "" {
+		klog.Fatal("--tls-private-key-file must be specified")
+	}
+
+	// Create new cert watcher
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel() // stops certwatcher
```
Madhu-1: lets create the context when required, at line 76?

cmd/snapshot-conversion-webhook/main.go
```diff
@@ -0,0 +1,81 @@
...
+
+package main
+
+import (
+	"context"
+	"crypto/tls"
+	"flag"
+
+	"github.com/kubernetes-csi/csi-lib-utils/standardflags"
+	"github.com/kubernetes-csi/external-snapshotter/v8/pkg/webhook"
```
Madhu-1: move this to line 30 to have a separate grouping for local import?

deploy/kubernetes/webhook-example/webhook.yaml
```diff
@@ -0,0 +1,49 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: snapshot-conversion-webhook-deployment
+  namespace: default # NOTE: change the namespace
```
Madhu-1, line 6: use `kube-system` which is the namespace where we have the controller running, to keep it parity?

pkg/sidecar-controller/groupsnapshot_helper.go
```diff
@@ -30,7 +30,8 @@ import (
 	"k8s.io/client-go/tools/cache"
 	klog "k8s.io/klog/v2"
 
-	crdv1beta1 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta1"
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	crdv1beta2 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta2"
```
Madhu-1, line 34: a new line is required for import grouping

pkg/webhook/convert.go
```diff
@@ -0,0 +1,172 @@
...
+
+	convertedObject := obj.DeepCopy()
+	fromVersion := obj.GetAPIVersion()
+	kind := obj.GetKind()
+
+	if toVersion == fromVersion {
+		return nil, statusErrorWithMessage("conversion from a version to itself should not call the webhook: %s", toVersion)
+	}
+
+	switch obj.GetAPIVersion() {
```
Madhu-1: How about below version to avoid multiple switch statments

```go
v1beta1Version:"groupsnapshot.storage.k8s.io/v1beta1"
v1beta2Version:="groupsnapshot.storage.k8s.io/v1beta2"
if kind != "VolumeGroupSnapshotContent" {
  return nil, statusErrorWithMessage("unexpected conversion kind %q", kind)
}
switch {
case fromVersion==v1beta1Version && toVersion==v1beta2Version:
if err := convertVolumeGroupSnapshotContentFromV1beta1ToV1beta2(convertedObject); err != nil {
	return nil, statusErrorWithMessage("%s", err.Error())
}

case fromVersion==v1beta2Version && toVersion==v1beta1Version{
if err := convertVolumeGroupSnapshotContentFromV1beta2ToV1beta1(convertedObject); err != nil {
	return nil, statusErrorWithMessage("%s", err.Error())
}
default:
	return nil, statusErrorWithMessage("unexpected conversion from %q to %q", fromVersion, toVersion)
}
return convertedObject, statusSucceed()
```

pkg/webhook/convert.go
```diff
@@ -0,0 +1,172 @@
...
+		obj.SetAnnotations(annotations)
+
+		// Step 2: convert volumeSnapshotInfoList to volumeSnapshotHandlePairList
+		for i, entry := range volumeSnapshotInfoList {
+			if mapEntry, ok := entry.(map[string]interface{}); ok {
+				delete(mapEntry, "creationTime")
+				delete(mapEntry, "readyToUse")
+				delete(mapEntry, "restoreSize")
+			} else {
+				return fmt.Errorf("unexpected content in .status.volumeSnapshotInfoList[%q]: expected map", i)
```
Madhu-1: use %d instead of %q as i is int? need to return entry object?

pkg/webhook/webhook_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"math/big"
+	"net"
+	"os"
+	"testing"
+	"time"
+)
+
+func TestWebhookCertReload(t *testing.T) {
+	// Initialize test space
+	tmpDir := os.TempDir() + "/webhook-cert-tests"
```
Madhu-1: use `t.TempDir()` as it gets auto cleaned up when tests are done?

deploy/kubernetes/webhook-example/webhook.yaml
leonardoce: I'm not sure of the rationale of it. This comes from the old validating webhook.
Maybe we want to the user to not forget to change the namespace.

leonardoce:
> sample yamls files in client/hack/cel-tests need to be updated and the example yamls for the volumesnapshotgroup need to be updated as well.

Done!
I also removed a rule that was deprecated by https://github.com/kubernetes-csi/external-snapshotter/pull/1184

deploy/kubernetes/webhook-example/patch-ca-bundle.sh
xing-yang: This comment is not addressed yet.

deploy/kubernetes/webhook-example/webhook.yaml
```diff
@@ -0,0 +1,48 @@
...
+    matchLabels:
+      app.kubernetes.io/name: snapshot-conversion-webhook
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: snapshot-conversion-webhook
+    spec:
+      containers:
+      - name: snapshot-conversion-webhook
+        image: registry.k8s.io/sig-storage/snapshot-conversion-webhook:v8.0.1 # change the image if you wish to use your own custom conversion server image
```
xing-yang, line 21: We don't have this image tag for the snapshot-conversion-webhook yet: v8.0.1

xing-yang:
Can you add what tests you have run?

deploy/kubernetes/webhook-example/patch-ca-bundle.sh
leonardoce: Yes. But I'm not so sure on what we should do about it.
This comes from the old validation webhook, even if it has been heavily changed.
I'm removing it for now. Do you have a better idea?

deploy/kubernetes/webhook-example/webhook.yaml
leonardoce: Yes. Still I believe we should provide an image name for testing purposes.
Should we use the tag of the version where we'll release this? What `v9.0.0` be fine?

leonardoce:
> Can you add what tests you have run?

I run my own set of E2e tests that use the v1beta1 (old) API, covering:

* Volume Group Snapshots
    * Dynamic provisioning
        * DeletionPolicy = Delete
        * DeletionPolicy = Retain
    * Static provisioning
        * DeletionPolicy = Delete
        * DeletionPolicy = Retain
* Volume Snapshots
    * Dynamic provisioning
        * DeletionPolicy = Delete
        * DeletionPolicy = Retain
    * Static provisioning
        * DeletionPolicy = Delete
        * DeletionPolicy = Retain

I run them with both the ListSnapshots RPC call enabled and disabled.

This is an example of a dynamic provisioned VolumeGroupSnaphot:

```yaml
apiVersion: groupsnapshot.storage.k8s.io/v1beta2
kind: VolumeGroupSnapshot
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"groupsnapshot.storage.k8s.io/v1beta2","kind":"VolumeGroupSnapshot","metadata":{"annotations":{},"name":"new-groupsnapshot-demo","namespace":"default"},"spec":{"source":{"selector":{"matchLabels":{"cnpg.io/instanceName":"cluster-example-1"}}},"volumeGroupSnapshotClassName":"csi-hostpath-groupsnapclass"}}
  creationTimestamp: "2025-08-05T12:08:44Z"
  finalizers:
  - groupsnapshot.storage.kubernetes.io/volumegroupsnapshot-bound-protection
  generation: 1
  name: new-groupsnapshot-demo
  namespace: default
  resourceVersion: "4913"
  uid: 0b61d9aa-c3c3-4d40-ab12-8090f39ead9c
spec:
  source:
    selector:
      matchLabels:
        cnpg.io/instanceName: cluster-example-1
  volumeGroupSnapshotClassName: csi-hostpath-groupsnapclass
status:
  boundVolumeGroupSnapshotContentName: groupsnapcontent-0b61d9aa-c3c3-4d40-ab12-8090f39ead9c
  creationTime: "2025-08-05T12:08:44Z"
  readyToUse: true
```

And this is the relative VolumeGroupSnapshotContent as per the v1beta2 API.

```yaml
apiVersion: groupsnapshot.storage.k8s.io/v1beta2
kind: VolumeGroupSnapshotContent
metadata:
  creationTimestamp: "2025-08-05T12:08:44Z"
  finalizers:
  - groupsnapshot.storage.kubernetes.io/volumegroupsnapshotcontent-bound-protection
  generation: 1
  name: groupsnapcontent-0b61d9aa-c3c3-4d40-ab12-8090f39ead9c
  resourceVersion: "4880"
  uid: 3113c30e-59b1-4e01-bd26-0a458e0d0e90
spec:
  deletionPolicy: Delete
  driver: hostpath.csi.k8s.io
  source:
    volumeHandles:
    - f49f1cb2-71f2-11f0-b3d2-5258f6f0f310
    - f49f025c-71f2-11f0-b3d2-5258f6f0f310
  volumeGroupSnapshotClassName: csi-hostpath-groupsnapclass
  volumeGroupSnapshotRef:
    apiVersion: groupsnapshot.storage.k8s.io/v1beta2
    kind: VolumeGroupSnapshot
    name: new-groupsnapshot-demo
    namespace: default
    resourceVersion: "4870"
    uid: 0b61d9aa-c3c3-4d40-ab12-8090f39ead9c
status:
  creationTime: "2025-08-05T12:08:44Z"
  readyToUse: true
  volumeGroupSnapshotHandle: ef337d91-71f4-11f0-b3d2-5258f6f0f310
  volumeSnapshotInfoList:
  - creationTime: 1754395724757339958
    readyToUse: true
    restoreSize: 1073741824
    snapshotHandle: ef337da2-71f4-11f0-b3d2-5258f6f0f310
    volumeHandle: f49f1cb2-71f2-11f0-b3d2-5258f6f0f310
  - creationTime: 1754395724757339958
    readyToUse: true
    restoreSize: 1073741824
    snapshotHandle: ef6c9521-71f4-11f0-b3d2-5258f6f0f310
    volumeHandle: f49f025c-71f2-11f0-b3d2-5258f6f0f310
```

This object, when converted to the v1beta1 API, looks like this:

```yaml
% kubectl get volumegroupsnapshotcontent.v1beta1.groupsnapshot.storage.k8s.io groupsnapcontent-0b61d9aa-c3c3-4d40-ab12-8090f39ead9c -o yaml
apiVersion: groupsnapshot.storage.k8s.io/v1beta1
kind: VolumeGroupSnapshotContent
metadata:
  annotations:
    groupsnapshot.storage.kubernetes.io/volume-snapshot-info-list: '[{"creationTime":1754395724757339958,"readyToUse":true,"restoreSize":1073741824,"snapshotHandle":"ef337da2-71f4-11f0-b3d2-5258f6f0f310","volumeHandle":"f49f1cb2-71f2-11f0-b3d2-5258f6f0f310"},{"creationTime":1754395724757339958,"readyToUse":true,"restoreSize":1073741824,"snapshotHandle":"ef6c9521-71f4-11f0-b3d2-5258f6f0f310","volumeHandle":"f49f025c-71f2-11f0-b3d2-5258f6f0f310"}]'
  creationTimestamp: "2025-08-05T12:08:44Z"
  finalizers:
  - groupsnapshot.storage.kubernetes.io/volumegroupsnapshotcontent-bound-protection
  generation: 1
  name: groupsnapcontent-0b61d9aa-c3c3-4d40-ab12-8090f39ead9c
  resourceVersion: "4880"
  uid: 3113c30e-59b1-4e01-bd26-0a458e0d0e90
spec:
  deletionPolicy: Delete
  driver: hostpath.csi.k8s.io
  source:
    volumeHandles:
    - f49f1cb2-71f2-11f0-b3d2-5258f6f0f310
    - f49f025c-71f2-11f0-b3d2-5258f6f0f310
  volumeGroupSnapshotClassName: csi-hostpath-groupsnapclass
  volumeGroupSnapshotRef:
    apiVersion: groupsnapshot.storage.k8s.io/v1beta2
    kind: VolumeGroupSnapshot
    name: new-groupsnapshot-demo
    namespace: default
    resourceVersion: "4870"
    uid: 0b61d9aa-c3c3-4d40-ab12-8090f39ead9c
status:
  creationTime: "2025-08-05T12:08:44Z"
  readyToUse: true
  volumeGroupSnapshotHandle: ef337d91-71f4-11f0-b3d2-5258f6f0f310
  volumeSnapshotHandlePairList:
  - snapshotHandle: ef337da2-71f4-11f0-b3d2-5258f6f0f310
    volumeHandle: f49f1cb2-71f2-11f0-b3d2-5258f6f0f310
  - snapshotHandle: ef6c9521-71f4-11f0-b3d2-5258f6f0f310
    volumeHandle: f49f025c-71f2-11f0-b3d2-5258f6f0f310
```

xing-yang:
Can you show an example of a VolumeSnapshot and VolumeSnapshotContent that are part of a group snapshot?

leonardoce:
Sure!

```yaml
% kubectl get volumesnapshot snapshot-67917c951a1b246f664709d3f3b5ac1d5d11b0371047835ae91ae9b78378c44f -o yaml
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  creationTimestamp: "2025-08-05T12:08:45Z"
  finalizers:
  - snapshot.storage.kubernetes.io/volumesnapshot-in-group-protection
  - snapshot.storage.kubernetes.io/volumesnapshot-as-source-protection
  generation: 1
  name: snapshot-67917c951a1b246f664709d3f3b5ac1d5d11b0371047835ae91ae9b78378c44f
  namespace: default
  ownerReferences:
  - apiVersion: groupsnapshot.storage.k8s.io/v1beta2
    kind: VolumeGroupSnapshot
    name: new-groupsnapshot-demo
    uid: 0b61d9aa-c3c3-4d40-ab12-8090f39ead9c
  resourceVersion: "4898"
  uid: 16d90945-465a-4d16-99e1-bbe2e0d99836
spec:
  source:
    persistentVolumeClaimName: cluster-example-1
status:
  boundVolumeSnapshotContentName: snapcontent-67917c951a1b246f664709d3f3b5ac1d5d11b0371047835ae91ae9b78378c44f
  creationTime: "2025-08-05T12:08:44Z"
  readyToUse: true
  restoreSize: 1Gi
  volumeGroupSnapshotName: new-groupsnapshot-demo
```

```yaml
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshotContent
metadata:
  annotations:
    groupsnapshot.storage.k8s.io/volumeGroupSnapshotHandle: ef337d91-71f4-11f0-b3d2-5258f6f0f310
  creationTimestamp: "2025-08-05T12:08:45Z"
  finalizers:
  - snapshot.storage.kubernetes.io/volumesnapshotcontent-bound-protection
  generation: 2
  name: snapcontent-67917c951a1b246f664709d3f3b5ac1d5d11b0371047835ae91ae9b78378c44f
  resourceVersion: "4887"
  uid: 7346342b-8635-4794-b566-2d245af6ac47
spec:
  deletionPolicy: Delete
  driver: hostpath.csi.k8s.io
  source:
    volumeHandle: f49f1cb2-71f2-11f0-b3d2-5258f6f0f310
  sourceVolumeMode: Filesystem
  volumeSnapshotRef:
    kind: VolumeSnapshot
    name: snapshot-67917c951a1b246f664709d3f3b5ac1d5d11b0371047835ae91ae9b78378c44f
    namespace: default
    uid: 16d90945-465a-4d16-99e1-bbe2e0d99836
status:
  creationTime: 1754395724757339958
  readyToUse: true
  restoreSize: 1073741824
  snapshotHandle: ef337da2-71f4-11f0-b3d2-5258f6f0f310
  volumeGroupSnapshotHandle: ef337d91-71f4-11f0-b3d2-5258f6f0f310
```

deploy/kubernetes/webhook-example/webhook.yaml
xing-yang: v9.0.0 does not exist yet. Okay, just leave it as is.

deploy/kubernetes/webhook-example/patch-ca-bundle.sh
jsafrane: I am looking at the old external-snapshotter ([v3](https://github.com/kubernetes-csi/external-snapshotter/blob/v3.0.0/deploy/kubernetes/webhook-example/patch-ca-bundle.sh)) that had this comment and it looks like a very different script now. Did anything actually remain from the original banzaicloud webhook-patch-ca-bundle.sh?

jsafrane: I can see the comment is gone now.

vendor/github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -0,0 +1,441 @@
...
+	// Required.
+	Source VolumeGroupSnapshotContentSource `json:"source" protobuf:"bytes,5,opt,name=source"`
+}
+
+// The VolumeSnapshotInfo struct is added in v1beta2
+// VolumeSnapshotInfo contains information for a snapshot
+type VolumeSnapshotInfo struct {
+	// VolumeHandle is a unique id returned by the CSI driver to identify a volume
+	// on the storage system
+	VolumeHandle string `json:"volumeHandle,omitempty" protobuf:"bytes,1,opt,name=volumeHandle"`
```
jsafrane:
```suggestion
	// VolumeHandle specifies the CSI "volume_id" of the volume from which this snapshot
	// was taken from.
	VolumeHandle string `json:"volumeHandle,omitempty" protobuf:"bytes,1,opt,name=volumeHandle"`
```

vendor/github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -0,0 +1,441 @@
...
+
+	// SnapshotHandle is a unique id returned by the CSI driver to identify a volume
+	// snapshot on the storage system
+	SnapshotHandle string `json:"snapshotHandle,omitempty" protobuf:"bytes,2,opt,name=snapshotHandle"`
```
jsafrane:
```suggestion
	// SnapshotHandle is the CSI "snapshot_id" of this snapshot on the underlying storage system.
	SnapshotHandle string `json:"snapshotHandle,omitempty" protobuf:"bytes,2,opt,name=snapshotHandle"`
```

vendor/github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -0,0 +1,441 @@
...
+	// +optional
+	CreationTime *int64 `json:"creationTime,omitempty" protobuf:"varint,3,opt,name=creationTime"`
+
+	// ReadyToUse indicates if the snapshot is ready to be used to restore a volume.
+	// +optional
+	ReadyToUse *bool `json:"readyToUse,omitempty" protobuf:"varint,4,opt,name=readyToUse"`
+
+	// RestoreSize represents the complete size of the snapshot in bytes.
+	// +optional
+	RestoreSize *int64 `json:"restoreSize,omitempty" protobuf:"bytes,5,opt,name=restoreSize"`
```
jsafrane:
```suggestion
	// RestoreSize represents the minimum size of volume required to create a volume
	// from this snapshot.
	// +optional
	RestoreSize *int64 `json:"restoreSize,omitempty" protobuf:"bytes,5,opt,name=restoreSize"`
```

xing-yang:
Trivy failure is unrelated to this change.

client/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -0,0 +1,441 @@
...
+type VolumeGroupSnapshotStatus struct {
+	// BoundVolumeGroupSnapshotContentName is the name of the VolumeGroupSnapshotContent
+	// object to which this VolumeGroupSnapshot object intends to bind to.
+	// If not specified, it indicates that the VolumeGroupSnapshot object has not
+	// been successfully bound to a VolumeGroupSnapshotContent object yet.
+	// NOTE: To avoid possible security issues, consumers must verify binding between
+	// VolumeGroupSnapshot and VolumeGroupSnapshotContent objects is successful
+	// (by validating that both VolumeGroupSnapshot and VolumeGroupSnapshotContent
+	// point at each other) before using this object.
+	// +optional
```
msau42, line 83: Should this be immutable once set?

client/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -0,0 +1,441 @@
...
+	BoundVolumeGroupSnapshotContentName *string `json:"boundVolumeGroupSnapshotContentName,omitempty" protobuf:"bytes,1,opt,name=boundVolumeGroupSnapshotContentName"`
+
+	// CreationTime is the timestamp when the point-in-time group snapshot is taken
+	// by the underlying storage system.
+	// If not specified, it may indicate that the creation time of the group snapshot
+	// is unknown.
+	// The format of this field is a Unix nanoseconds time encoded as an int64.
```
msau42, line 90: This doesn't seem accurate since the type is a metav1.Time?

client/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -0,0 +1,441 @@
...
+// uses when creating a volume group snapshot. A specific VolumeGroupSnapshotClass
+// is used by specifying its name in a VolumeGroupSnapshot object.
+// VolumeGroupSnapshotClasses are non-namespaced.
+// +kubebuilder:object:root=true
+// +kubebuilder:storageversion
+// +kubebuilder:resource:scope=Cluster,shortName=vgsclass;vgsclasses
+// +kubebuilder:printcolumn:name="Driver",type=string,JSONPath=`.driver`
+// +kubebuilder:printcolumn:name="DeletionPolicy",type=string,JSONPath=`.deletionPolicy`,description="Determines whether a VolumeGroupSnapshotContent created through the VolumeGroupSnapshotClass should be deleted when its bound VolumeGroupSnapshot is deleted."
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+type VolumeGroupSnapshotClass struct {
```
msau42, line 171: For each of the fields here, are they mutable?

client/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -0,0 +1,441 @@
...
+
+// VolumeGroupSnapshotContentSpec describes the common attributes of a group snapshot content
+type VolumeGroupSnapshotContentSpec struct {
+	// VolumeGroupSnapshotRef specifies the VolumeGroupSnapshot object to which this
+	// VolumeGroupSnapshotContent object is bound.
+	// VolumeGroupSnapshot.Spec.VolumeGroupSnapshotContentName field must reference to
+	// this VolumeGroupSnapshotContent's name for the bidirectional binding to be valid.
+	// For a pre-existing VolumeGroupSnapshotContent object, name and namespace of the
+	// VolumeGroupSnapshot object MUST be provided for binding to happen.
+	// This field is immutable after creation.
```
msau42, line 270: Is there enforcement of all the immutable fields?

client/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -0,0 +1,441 @@
...
+	// For pre-existing snapshots, users MUST specify this field when creating the
+	// VolumeGroupSnapshotContent object.
+	// Required.
+	DeletionPolicy snapshotv1.DeletionPolicy `json:"deletionPolicy" protobuf:"bytes,2,opt,name=deletionPolicy"`
+
+	// Driver is the name of the CSI driver used to create the physical group snapshot on
+	// the underlying storage system.
+	// This MUST be the same as the name returned by the CSI GetPluginName() call for
+	// that driver.
+	// Required.
```
msau42, line 295: Should this be immutable?

client/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -0,0 +1,441 @@
...
+	Driver string `json:"driver" protobuf:"bytes,3,opt,name=driver"`
+
+	// VolumeGroupSnapshotClassName is the name of the VolumeGroupSnapshotClass from
+	// which this group snapshot was (or will be) created.
+	// Note that after provisioning, the VolumeGroupSnapshotClass may be deleted or
+	// recreated with different set of values, and as such, should not be referenced
+	// post-snapshot creation.
+	// For dynamic provisioning, this field must be set.
+	// This field may be unset for pre-provisioned snapshots.
+	// +optional
```
msau42, line 305: Should this be immutable?

client/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -0,0 +1,441 @@
...
+	RestoreSize *int64 `json:"restoreSize,omitempty" protobuf:"bytes,5,opt,name=restoreSize"`
+}
+
+// VolumeGroupSnapshotContentStatus defines the observed state of VolumeGroupSnapshotContent.
+type VolumeGroupSnapshotContentStatus struct {
+	// VolumeGroupSnapshotHandle is a unique id returned by the CSI driver
+	// to identify the VolumeGroupSnapshot on the storage system.
+	// If a storage system does not provide such an id, the
+	// CSI driver can choose to return the VolumeGroupSnapshot name.
+	// +optional
```
msau42, line 346: Should this be immutable?

client/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -0,0 +1,441 @@
...
+	VolumeGroupSnapshotHandle *string `json:"volumeGroupSnapshotHandle,omitempty" protobuf:"bytes,1,opt,name=volumeGroupSnapshotHandle"`
+
+	// CreationTime is the timestamp when the point-in-time group snapshot is taken
+	// by the underlying storage system.
+	// If not specified, it indicates the creation time is unknown.
+	// If not specified, it means the readiness of a group snapshot is unknown.
+	// The format of this field is a Unix nanoseconds time encoded as an int64.
```
msau42, line 353: format is not int64?

client/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -0,0 +1,441 @@
...
+	// VolumeSnapshotHandlePairList is a list of CSI "volume_id" and "snapshot_id"
+	// pair returned by the CSI driver to identify snapshots and their source volumes
+	// on the storage system.
+	// +optional
+	// VolumeSnapshotHandlePairList []VolumeSnapshotHandlePair `json:"volumeSnapshotHandlePairList,omitempty" protobuf:"bytes,6,opt,name=volumeSnapshotHandlePairList"`
+
+	// This field is introduced in v1beta2
+	// It is replacing VolumeSnapshotHandlePairList
+	// VolumeSnapshotInfoList is a list of snapshot information returned by
+	// by the CSI driver to identify snapshots on the storage system.
```
msau42, line 385: Can the number of snapshots in the list change?

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
```diff
@@ -90,5 +90,79 @@ spec:
         - driver
         type: object
     served: true
+    storage: false
```
msau42, line 93: should we deprecate v1beta1? https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-deprecation

client/apis/volumegroupsnapshot/v1beta2/types.go
xing-yang: Yes

client/apis/volumegroupsnapshot/v1beta2/types.go
xing-yang: Good catch.  We forgot the update the comments after changing the type.

client/apis/volumegroupsnapshot/v1beta2/types.go
xing-yang: Fields in VolumeGroupSnapshotClass are immutable.

client/apis/volumegroupsnapshot/v1beta2/types.go
xing-yang: Yes.

client/apis/volumegroupsnapshot/v1beta2/types.go
xing-yang: Yes

client/apis/volumegroupsnapshot/v1beta2/types.go
xing-yang: Yes

client/apis/volumegroupsnapshot/v1beta2/types.go
xing-yang: No, we changed this to metav1.Time, but forgot the update the comments.

client/apis/volumegroupsnapshot/v1beta2/types.go
xing-yang: No.  Once taken, it can't be changed.

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
xing-yang: Yes, we should do that.

client/apis/volumegroupsnapshot/v1beta2/types.go
xing-yang: We have enforcement for some immutable fields.  Let me go through them and see anything is missing.

PR #1331, opened 2025-09-05
WIP: Enable the VolumeGroupSnapshot tests

Phaow:
**What type of PR is this?**
 /kind feature

- Enable the VolumeGroupSnapshot tests

Fixes #

```release-note
NONE
```

### DIFF
--- a/release-tools/prow.sh
+++ b/release-tools/prow.sh
@@ -381,7 +381,9 @@ default_csi_snapshotter_version () {
 		echo "v4.0.0"
 	fi
 }
-export CSI_SNAPSHOTTER_HACK_VERSION="master"
+# export CSI_SNAPSHOTTER_HACK_VERSION="master"
+# Hack to be consistent with 1.33 job
+export CSI_SNAPSHOTTER_HACK_VERSION="v8.2.0"
 configvar CSI_SNAPSHOTTER_VERSION "$(default_csi_snapshotter_version)" "external-snapshotter version tag"
 
 # Enable installing VolumeGroupSnapshot CRDs (off by default, can be set to true in prow jobs)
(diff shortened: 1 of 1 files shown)


release-tools/prow.sh
```diff
@@ -37,6 +37,8 @@
 # - kind (https://github.com/kubernetes-sigs/kind) installed
 # - optional: Go already installed
 
+set -x
```
xing-yang, line 40: We can't make changes under release-tools in the sidecar repo directly.

pkg/features/features.go
```diff
@@ -39,6 +39,6 @@ func init() {
 // defaultKubernetesFeatureGates consists of all known feature keys specific to external-snapshotter.
 // To add a new feature, define a key for it above and add it here.
 var defaultKubernetesFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
-	VolumeGroupSnapshot:         {Default: false, PreRelease: featuregate.Beta},
+	VolumeGroupSnapshot:         {Default: true, PreRelease: featuregate.Beta},
```
xing-yang: VolumeGroupSnapshot is still a Beta API.  It needs to be false by default.

Phaow: @xing-yang Thank you for reviewing and look into this! Yeah, I just want to make some hack changes to run the vgs tests, I'll add a new job enbale the feature gate to run vgs tests in presubmit CI after my research. Sorry I forgot to make it as draft.

release-tools/prow.sh
Phaow: Sorry I am not familiar with this part, so does it mean I need to make changes in https://github.com/kubernetes-csi/csi-release-tools？ Will go through the detail docs, thanks for the info!

Phaow:
Will sync the release-tools back after https://github.com/kubernetes-csi/csi-release-tools/pull/283 megred.

PR #1334, opened 2025-09-23
WIP: Change codes.ResourceExhausted from non-final to final error

xing-yang:
**What type of PR is this?**
/kind cleanup

This PR changes codes.ResourceExhausted from non-final to final error.

Fixes #

Resource_exhausted was initially treated as a non-final error in external-provisioner as well: https://github.com/kubernetes-csi/external-provisioner/pull/312/files.
Later on, it got changed to be a final error: https://github.com/kubernetes-csi/external-provisioner/pull/405/files#diff-243ebed2765f75e6a54f57167212fefb08c3b2a85967ad2acbc0eb78919019c1.
I think we can make a similar change in external-snapshotter. Resource_exhausted due to out of capacity is not a problem that can be fixed quickly. It usually results in infinitely retries if it is treated as a non-final error.

There's a slack thread about this issue: https://kubernetes.slack.com/archives/C09QZFCE5/p1688119656274719

```release-note
Change codes.ResourceExhausted from non-final to final error
```

### DIFF
--- a/pkg/sidecar-controller/snapshot_controller.go
+++ b/pkg/sidecar-controller/snapshot_controller.go
@@ -744,12 +744,19 @@ func isCSIFinalError(err error) bool {
 	}
 	switch st.Code() {
 	case codes.Canceled, // gRPC: Client Application cancelled the request
-		codes.DeadlineExceeded,  // gRPC: Timeout
-		codes.Unavailable,       // gRPC: Server shutting down, TCP connection broken - previous CreateSnapshot() may be still in progress.
-		codes.ResourceExhausted, // gRPC: Server temporarily out of resources - previous CreateSnapshot() may be still in progress.
-		codes.Aborted:           // CSI: Operation pending for Snapshot
+		codes.DeadlineExceeded, // gRPC: Timeout
+		codes.Unavailable,      // gRPC: Server shutting down, TCP connection broken - previous CreateSnapshot() may be still in progress.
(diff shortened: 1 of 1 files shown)


tiagolobocastro:
/reopen

I think this would still be useful to match external-provisioner

yugchaudhari:
@xing-yang Would you be open to reopening this PR? We're actively hitting this issue in production with OpenEBS
Mayastor, `ResourceExhausted` (507 Insufficient Storage) causes infinite retry loops when a pool is full, even
though the error is permanent and cannot resolve without manual intervention.

As you noted in the code comment, `external-provisioner` already treats `ResourceExhausted` as a final error
([PR #405](https://github.com/kubernetes-csi/external-provisioner/pull/405)). Making `external-snapshotter`
consistent would fix a real pain point for CSI drivers that return capacity errors during snapshot/clone
operations.

Re: the CSI spec ambiguity (https://github.com/container-storage-interface/spec/issues/419): I think your
reasoning in the comment is sound: transport-layer "message size exceeded" errors are unlikely in practice for
snapshot/volume operations. The pragmatic choice is to treat it as final, matching what `external-provisioner`
already does.

yugchaudhari:
@xing-yang can we reopen/revive this please?

yugchaudhari:
Picked this up in a fresh PR since the bot closed this one and reopen needs author or collaborator permissions: https://github.com/kubernetes-csi/external-snapshotter/pull/1435

PR #1337, opened 2025-10-07
Deprecate VolumeGroupSnapshot v1beta1, address API review comments

leonardoce:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Addresses API review comments in #1312

Fixes #

```release-note
Several VolumeGroupSnapshot, VolumeGroupSnapshotClass and VolumeGroupSnapshotContent fields are now immutable. The `v1beta1` VolumeGroupSnapshot API is now marked as deprecated.
```

### DIFF
--- a/client/hack/cel-tests/volumegroupsnapshotcontent/vgsc-class-immutable.post.yaml
+++ b/client/hack/cel-tests/volumegroupsnapshotcontent/vgsc-class-immutable.post.yaml
@@ -1,0 +1,18 @@
+---
+apiVersion: groupsnapshot.storage.k8s.io/v1beta2
+kind: VolumeGroupSnapshotContent
+metadata:
+  name: new-groupsnapshotcontent-demo
+spec:
+  volumeGroupSnapshotRef:
+    name: new-groupsnapshot-demo
+    namespace: default
+  driver: hostpath.csi.k8s.io
+  volumeGroupSnapshotClassName: class-post
+  source:
+    groupSnapshotHandles:
+      volumeGroupSnapshotHandle: this-handle
+      volumeSnapshotHandles:
+      - handle
+      - another-handle
+  deletionPolicy: Retain
--- a/client/hack/README.md
+++ b/client/hack/README.md
@@ -121,18 +121,9 @@ are correctly working.
 ```
   ./hack/run-cel-tests.sh
 
-  cel-tests/volumegroupsnapshotcontent/vgsc-change-ref-namespace.post.yaml: SUCCESS
-  cel-tests/volumegroupsnapshotcontent/vgsc-source-volume-to-groupsnapshot.post.yaml: SUCCESS
-  cel-tests/volumegroupsnapshotcontent/vgsc-source-empty.yaml: SUCCESS (expected failure)
-  cel-tests/volumegroupsnapshotcontent/vgsc-change-ref-namespace.pre.yaml: SUCCESS
-  cel-tests/volumegroupsnapshotcontent/vgsc-ref-only-name.yaml: SUCCESS (expected failure)
-  [...]
-  cel-tests/volumegroupsnapshotcontent/vgsc-change-ref-namespace.pre.yaml -> cel-tests/volumegroupsnapshotcontent/vgsc-change-ref-namespace.post.yaml: SUCCESS (expected failure)
-  cel-tests/volumegroupsnapshotcontent/vgsc-source-volume-immutable.pre.yaml -> cel-tests/volumegroupsnapshotcontent/vgsc-source-volume-immutable.post.yaml: SUCCESS (expected failure)
-  cel-tests/volumegroupsnapshotcontent/vgsc-source-volume-to-groupsnapshot.pre.yaml -> cel-tests/volumegroupsnapshotcontent/vgsc-source-volume-to-groupsnapshot.post.yaml: SUCCESS (expected failure)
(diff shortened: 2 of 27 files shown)


client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
```diff
@@ -91,6 +91,7 @@ spec:
         type: object
     served: true
     storage: false
+    deprecated: true
```
xing-yang: Is this added manually?  If so, we need to document this step in README.

xing-yang:
For VolumeGroupSnapshotRef in VolumeGroupSnapshotContentSpec, we can't make VolumeGroupSnapshotRef itself immutable because name and namespace are set by the admin and UID is set by the controller for static provisioning. 
Can we add rule to make sure name, namespace, or UID is immutable once set?

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
leonardoce: I used the `deprecatedversion` for that now.

leonardoce:
> For VolumeGroupSnapshotRef in VolumeGroupSnapshotContentSpec, we can't make VolumeGroupSnapshotRef itself immutable because name and namespace are set by the admin and UID is set by the controller for static provisioning. Can we add rule to make sure name, namespace, or UID is immutable once set?

I added a CEL rule for that.

client/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -194,6 +194,7 @@ type VolumeGroupSnapshotClass struct {
 	// "Delete" means that the VolumeGroupSnapshotContent and its physical group
 	// snapshot on underlying storage system are deleted.
 	// Required.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="deletionPolicy is immutable once set"
```
msau42, line 197: I think we do allow users to change the deletion policy for SC.

client/apis/volumegroupsnapshot/v1beta2/types.go
```diff
@@ -80,16 +80,14 @@ type VolumeGroupSnapshotStatus struct {
 	// VolumeGroupSnapshot and VolumeGroupSnapshotContent objects is successful
 	// (by validating that both VolumeGroupSnapshot and VolumeGroupSnapshotContent
 	// point at each other) before using this object.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="boundVolumeGroupSnapshotContentName is immutable once set"
```
msau42, line 83: I'm not actually sure, but does this handle the scenario where we need to set it the first time?

msau42:
Do we have tests running for the groupsnapshot functionality?

client/apis/volumegroupsnapshot/v1beta2/types.go
xing-yang: Do we allow that?  I didn't know we can modify SC at all.
I thought it is the ReclaimPolicy in PV that we can change.

client/apis/volumegroupsnapshot/v1beta2/types.go
xing-yang: It does, it is immutable after it is set.

xing-yang:
> Do we have tests running for the groupsnapshot functionality?

Yes, we added e2e test before moving group snapshot to v1beta1.

xing-yang:
@leonardoce can you add cel tests for changes you made in this PR?  https://github.com/kubernetes-csi/external-snapshotter/tree/master/client/hack/cel-tests

client/apis/volumegroupsnapshot/v1beta2/types.go
msau42: hm ok I do see it is immutable for SC: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/storage/validation/validation.go#L79

But do we actually want that behavior?

xing-yang: Yes, we don't want to change deletion policy of a VolumeGroupSnapshotClass.
We only want to change deletion policy of a VolumeGroupSnapshotContent.

leonardoce:
> can you add cel tests for changes you made in this PR? https://github.com/kubernetes-csi/external-snapshotter/tree/master/client/hack/cel-tests

I added the test cases. Thanks!

client/hack/cel-tests/volumegroupsnapshotcontent/vgsc-class-immutable.post.yaml
```diff
@@ -0,0 +1,18 @@
...
+    namespace: default
+  driver: hostpath.csi.k8s.io
+  volumeGroupSnapshotClassName: class-post
+  source:
+    groupSnapshotHandles:
+      volumeGroupSnapshotHandle: this-handle
+      volumeSnapshotHandles:
+      - handle
+      - another-handle
+  deletionPolicy: Retain
```
xing-yang, line 18: Do you have a positive test case for deletionPolicy in VolumeGroupSnapshotContent?

leonardoce: I just added one

PR #1354, opened 2025-12-10
Fix for cases where the resource is already present on the cluster

### LINKED ISSUE #1280: Group snapshot not taken on conflict
**What happened**:

I do not know how exactly it happened, but I got into a situation when the snapshot controller was processing a VolumeGroupSnapshot and corresponding VolumeSnapshot already existed. The controller reacts to `409 Already Exists` here:

https://github.com/kubernetes-csi/external-snapshotter/blob/59d72977e7a24ccdfabed3a20720e41beddc8216/pkg/common-controller/groupsnapshot_controller_helper.go#L626-L630

You can see the code continues processing when the snapshot already exists. However, the `createdVolumeSnapshot` variable has undefined content (it points to an object with empty namespace / name in my case) and later on, when the controller tries to update it, `Patch()` fails here:

https://github.com/kubernetes-csi/external-snapshotter/blob/59d72977e7a24ccdfabed3a20720e41beddc8216/pkg/common-controller/groupsnapshot_controller_helper.go#L649

Error:
`E0310 21:18:31.250917       1 groupsnapshot_controller_helper.go:257] could not sync group snapshot "e2e-volumegroupsnapshottable-8755/group-snapshot-z6w49": createSnapshotsForGroupSnapshotContent: binding volumesnapshot to volumesnapshotcontent resource name may not be empty`

The controller should perhaps `Get()` the VolumeSnapshot from its informer before creating it. And if it already exists, then fail + expect the VolumeSnapshot appears in the informer in the next retry.

All `IsAlreadyExists` should be then handled in a similar way, not just VolumeSnapshot.

**How to reproduce it**:
I don't know. It happened only once in OpenShift e2e tests. I am stress-testing the test, `1024 runs so far, 0 failures`.

Logs from OpenShift e2e tests:
* [snapshot-controller](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/openshift-kubernetes-2232-ci-4.19-e2e-aws-ovn-techpreview/1899177230127337472/artifacts/e2e-aws-ovn-techpreview/gather-extra/artifacts/pods/openshift-cluster-storage-operator_csi-snapshot-controller-6f7f496556-bwn5h_snapshot-controller.log) (look for `binding volumesnapshot to volumesnapshotcontent resource name may not be empty`)
* [CSI driver with external-snapshotter](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/openshift-kubernetes-2232-ci-4.19-e2e-aws-ovn-techpreview/1899177230127337472/artifacts/e2e-aws-ovn-techpreview/openshift-e2e-test/artifacts/junit/_sig-storage_OCP_CSI_Volumes/should_create_snapshots_for_multiple_volumes_in_a_pod/_Driver_csi-hostpath-groupsnapshot_OCPFeatureGate_VolumeGroupSnapshot_/should_create_snapshots_for_multiple_volumes_in_a_pod/_Testpattern_delete_policy_volumegroupsnapshottable_Feature_volumegroupsnapshot_/should_create_snapshots_for_multiple_volumes_in_a_pod/VolumeGroupSnapshottable/should_create_snapshots_for_multiple_volumes_in_a_pod/should_create_snapshots_for_multiple_volumes_in_a_pod/)
* [e2e test](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/openshift-kubernetes-2232-ci-4.19-e2e-aws-ovn-techpreview/1899177230127337472/artifacts/e2e-aws-ovn-techpreview/openshift-e2e-test/artifacts/e2e.log) (look for `VolumeSnapshot group-snapshot-z6w49 found but is not ready`)

As you can see, it would be helpful to log names of created VolumeSnapshots, VolumeSnapshotContents, and VolumeGroupContents - it's quite hard to map which VolumeSnapshot failed to be patched and what's the corresponding VGSC.

**Environment**:
- Driver version: csi-driver-hostpath + e2e tests as in Kubernetes 1.32
- Kubernetes version (use `kubectl version`): 1.32-ish

black-dragon74:
**What type of PR is this?**
/kind bug

This patch addresses the following concerns:

Reading the incorrect `VolumeSnapshotContent` in case it already exists:
```
This patch fixes an issue where the `VolumeSnapshot` would end up in a reconcile loop due
to the controller not being able to find the `VolumeGroupSnapshot` that was set in
the `VolumeSnapshotContent`'s status for a `VolumeSnapshot`.

The issue arose from a logic error which led to the
`VolumeSnapshotContent.Status.VolumeGroupSnapshotHandle` never being updated
except for the first/initial creation.

More specifically, the resource being passed to patch was empty in the cases
where `VolumeSnapshotContent` already existed on the cluster.

This patch passes the correct name to patch function to fix the issue.
```

Empty UID in case a `VolumeSnapshot` already exists:
```
The function `createSnapshotsForGroupSnapshotContent` responsible for
creating `VolumeSnapshots` and binding them to the
`VolumeSnapshotContent` has a logic error where it reads an empty
`UID` for the `VolumeSnapshot` if it already exists.

This patch addresses the issue by fetching the existing
`VolumeSnapshot` if it already exists. That way the `UID` will
always be set by the API server. An error is returned to retry the
reconcile if an empty `UID` is found on this fetched object.
```

Fixes #1280 

The `PatchVolumeSnapshotContent` only reads `ObjectMeta.Name`so there is no need to fetch the existing `VolumeSnapshotContent` from the API server.

```release-note
NONE
```

### DIFF
--- a/pkg/common-controller/groupsnapshot_controller_helper.go
+++ b/pkg/common-controller/groupsnapshot_controller_helper.go
@@ -625,9 +625,20 @@ func (ctrl *csiSnapshotCommonController) createSnapshotsForGroupSnapshotContent(
 		}
 
 		createdVolumeSnapshot, err := ctrl.clientset.SnapshotV1().VolumeSnapshots(volumeSnapshotNamespace).Create(ctx, volumeSnapshot, metav1.CreateOptions{})
-		if err != nil && !apierrs.IsAlreadyExists(err) {
+		if apierrs.IsAlreadyExists(err) {
+			createdVolumeSnapshot, err = ctrl.clientset.SnapshotV1().
+				VolumeSnapshots(volumeSnapshotNamespace).
+				Get(ctx, volumeSnapshot.Name, metav1.GetOptions{})
+		}
+		if err != nil {
 			return groupSnapshotContent, fmt.Errorf(
-				"createSnapshotsForGroupSnapshotContent: creating volumesnapshot %w", err)
+				"createSnapshotsForGroupSnapshotContent: error creating or fetching volumesnapshot %w", err)
+		}
+
+		// FIX for cases where the UID might be empty
+		if createdVolumeSnapshot.GetUID() == "" {
+			return groupSnapshotContent, fmt.Errorf(
+				"createSnapshotsForGroupSnapshotContent: created snapshot %s has an empty UID", createdVolumeSnapshot.Name)
 		}
 
 		// bind the volume snapshot content to the volume snapshot


pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -618,16 +618,27 @@ func (ctrl *csiSnapshotCommonController) createSnapshotsForGroupSnapshotContent(
...
+				Get(ctx, volumeSnapshot.Name, metav1.GetOptions{})
+		}
+		if err != nil {
 			return groupSnapshotContent, fmt.Errorf(
-				"createSnapshotsForGroupSnapshotContent: creating volumesnapshot %w", err)
+				"createSnapshotsForGroupSnapshotContent: error creating or fetching volumesnapshot %w", err)
+		}
+
+		// FIX for cases where the UID might be empty
+		if createdVolumeSnapshot.GetUID() == "" {
```
Madhu-1, line 639: i assume this can never happen as we go Create/Get.

black-dragon74: Should never happen, it is just a safety net (_can get rid of it if required_)

pkg/common-controller/groupsnapshot_controller_helper.go
```diff
@@ -667,7 +678,7 @@ func (ctrl *csiSnapshotCommonController) createSnapshotsForGroupSnapshotContent(
 		// set the snapshot handle and the group snapshot handle
 		// inside the volume snapshot content to allow
 		// the CSI Snapshotter sidecar to reconcile its status
-		_, err = utils.PatchVolumeSnapshotContent(createdVolumeSnapshotContent, []utils.PatchOp{
+		_, err = utils.PatchVolumeSnapshotContent(volumeSnapshotContent, []utils.PatchOp{
```
jsafrane, line 681: Since this function can be called several times for a single GroupSnapshot, it will always Patch the objects that might have been already patched. I don't think it's a big issue, but I can see some room for improvement.

PR #1364, opened 2026-01-18
Add aggregated RBAC for VolumeSnapshots

RoniKishner:
**What type of PR is this?**
>

/kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR adds aggregated ClusterRoles (snapshot-admin, snapshot-edit, snapshot-view) that automatically extend the built-in Kubernetes admin, edit, and view roles with permissions for VolumeSnapshot resources.

Non-admin users can GET/list PVCs but cannot access VolumeSnapshots, even when they have admin, edit, or view roles in a namespace. This is because PVC permissions are built into Kubernetes core, while VolumeSnapshot (as a CRD) requires explicit aggregated RBAC rules

Non-admin users can GET/list PVCs but cannot access VolumeSnapshots, even when they have admin, edit, or view roles in a namespace. This is because PVC permissions are built into Kubernetes core, while VolumeSnapshot (as a CRD) requires explicit aggregated RBAC rules. Since VolumeSnapshots are created from PVCs and used to restore PVCs, users who can manage PVCs should have equivalent access to VolumeSnapshots.

```release-note
Users with admin, edit, or view roles in a namespace can now access VolumeSnapshots without requiring cluster-admin privileges.
```

### DIFF
--- a/deploy/kubernetes/snapshot-controller/kustomization.yaml
+++ b/deploy/kubernetes/snapshot-controller/kustomization.yaml
@@ -3,4 +3,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - rbac-snapshot-controller.yaml
+  - rbac-snapshot-user.yaml
   - setup-snapshot-controller.yaml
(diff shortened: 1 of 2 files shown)


RoniKishner:
No response from dev team, closing this since no one responded to my messages

PR #1367, opened 2026-01-25
Add unit tests for group snapshots in snapshot-controller

xing-yang:
**What type of PR is this?**
/kind cleanup

This PR adds unit tests for group snapshots in the snapshot-controller.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/common-controller/groupsnapshot_sync_test.go
+++ b/pkg/common-controller/groupsnapshot_sync_test.go
@@ -1,0 +1,697 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common_controller
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	crdv1beta2 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta2"
+	crdv1 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumesnapshot/v1"
+	"github.com/kubernetes-csi/external-snapshotter/v8/pkg/utils"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/common-controller/groupsnapshot_sync_test.go
```diff
@@ -0,0 +1,517 @@
...
+	}
+
+	// Verify VolumeSnapshotInfoList is populated
+	if groupSnapshotContent.Status == nil || len(groupSnapshotContent.Status.VolumeSnapshotInfoList) == 0 {
+		return nil // No snapshot info, which is OK
+	}
+
+	// Success - the content is ready and has snapshot info
+	return nil
+}
```
jsafrane: This function never returns an error - what does it check?

pkg/common-controller/groupsnapshot_sync_test.go
```diff
@@ -0,0 +1,517 @@
...
+		},
+	}
+	runSyncTests(t, tests, nil, groupSnapshotClasses)
+}
+
+// TestGroupSnapshotContentSync tests the syncGroupSnapshotContent function
+func TestGroupSnapshotContentSync(t *testing.T) {
+	tests := []controllerTest{
+		{
+			name: "7-1 - sync group snapshot content triggers group snapshot update",
```
jsafrane: The test calls `testSyncGroupSnapshot`, not `testSyncGroupSnapshotContent`. Which makes me question what exactly does it test, if it succeeds, even when it calls a wrong function?

pkg/common-controller/groupsnapshot_sync_test.go
```diff
@@ -0,0 +1,517 @@
...
+// testSetDefaultGroupSnapshotClass is a test function that calls SetDefaultGroupSnapshotClass directly
+func testSetDefaultGroupSnapshotClass(ctrl *csiSnapshotCommonController, reactor *snapshotReactor, test controllerTest) error {
+	groupSnapshot := test.initialGroupSnapshots[0]
+	class, newGroupSnapshot, err := ctrl.SetDefaultGroupSnapshotClass(groupSnapshot)
+
+	// For pre-provisioned snapshots, class should be nil and no error
+	if groupSnapshot.Spec.Source.VolumeGroupSnapshotContentName != nil {
+		if class != nil {
+			return nil // This is actually OK, just means no default class was needed
+		}
```
jsafrane: The comment ("For pre-provisioned snapshots, class should be nil and no error") does not correspond with `if class != nil { return nil }`.

pkg/common-controller/groupsnapshot_sync_test.go
xing-yang: Fixed.

pkg/common-controller/groupsnapshot_sync_test.go
xing-yang: Fixed.

pkg/common-controller/groupsnapshot_sync_test.go
xing-yang: Fixed

xing-yang:
Comments are addressed.  Thanks.

PR #1368, opened 2026-01-25
Move VolumeGroupSnapshot API to V1

xing-yang:
**What type of PR is this?**
/kind feature

This PR adds VolumeGroupSnapshot V1 API and tries to move the feature to GA.

Fixes #

```release-note
Adds VolumeGroupSnapshot V1 API and moves the feature to GA.
```

### DIFF
--- a/pkg/common-controller/groupsnapshot_helper_test.go
+++ b/pkg/common-controller/groupsnapshot_helper_test.go
@@ -21,7 +21,7 @@ import (
 	"fmt"
 	"testing"
 
-	crdv1beta2 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta2"
+	groupsnapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1"
 	crdv1 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumesnapshot/v1"
 	"github.com/kubernetes-csi/external-snapshotter/client/v8/clientset/versioned/fake"
 	"github.com/kubernetes-csi/external-snapshotter/v8/pkg/utils"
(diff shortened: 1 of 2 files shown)


client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
```diff
@@ -174,5 +257,5 @@ spec:
         - driver
         type: object
     served: true
-    storage: true
+    storage: false
```
jsafrane: How much do we care about downgrade of the external-snapshotter or snapshot-controller? If we store group snapshots as v1 and someone needs to downgrade the CRD, then they won't be able to access v1 objects as v1beta2.

If we provide both v1 and v1beta2 APIs, but store them internally as v1beta2, downgrade will work smoothly.

jsafrane:
I checked the v1 and v1beta2 types.go are the same + the generated CRDs look good. I did not check the generated code.

* Should we update also the controller / snapshotter to use v1 API in the same PR?
* Should this be merged as the last PR, when all other tasks that block GA are merged (unit tests, e2e tests)?

client/config/crd/groupsnapshot.storage.k8s.io_volumegroupsnapshotclasses.yaml
xing-yang: When moving VolumeSnapshot to GA, we only store them as v1. Why should we do something different for VolumeGroupSnapshot?  @msau42 what do you think?

xing-yang:
/hold
Merge tests before merging this PR.

xing-yang:
Stress test is merged: https://github.com/kubernetes/kubernetes/pull/136845

jsafrane:
I've checked e2e tests and unit tests. E2e is in a decent shape. I see quite some deficiency in the unit tests:

* `pkg/common-controller/groupsnapshot_controller_helper.go` has 57% coverage, which may seem enough, but important function [createIndividualSnapshotForGroupSnapshot](https://github.com/kubernetes-csi/external-snapshotter/blob/668fe10860bb8e9abda43746324a99c417661d89/pkg/common-controller/groupsnapshot_controller_helper.go#L580) is not covered by any test. As result,  various functions like `createOrGetVolumeSnapshotContent`, `createOrGetVolumeSnapshot`, `bindSnapshotContentToSnapshot` and `updateVolumeSnapshotContentStatus` have no tests either.

* `pkg/sidecar-controller/groupsnapshot_helper.go` has 82% coverage and all important functions seem to be covered.

xing-yang:
Submitted a unit test PR: https://github.com/kubernetes-csi/external-snapshotter/pull/1407

PR #1394, opened 2026-03-14
Fix VolumeSnapshot deletion not retrying when used for PVC restore

### LINKED ISSUE #1366: VolumeSnapshot deletion failed
**What happened**:
If a PVC is created using a VolumeSnapshot, and the VolumeSnapshot is quickly deleted while the PVC is still pending, the VolumeSnapshot deletion will fail.

There should be a retry mechanism for deleting VolumeSnapshot.

**How to reproduce it**:
https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/common-controller/snapshot_controller.go#L291

Looking at the code implementation, it doesn't retry if the deletion fails even once, which is unreasonable.

**Environment**:
- Driver version: Any Version
- Kubernetes version (use `kubectl version`): Any Version
- OS (e.g. from /etc/os-release): Any OS
- Kernel (e.g. `uname -a`):  Any
- Install tools:
- Others:

Paramesh324:
**What type of PR is this?**
/kind bug

When a VolumeSnapshot is deleted while a PVC is still being created from it, the controller blocked deletion but returned `nil`, so the workqueue never retried. This change returns an error instead so the snapshot is requeued and deletion is retried once the PVC is no longer in use.

The same fix is applied for VolumeGroupSnapshot in the group snapshot helper.

Fixes #1366

```release-note
Fixed VolumeSnapshot (and VolumeGroupSnapshot) deletion getting stuck when the snapshot was in use for a PVC restore. The controller now requeues and retries deletion until the PVC is no longer in use.```

### DIFF
--- a/pkg/common-controller/groupsnapshot_controller_helper.go
+++ b/pkg/common-controller/groupsnapshot_controller_helper.go
@@ -1611,8 +1611,8 @@ func (ctrl *csiSnapshotCommonController) processGroupSnapshotWithDeletionTimesta
 			msg := fmt.Sprintf("Snapshot %s belonging to VolumeGroupSnapshot %s is being used to restore a PVC", utils.SnapshotKey(snapshot), utils.GroupSnapshotKey(groupSnapshot))
 			klog.V(4).Info(msg)
 			ctrl.eventRecorder.Event(groupSnapshot, v1.EventTypeWarning, "SnapshotDeletePending", msg)
-			// TODO(@xiangqian): should requeue this?
-			return nil
+			// Return error so the workqueue will requeue and retry deletion when the PVC is no longer in use.
+			return fmt.Errorf("%s, will retry deletion", msg)
 		}
 
 	}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by Paramesh324** [comment]
@pohly @RaunakShah Can you please review this.

**Comment by nevermore-muyi** [comment]
LGTM

**Comment by Paramesh324** [comment]
@pohly @RaunakShah @nevermore-muyi Can you please review this.

**Comment by nevermore-muyi** [comment]
LGTM

File: pkg/common-controller/groupsnapshot_controller_helper.go [hunk 1]
```diff
@@ -1611,8 +1611,8 @@ func (ctrl *csiSnapshotCommonController) processGroupSnapshotWithDeletionTimesta
 			msg := fmt.Sprintf("Snapshot %s belonging to VolumeGroupSnapshot %s is being used to restore a PVC", utils.SnapshotKey(snapshot), utils.GroupSnapshotKey(groupSnapshot))
 			klog.V(4).Info(msg)
 			ctrl.eventRecorder.Event(groupSnapshot, v1.EventTypeWarning, "SnapshotDeletePending", msg)
-			// TODO(@xiangqian): should requeue this?
-			return nil
+			// Return error so the workqueue will requeue and retry deletion when the PVC is no longer in use.
+			return fmt.Errorf("%s, will retry deletion", msg)
```
landreasyan: I think the error should say that snapshot is in use.

Paramesh324: Thanks for the review. Updated the returned error to state explicitly that the snapshot is in use.

Paramesh324:
@landreasyan @xing-yang @richabanker if you could help review from sig-storage-approvers side, that'd be of great help.
Thanks.

xing-yang:
Can you add a unit test?

Paramesh324:
> Can you add a unit test?

@xing-yang Could you please review now?

PR #1397, opened 2026-03-18
Added validation that creation time is not zero (epoch)

tomerf:
**What type of PR is this?**

/kind bug

This PR updates the handling of `CreationTime` in the snapshot controller to avoid persisting a zero value (`*int64 == 0`) when returned by the CSI driver.

Currently, if the CSI driver returns `CreationTime = 0`, the snapshot controller treats it as a valid value and sets it on `VolumeSnapshotContent` and `VolumeSnapshot`. Since `CreationTime` is only set once, this prevents any future updates with the actual timestamp.

Some CSI drivers create snapshots asynchronously and cannot determine the creation time at the moment of the initial `CreateSnapshot` call. In such cases, returning `0` is used to indicate that the real value is not yet available.

Persisting this zero value results in incorrect snapshot metadata (epoch time) and prevents updating the field later when the actual creation time becomes available.

This PR changes the behavior so that a zero value is treated as "not set":
- If the CSI driver returns `CreationTime = 0`, the field is not set on `VolumeSnapshotContent` and `VolumeSnapshot`
- This allows the controller to populate it later when a valid timestamp is returned

NONE

This change preserves existing behavior for drivers that return a valid `CreationTime`, and improves compatibility with drivers that perform asynchronous snapshot creation.

```release-note
Fix handling of CreationTime in external-snapshotter: zero values returned by CSI drivers are now treated as "not set" and will not be persisted, allowing the field to be updated later when a valid timestamp is available.

### DIFF
--- a/pkg/common-controller/snapshot_controller.go
+++ b/pkg/common-controller/snapshot_controller.go
@@ -1189,7 +1189,7 @@ func (ctrl *csiSnapshotCommonController) needsUpdateSnapshotStatus(snapshot *crd
 	if snapshot.Status.BoundVolumeSnapshotContentName == nil {
 		return true
 	}
-	if snapshot.Status.CreationTime == nil && content.Status.CreationTime != nil {
+	if snapshot.Status.CreationTime == nil && content.Status.CreationTime != nil && *content.Status.CreationTime != 0 {
 		return true
 	}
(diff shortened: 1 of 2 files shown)


tomerf:
Who's responsible to review this? @xing-yang

tomerf:
/remove-lifecycle stale
@humblec @sunnylovestiramisu Can you take a look in this small change?

PR #1435, opened 2026-06-05
Change codes.ResourceExhausted from non-final to final error

### LINKED ISSUE #1434: `ResourceExhausted` is retried indefinitely on snapshot/clone create, doesn't match external-provisioner behavior

`external-snapshotter` treats `codes.ResourceExhausted` from the CSI driver as a non-final error and keeps retrying. For capacity-driven failures (no pool with enough free space to host the snapshot or clone), the condition is permanent until an operator does something about it, but the side-car keeps hammering the control-plane until either a human intervenes or the pool magically gets more space. The error never surfaces on the VolumeSnapshot object so the user has no terminal signal either.

`external-provisioner` already made this change in [kubernetes-csi/external-provisioner#675](https://github.com/kubernetes-csi/external-provisioner/pull/675) for exactly this reason. `external-snapshotter` was left out, and the two side-cars now disagree on what `ResourceExhausted` means.

We've been hitting this in production on OpenEBS Mayastor. Snapshot create or restore returns `ResourceExhausted` when a pool is full (HTTP 507 from the REST layer, mapped to `codes.ResourceExhausted` in the CSI gRPC response), and the snapshotter retries forever.

Move `codes.ResourceExhausted` from the non-final list to the final list on the snapshot create path so the side-car surfaces the error to the VolumeSnapshot status instead of retrying indefinitely. Same shape as the external-provisioner fix in [kubernetes-csi/external-provisioner#675](https://github.com/kubernetes-csi/external-provisioner/pull/675), which was the clean "treat ResourceExhausted as final" change in that side-car.

[kubernetes-csi/external-snapshotter#1334](https://github.com/kubernetes-csi/external-snapshotter/pull/1334) (authored by @xing-yang) already did exactly this. It was small (11/4 lines), had the `approved` label, but stayed in WIP and got auto-closed by the triage bot after ~150 days of inactivity.

Can open a new PR with the same change if it helps move things along.

- Returning `FailedPrecondition` from the CSI driver instead of `ResourceExhausted` for capacity errors. This works as a workaround (we're doing this in [openebs/mayastor-control-plane#1114](https://github.com/openebs/mayastor-control-plane/pull/1114)) but it's misleading semantically, and we'd rather revert it once the side-car is fixed.
- Letting the retries continue and handling the noise on the control-plane side. Doesn't help the user, who still sees no terminal error on the VolumeSnapshot object.

**Additional context**

- [container-storage-interface/spec#419](https://github.com/container-storage-interface/spec/issues/419): the CSI spec is ambiguous about whether `ResourceExhausted` should be final or retriable. external-provisioner picked "final".
- xing-yang's reasoning on the original PR thread covers the same ground, including a [#sig-storage Slack discussion](https://kubernetes.slack.com/archives/C09QZFCE5/p1688119656274719) from 2023.

cc @xing-yang @jingxu97

yugchaudhari:
**What type of PR is this?**

/kind cleanup

`external-snapshotter` currently treats `codes.ResourceExhausted` from the CSI driver as a non-final error and keeps retrying. For capacity failures (no pool has enough free space to host the snapshot or clone), the condition is permanent until an operator does something about it, so retrying forever is wasted work on the control-plane and the user never gets a terminal signal on the VolumeSnapshot object.

`external-provisioner` already made the same change in [kubernetes-csi/external-provisioner#675](https://github.com/kubernetes-csi/external-provisioner/pull/675). This PR brings `external-snapshotter` in line with that behavior.

The transport-layer "message size exceeded" interpretation of `ResourceExhausted` is unlikely in practice for snapshot operations. The comment block added in the code calls that out and links the spec discussion ([container-storage-interface/spec#419](https://github.com/container-storage-interface/spec/issues/419)) for posterity.

Fixes #1434

This carries over the change from [#1334](https://github.com/kubernetes-csi/external-snapshotter/pull/1334) (authored by @xing-yang), which was approved but stayed in WIP and got auto-closed by the triage bot after ~150 days of inactivity. The diff here is identical to xing-yang's.

We've been hitting this in production on OpenEBS Mayastor: snapshot create or restore returns `ResourceExhausted` when a pool is full, and the snapshotter retries forever. We're shipping a `FailedPrecondition` workaround on the driver side ([openebs/mayastor-control-plane#1114](https://github.com/openebs/mayastor-control-plane/pull/1114)) while this lands, and would like to revert that once the side-car is fixed.

```release-note
Change codes.ResourceExhausted from non-final to final error in the snapshot create path. The side-car no longer retries indefinitely on capacity failures from the CSI driver; the error is surfaced to the VolumeSnapshot status instead. Matches existing external-provisioner behavior.
```

### DIFF
--- a/pkg/sidecar-controller/snapshot_controller.go
+++ b/pkg/sidecar-controller/snapshot_controller.go
@@ -735,12 +735,19 @@ func isCSIFinalError(err error) bool {
 	}
 	switch st.Code() {
 	case codes.Canceled, // gRPC: Client Application cancelled the request
-		codes.DeadlineExceeded,  // gRPC: Timeout
-		codes.Unavailable,       // gRPC: Server shutting down, TCP connection broken - previous CreateSnapshot() may be still in progress.
-		codes.ResourceExhausted, // gRPC: Server temporarily out of resources - previous CreateSnapshot() may be still in progress.
-		codes.Aborted:           // CSI: Operation pending for Snapshot
+		codes.DeadlineExceeded, // gRPC: Timeout
+		codes.Unavailable,      // gRPC: Server shutting down, TCP connection broken - previous CreateSnapshot() may be still in progress.
+		codes.Aborted:          // CSI: Operation pending for Snapshot
 		return false
 	}
+	// Note: codes.ResourceExhausted is treated as a final error.
+	// gRPC: Server out of resources.
+	// However, it also could be from the transport layer for "message size exceeded".
+	// Cannot be decided properly here and needs to be resolved in the spec
+	// https://github.com/container-storage-interface/spec/issues/419.
+	// What we assume here for now is that message size limits are large enough that
+	// the error really comes from the CSI driver.
+
 	// All other errors mean that creating snapshot either did not
 	// even start or failed. It is for sure not in progress.
 	return true


yugchaudhari:
> /ok-to-test

@xing-yang gentle ping for /approve when you have a moment. Happy to address anything outstanding.

PR #1475, opened 2026-08-17
Enable VGS feature gate

xing-yang:
**What type of PR is this?**
/kind cleanup

This PR enables CSIVolumeGroupSnapshot feature gate to true by default.
(1) Made changes so if feature gate is not set all (default) and if CRD not installed, a warning message will be logged but no failure (VGS feature will be disabled).
(2) If feature gate is set to true and if CRD not installed, a warning message will be logged but no failure (same as if VGS feature is disabled).
(3) If feature gate is set to false, then CRD won’t be required and no warning msg.

- Enable feature gate but don't remove the feature gate in this release.
- In the next release, the feature gate will be deleted.

Fixes #

```release-note
Enable CSIVolumeGroupSnapshot feature gate to true by default.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,13 +116,11 @@ Other than this, the NODE_NAME environment variable must be set where the CSI sn
 
 ### Volume Group Snapshot Support
 
-The following requisites must be met to enable the volume group snapshot feature:
+The `CSIVolumeGroupSnapshot` feature gate is General Availability (GA) and enabled by default.
 
-* the Volume Group Snapshot CRDs are installed in the cluster
-* the `--feature-gates=CSIVolumeGroupSnapshot=true` option is being passed to the snapshot controller
-* the `--feature-gates=CSIVolumeGroupSnapshot=true` option is being passed to the CSI snapshotter sidecar
+To use the volume group snapshot feature, install the Volume Group Snapshot CRDs in the cluster.
 
-Specifically, `deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml` needs to be updated with `--feature-gates=CSIVolumeGroupSnapshot=true` in order to enable this feature in the snapshot controller.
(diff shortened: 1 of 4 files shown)


cmd/snapshot-controller/main.go
```diff
@@ -230,6 +232,40 @@ func main() {
...
+	// other feature gates it may be "on" without the admin ever having asked
+	// for it. If its CRDs aren't installed and the gate wasn't explicitly set
+	// on the command line, don't fail startup of the whole controller for a
+	// feature nobody asked to use -- disable it for this run and carry on.
+	// If the gate was explicitly requested, missing CRDs are a hard error.
+	_, volumeGroupSnapshotExplicitlySet := featureGates[string(features.VolumeGroupSnapshot)]
+	enableVolumeGroupSnapshots := utilfeature.DefaultFeatureGate.Enabled(features.VolumeGroupSnapshot)
+	if enableVolumeGroupSnapshots {
+		if err := ensureVolumeGroupSnapshotCRDsExist(snapClient); err != nil {
+			if volumeGroupSnapshotExplicitlySet {
```
jsafrane: I think it should not matter if the feature gate was enabled explicitly or not. In both cases, the  snapshoto controller should log a warning and continue with disabled group snapshot controllers. So users (kubernetes vendors) have consistent behavior when we remove the feature gate in the future.

README.md
```diff
@@ -159,7 +165,7 @@ Specifically, `deploy/kubernetes/snapshot-controller/setup-snapshot-controller.y
 
 #### Volume Group Snapshot support
 
-* `--feature-gates=CSIVolumeGroupSnapshot=true`: Enables support for Volume Group Snapshots. If this option is enabled, the VolumeGroupSnapshots CRD should be available on the cluster.
+* `--feature-gates=CSIVolumeGroupSnapshot=true`: Enables support for Volume Group Snapshots. This feature is GA and enabled by default. If this option is enabled, the VolumeGroupSnapshots CRD should be available on the cluster.
```
jsafrane: What if they are not? Here the CSI driver vendor has very little control over CRDs available in a random cluster. IMO the external-snapshotter should do the same as the controller - if the CRDs are availabe, run group snapshot controller, and if not, then log a warning and continue without them.

cmd/snapshot-controller/main.go
nixpanic: Agreed, I like this approach more too. Users may update to the latest snapshot controller without having the group snapshot CRD. A mentioning in the release notes about the required CRD is easily missed. Troubleshooting why snapshots do not work anymore after an update is not a pleasant exercise.

xing-yang: Done

README.md
xing-yang: Done

jsafrane:
this needs to merge first: https://github.com/kubernetes-csi/external-snapshotter/pull/1472

PR #1485, opened 2026-09-14
Reject snapshot when class driver differs from source PV CSI driver

### LINKED ISSUE #1484: Content created when VolumeSnapshotClass driver != source PV CSI driver
**What happened**:

When a `VolumeSnapshot` explicitly sets `spec.volumeSnapshotClassName` to a `VolumeSnapshotClass` whose `driver` differs from the CSI driver of the PV bound to the source PVC, the common snapshot-controller still creates a `VolumeSnapshotContent` with `spec.driver` copied from the class and `spec.source.volumeHandle` copied from the (foreign) PV.

The csi-snapshotter sidecar for the class driver then calls `CreateSnapshot` with a volume handle it can never understand. The gRPC error is `Internal`/`InvalidArgument`, which `isCSIFinalError` treats as final only for the purpose of removing the being-created annotation; the content is requeued indefinitely. Result: the snapshot stays `readyToUse: false` forever, and the error surfaced on both the `VolumeSnapshot` and the `VolumeSnapshotContent` is the storage provider's confusing message about an invalid volume ID rather than a driver mismatch.

Observed in practice with a `VolumeSnapshotClass` for `ebs.csi.aws.com` applied namespace-wide by a snapshot scheduler to PVCs that were provisioned by `efs.csi.aws.com`. Sanitized logs from the EBS controller pod:

```
# ebs-plugin
driver.go:135] "GRPC error" err="rpc error: code = Internal desc = Could not create snapshot \"snapshot-<uid>\": error creating snapshot of volume efs:fs-<id>::fsap-<id>: operation error EC2: CreateSnapshot, https response error StatusCode: 400, ... InvalidParameterValue: Value (efs:fs-<id>::fsap-<id>) for parameter volumeId is invalid. Expected: 'vol-...'."

# csi-snapshotter sidecar
snapshot_controller.go] createSnapshotWrapper: CreateSnapshot for content snapcontent-<uid> returned error: rpc error: code = Internal desc = Could not create snapshot ...
snapshot_controller_base.go] could not sync content "snapcontent-<uid>": failed to take snapshot of the volume efs:fs-<id>::fsap-<id>: ...
```

```
$ kubectl get volumesnapshotcontent -o json | jq '.items[] | select(.spec.source.volumeHandle|startswith("efs:")) | {driver: .spec.driver, class: .spec.volumeSnapshotClassName, handle: .spec.source.volumeHandle, ready: .status.readyToUse}'
{"driver": "ebs.csi.aws.com", "class": "<ebs-class>", "handle": "efs:fs-<id>::fsap-<id>", "ready": false}
```

Over a 72h window one cluster produced ~1700 such `GRPC error` lines across 54 distinct snapshot ids from four permanently-failing contents.

**Where in the code** (master @ afb2f0d):

* `getCreateSnapshotInput` resolves the class and the PV independently and never compares `class.Driver` with `volume.Spec.CSI.Driver`: https://github.com/kubernetes-csi/external-snapshotter/blob/afb2f0d85e00e40cda621549d56fda11ccd3558b/pkg/common-controller/snapshot_controller.go#L824-L855
* `createSnapshotContent` then builds the content with `Driver: class.Driver` and `VolumeHandle: &volume.Spec.CSI.VolumeHandle`; the only PV-side check is `volume.Spec.CSI == nil`: https://github.com/kubernetes-csi/external-snapshotter/blob/afb2f0d85e00e40cda621549d56fda11ccd3558b/pkg/common-controller/snapshot_controller.go#L739-L759
* The default-class path *does* enforce `pvDriver == class.Driver` (via `pvDriverFromSnapshot`), so the gap only bites when `spec.volumeSnapshotClassName` is set explicitly: https://github.com/kubernetes-csi/external-snapshotter/blob/afb2f0d85e00e40cda621549d56fda11ccd3558b/pkg/common-controller/snapshot_controller.go#L1487-L1499
* `VolumeGroupSnapshot` already has exactly this guard since #1098 (fixes #1096, "Volume CSI driver (%s) mismatch with VolumeGroupSnapshotClass"): https://github.com/kubernetes-csi/external-snapshotter/blob/afb2f0d85e00e40cda621549d56fda11ccd3558b/pkg/common-controller/groupsnapshot_controller_helper.go#L1190-L1200 — the single-snapshot path never got the equivalent.
* Sidecar: `isCSIFinalError` returns true for `Internal`, but the content is still requeued by the sync loop, so the failure never becomes terminal: https://github.com/kubernetes-csi/external-snapshotter/blob/afb2f0d85e00e40cda621549d56fda11ccd3558b/pkg/sidecar-controller/snapshot_controller.go#L725-L746

The controller should refuse to create a `VolumeSnapshotContent` whose `spec.driver` does not own the source PV, record a clear error on the `VolumeSnapshot` (`status.error.message` naming both drivers) and emit `Warning SnapshotContentCreationFailed` — the same behaviour the default-class path and the VolumeGroupSnapshot path already have. No orphaned content, no infinite `CreateSnapshot` retry loop against the wrong driver.

**How to reproduce it**:

Unit test (no cluster needed), `pkg/common-controller`, run with `go test ./pkg/common-controller/ -run TestReproDriverMismatch -v`:

```go
func TestReproDriverMismatch(t *testing.T) {
	// class "gold" belongs to a driver other than the one that provisioned the PV (mockDriverName = "csi-mock-plugin")
	classes := []*crdv1.VolumeSnapshotClass{{ObjectMeta: metav1.ObjectMeta{Name: classGold}, Driver: "csi-other-plugin", DeletionPolicy: crdv1.VolumeSnapshotContentDelete}}
	tests := []controllerTest{
		{
			name:              "X-1 - refuse to create content when class driver != PV driver",
			initialContents:   nocontents,
			expectedContents:  nocontents,
			initialSnapshots:  newSnapshotArray("snapX-1", "snapuidX-1", "claimX-1", "", classGold, "", &False, nil, nil, nil, false, true, nil),
			expectedSnapshots: newSnapshotArray("snapX-1", "snapuidX-1", "claimX-1", "", classGold, "", &False, nil, nil, newVolumeError("Failed to create snapshot content with error failed to get input parameters to create snapshot snapX-1: \"VolumeSnapshotClass \\\"gold\\\" has driver \\\"csi-other-plugin\\\" but the source PV \\\"volumeX-1\\\" is provisioned by driver \\\"csi-mock-plugin\\\"\""), false, true, nil),
			initialClaims:     newClaimArray("claimX-1", "pvc-uidX-1", "1Gi", "volumeX-1", v1.ClaimBound, &classEmpty),
			initialVolumes:    newVolumeArray("volumeX-1", "pv-uidX-1", "pv-handleX-1", "1Gi", "pvc-uidX-1", "claimX-1", v1.VolumeBound, v1.PersistentVolumeReclaimDelete, classEmpty),
			expectedEvents:    []string{"Warning SnapshotContentCreationFailed"},
			errors:            noerrors,
			test:              testSyncSnapshot,
		},
	}
	runSyncTests(t, tests, classes, nil)
}
```

On master this fails: the harness observes a `VolumeSnapshotContent` created with `Driver: "csi-other-plugin"` and `Source.VolumeHandle: "pv-handleX-1"`, and the snapshot has no error. (Note: build the class as a struct literal — `newSnapshotClassArray` stamps a namespace, which makes the cluster-scoped lister `Get` miss.)

Manual: create a `VolumeSnapshotClass` for driver A, a PVC provisioned by driver B, and a `VolumeSnapshot` naming class A. A content with `spec.driver: A` and B's volume handle appears and never becomes ready.

Proposed fix — in `getCreateSnapshotInput`, right after `getVolumeFromVolumeSnapshot` succeeds:

```go
// The content is handed to the sidecar of class.Driver, so the PV must belong to that driver.
if volume.Spec.CSI != nil && volume.Spec.CSI.Driver != class.Driver {
	return nil, nil, "", nil, fmt.Errorf("VolumeSnapshotClass %q has driver %q but the source PV %q is provisioned by driver %q", class.Name, class.Driver, volume.Name, volume.Spec.CSI.Driver)
}
```

The existing error path in `syncUnreadySnapshot` → `createSnapshotContent` already records the error on the `VolumeSnapshot` and emits `SnapshotContentCreationFailed`, so no orphaned content is created. Verified locally on afb2f0d: with this patch the test above passes and the full `pkg/common-controller` suite stays green. Happy to send the PR with the unit test.

Analysis prepared with an AI agent operated by KR-Ravindra, who verified the code paths.

**Environment**:
- Driver version: snapshot-controller / csi-snapshotter v8.x (code paths unchanged on master afb2f0d); observed with aws-ebs-csi-driver + aws-efs-csi-driver
- Kubernetes version (use `kubectl version`): v1.35 (EKS)
- OS (e.g. from /etc/os-release): n/a (controller bug, reproduced by unit test)
- Kernel (e.g. `uname -a`): n/a
- Install tools: Helm
- Others:

KR-Ravindra:
**What type of PR is this?**

/kind bug

#### Problem
A `VolumeSnapshot` that explicitly sets `spec.volumeSnapshotClassName` to a `VolumeSnapshotClass` whose `driver` differs from the CSI driver of the PV bound to the source PVC still gets a `VolumeSnapshotContent`. The content carries `spec.driver` from the class and `spec.source.volumeHandle` from the foreign PV, so the class driver's csi-snapshotter sidecar retries `CreateSnapshot` forever with a handle it cannot parse. The snapshot never becomes `readyToUse` and the surfaced error is the storage backend's "invalid volume id" message rather than a driver mismatch. Seen in a multi-driver cluster where a scheduler applied one class namespace-wide.

Root cause: `getCreateSnapshotInput` (`pkg/common-controller/snapshot_controller.go`) resolves the class and the PV independently and never compares `class.Driver` with `volume.Spec.CSI.Driver`; `createSnapshotContent` then builds the content with `Driver: class.Driver`. The default-class path (`SetDefaultSnapshotClass`, `pvDriver == class.Driver`) and the VolumeGroupSnapshot path (`groupsnapshot_controller_helper.go`, "Volume CSI driver mismatch with VolumeGroupSnapshotClass") already guard against this; the explicit-class single-snapshot path was the only gap.

#### Fix
In `getCreateSnapshotInput`, right after the PV is resolved, return an error naming both drivers when `volume.Spec.CSI.Driver != class.Driver`. The existing error path in `syncUnreadySnapshot` records it on `status.error` and emits `Warning SnapshotContentCreationFailed`; no content is created, so nothing is handed to the wrong sidecar.

#### How tested
Added case `7-12` to `TestCreateSnapshotSync` with a class for a second driver (`csi-other-plugin`) and a PV provisioned by `csi-mock-plugin`.

Before the fix:
```
--- FAIL: TestCreateSnapshotSync (5.14s)
    Test "7-12 - fail create snapshot when snapshot class driver does not match the PV driver": content check failed [A-expected, B-got]:
        + 	"snapcontent-snapuid7-12": &{
        + 			Driver:                  "csi-other-plugin",
        + 			Source:                  v1.VolumeSnapshotContentSource{VolumeHandle: &"pv-handle7-12"},
```
After the fix:
```
ok  	github.com/kubernetes-csi/external-snapshotter/v8/pkg/common-controller	1.658s
```
`gofmt`, `go vet ./pkg/common-controller/` and `go build ./cmd/snapshot-controller/` are clean.

Fixes https://github.com/kubernetes-csi/external-snapshotter/issues/1484

The check is skipped when `volume.Spec.CSI == nil` so the existing "cannot find CSI PersistentVolumeSource" error in `createSnapshotContent` still fires for non-CSI PVs. This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

```release-note
The snapshot-controller now fails a VolumeSnapshot with a clear error instead of creating a VolumeSnapshotContent when the VolumeSnapshotClass driver differs from the CSI driver of the source PV.
```


KR-Ravindra:
Round 1 self-review.
1. Re-checked against master afb2f0d: `getCreateSnapshotInput`'s only caller is `createSnapshotContent`, so the new check cannot reach any other path, and the `volume.Spec.CSI == nil` skip leaves the existing "cannot find CSI PersistentVolumeSource" error intact.
2. The default-class path (`SetDefaultSnapshotClass`) and the VolumeGroupSnapshot path already enforce this driver match; the explicit-class snapshot path was the only one without it.
3. Test 7-12 fails on master with only the test files applied (content is created with driver `csi-other-plugin` and the event is `Normal CreatingSnapshot`) and passes with the fix; `go test ./pkg/common-controller/` is green locally.
4. Nit I can tidy on request: 7-12 sits above 7-11 in the table.
Marking ready for review; prow needs /ok-to-test from a member.

